PR #2, opened 2016-01-13
Reorg repo, setup sphynx for doc, py.test and tox for testing

cdeepakroy:
Summary of this pull request
* Reorganized the repository
  * Moved all analysis code into histomicstk (lemme know if you want to change the name but i think we should pick one now) and made it a python package. People who use our code as    a python library will import functionality from here. We can further organize the stuff here by making subdirectories for preprocessing, segmentation, feature_extraction etc. Each subdirectory under histomicstk will have a tests folder with tests for functions in the directory. test files should be named as test_*.py for py.test or nose to recognize them.
  * The server subdirectory will contain the code for the pipelines we want to expose as REST end points when used as a plugin of girder
* Setup [sphynx](http://www.sphinx-doc.org/en/stable/) for documentation. We will use [numpy style docstrings](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for in-code documentation. This style is used by scikit-learn as well and is human-friendly. [Here](https://github.com/numpy/numpy/blob/master/doc/example.py) is an example to give you a sense.
* Setup [tox](https://tox.readthedocs.org/en/latest/) for automated testing using py.test. See [this](http://pytest.org/latest/) page to learn how to write tests to be recognized by py.test

TODO:
* Publicly host the documentation on [ReadTheDocs](https://readthedocs.org/)
* Setup continuous integration testing using [TravisCI](https://travis-ci.com/)


cdeepakroy:
@cooperlab can you take a look

cooperlab:
Deepak,

This is really great - lots of progress.

I'd like to keep pushing ahead to get version 1 of a pipeline done and we may have to do some retrospective fixing later as I learn more about the documentation and testing frameworks.

-Lee

cdeepakroy:
@Lee agreed. Lets get the functions for a tangible pipeline first and then we can fix once retrospectively to learn the documentation and testing frameworks. 

Is the name histomicstk ok? It goes into a lot of files. So it would be nice if we could come up with one and fix it.

cooperlab:
I think we need to discuss the name with David in the loop and make a decision today.

I like it - definitely good to match what Hugo and their team are doing with "radiomics". Also like TK for consistency with other Kitware efforts.

Another U24 digital pathology team led by Case Western likes to use "pathomics" which could create some trouble.

-Lee

cdeepakroy:
@zachmullen when you can could you please take a look to see if i am setting up the documentation and testing right and aligns with the approach of other kitware repos? 

I am following the guidelines [here](https://www.jeffknupp.com/blog/2013/08/16/open-sourcing-a-python-project-the-right-way/) and [here](http://www.pydanny.com/cookie-project-templates-made-easy.html).

zachmullen:
Looks good, I haven't used tox but it seems like it should work fine. We should think about harnessing this inside girder's test infrastructure since this is a girder plugin.

Also, why the switch from plugin.yml to plugin.json?

cdeepakroy:
@zachmullen Just tried to [follow](https://github.com/Kitware/romanesco/blob/master/plugin.json) romanesco on the switch to plugin.json. Also it makes [this](https://github.com/cdeepakroy/HistomicsTK/blob/master/setup.py#L20) easier. Is there a reason to be using yaml?

Also, i still dont know how to write tests for the REST end points when this toolkit will be used as a plugin to girder. Do you foresee any setbacks if we use tox?

Also, according to [this](http://www.pydanny.com/pytest-no-boilerplate-testing.html) article py.test seems to reduce quite a bit of boiler plate when writing tests. Are there any strong reasons to not use it for this project?

zachmullen:
> Also it makes this easier. Is there a reason to be using yaml?

That sounds good to me. There's not much reason to prefer one over the other, so this is fine.

> Also, i still dont know how to write tests for the REST end points when this toolkit will be used as a plugin to girder. Do you foresee any setbacks if we use tox?

I have no experience with tox. The girder.tests.base module is written for unittest and contains a bunch of utilities that make it easy to test girder REST interfaces; as long as we can still use that, we should be OK.

> Also, according to this article py.test seems to reduce quite a bit of boiler plate when writing tests. Are there any strong reasons to not use it for this project?

After reading the article, the main reason not to use it would be that it simply doesn't provide us with anything we don't already have. We already have an [exception testing context manager](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises) via unittest, fixtures are pretty overrated in my opinion (due to not being reusable enough in practice), unittest gives us intuitive and easy setup/teardown at the module and test function level. That is to say: the "boilerplate" mentioned in the article, i.e. writing a class name and using a function for assertions, are trivial parts of writing unit tests and not important enough to merit introducing new technologies that might make us have to sacrifice some of our valuable existing testing utilities such as [this class](https://github.com/girder/girder/blob/master/tests/base.py#L127).

As far as test discovery, I don't consider that a useful feature to us since we like to harness our tests with CMake/CTest in order to get compelling features like running in parallel in accordance with an explicit dependency graph, named mutexes for tests, and config-time selection of what tests to run and how to run them (e.g. with or without coverage).

cdeepakroy:
> As far as test discovery, I don't consider that a useful feature to us since we like to harness our tests with CMake/CTest in order to get compelling features like running in parallel in accordance with an explicit dependency graph, named mutexes for tests, and config-time selection of what tests to run and how to run them (e.g. with or without coverage).

So what do i have to do to setup CMake/CTest to discover tests? Could you point me to some place to learn this?

jbeezley:
Personally, I like py.test and tox a lot.  All sorts of packages and cloud services integrate with them automagically.  There are also plugins to do things like running in parallel, and tox makes it trivial to test against multiple python versions.

As a girder plugin, you lose all of the ctest/cdash integration as well as (probably) most of the testing infrastructure girder provides.  I think it would be great to use a more feature-filled unittesting library for girder, but it is the kind of thing that belongs on the 2.0 wishlist.

zachmullen:
> So what do i have to do to setup CMake/CTest to discover tests? Could you point me to some place to learn this?

If we test this *inside* of girder, we can get all of its test infrastructure for free and just have our CI run a subset of the tests that correspond to this repo. I think this is what we should probably do anyway so we can get test coverage on the girder plugin side of things. We do this for COVALIC using a simple regex, and I'd like to add a little CMake logic into girder core that lets you only run the tests for certain plugins to avoid running all the core and built-in plugin tests as well.

> which of the following two ways do you think is better and works well with CMake/CTest:

Let's put them in a separate tests directory since we don't want them to be part of the package.

zachmullen:
> As a girder plugin, you lose all of the ctest/cdash integration as well as (probably) most of the testing infrastructure girder provides. I think it would be great to use a more feature-filled unittesting library for girder, but it is the kind of thing that belongs on the 2.0 wishlist.

Tox as a top-level test harness is a non-starter for girder IMO since we have unit tests for code in languages other than python. If we are certain that this repo will only ever contain python code, maybe we can go ahead and invest in a more off-the-shelf, python-specific harness, but I'd want to make sure we are all on board with that idea.

PR #5, opened 2016-01-20
Remove .idea folder

cdeepakroy:


cooperlab:
I'm going to take another look at the .gitignore and see why these are being imported.

cdeepakroy:
@cooperlab i think we have to remove the folder using git rem since it was already checked in to the repo. I did this and added .idea to .gitignore. Seems to have worked. I'll merge it.

cdeepakroy:
I've tried to fix the installation of numpy and scipy using miniconda while using tox. But brian's [pull request](https://github.com/DigitalSlideArchive/HistomicsTK/pull/9) that is setting up CMake/CTest based testing is going to solve this issue too. So i reverted my changes. Once, we merge that pull request, this will be fixed hopefully.

PR #9, opened 2016-01-20
Convert to using Girder for testing

brianhelba:


brianhelba:
This works, but it's way too slow to build all the Python prereqs (numpy, scipy, etc.).

I'm going to switch it to use Miniconda.

cdeepakroy:
@brianhelba was facing the same issue today and stumbled on a similar [solution](http://lmjohns3.com/2015/06/using-travis-ci-with-miniconda-scipy-and-nose.html).

brianhelba:
Yup, that is pretty similar to what I have.

cdeepakroy:
@brianhelba should i merge this and proceed?

brianhelba:
I still need to update the Travis script to install Romanesco.

We *could* merge it now, as the build succeeds, it's just the tests that fail: style tests are failing for obvious reasons, and the functional tests fail because of a missing Romansco.

brianhelba:
I've just updated this to add Romanesco. Let's see how it goes.

I'll merge once it's stable.

brianhelba:
Merged.

cdeepakroy:
@brianhelba do you think it would be a good idea to put the tests for the rest end points and the actual toolkit in separate folders? Or put everything in the plugin_tests folder?

PR #16, opened 2016-02-02
Fix style test issues and add style testing to examples

brianhelba:


brianhelba:
@cooperlab @cdeepakroy This makes the tests green. Future PRs should not be merged until they also pass all tests.

PR #22, opened 2016-02-04
Cooper

cooperlab:
Imports in the form "from . import package" were not working so I replaced these.

I also simplified the SimpleMask to pass the max McCabe complexity testing. A new function EstimateVariance was created from this.


brianhelba:
@cooperlab Looks good so far. As you can see from the Travis CI results, there are 4 places that need to have their style fixed.

Also, you should give your pull requests a better, more descriptive name. If you have multiple commits on a branch, GitHib will use the branch name as a default, but you can change this when creating the pull request.

Finally, you may also want to give your branches more descriptive names (though this is less important). Generally something like ``gaussian-voting`` would be fine.

cooperlab:
Thanks for the feedback - so it sounds like I should initiate a new branch (with a descriptive name) each time I add functionality - makes sense.

I fixed the style issues on my branch, so will the pull request update to the new files, or do I have to initiate again?

-Lee

brianhelba:
> so it sounds like I should initiate a new branch (with a descriptive name) each time I add functionality

Yes.

> will the pull request update to the new files, or do I have to initiate again

As long as you push your updates to the same branch name, the PR will update (and as you can see, it's now passing).

This LGTM ("looks good to me" = the magic phrase indicating approval of a PR), so feel free to hit the Merge button at your convenience. You should also press the button (which shows up after you merge) to delete the original branch from the server.

cdeepakroy:
LGTM :).

PR #41, opened 2016-02-29
Auto create REST end points for slicer clis

### LINKED ISSUE #33: Write code to automatically create REST routes for slicer execution model CLIs

cdeepakroy:
This is a PR to fix #33.

This is my first implementation of the functionality to automatically create REST end points for [slicer execution model CLI](https://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/SlicerExecutionModel#Module_Description). 

For each CLI, i create: 
1. A POST route that runs the CLI  and puts the results on girder 
2. A GET route that returns the xml spec of the CLI which can be used by the web-client to automatically generate the UI for driving the CLI. 

The code to automatically generate the REST end points is in [server/__init__.py](https://github.com/DigitalSlideArchive/HistomicsTK/blob/AutoCreateRESTEndPointForSlicerCLI/server/__init__.py#L463). 

I had to make a few assumptions which we can debate about:

- CLIs are located in subdirectories under the server directory. For each CLI, the name of the xml file, the python script, and the subdirectory in which they are placed should be same. You can creating more subdirectories within subdirectories to organize the CLIs into groups. 
- Inputs of type image, file, or directory are located in girder. [Other storage modes or ways of binding input data](http://romanesco.readthedocs.org/en/latest/api-docs.html#the-input-specification) such as mongodb are not supported for these types. When you make a REST call to run the CLI, i ask for the IDs of the item (if image/file) or folder (if directory) in girder. 
- Inputs other than image, file, or directory should be passed inline via the REST call.
- Outputs should only be of type image, file, or directory. 
- Outputs will be stored on girder. In the REST call to run the CLI, i ask for the names of each output and the girder ID of the parent folder 
- Analysis code is assumed to be an in-line python script i.e. Python script is read into a string which is then placed into the task spec of girder. In the future, i plan to develop something to support C++ and Standalone python CLIs via docker.

We need to discuss and decide where to put the functionality that generates the REST end points. Should we make it a plugin of girder say "slicer_execution_model_io"?

I also need some help refactoring it to make it more flexible. This should at some point in the future allow us to put ITK, TubeTK, or any analysis toolkit we develop on the web.


cdeepakroy:
@zachmullen can you please take a look? It now seems to be working. Once we think it is good, i will try to schedule a meeting with everyone to discuss the assumptions i made.

cdeepakroy:
@jbeezley This functionality is now working. There are now two CLIs in the server directory. You can use these as samples to work on the web front end.

jbeezley:
One thing that I would find useful is to respond from `GET /HistomicsTK/<algorithm>` is girder worker task spec as JSON rather than the original XML.  The client side I'm writing could be come more generally useful as a set of standard widgets for girder_worker io types.

Also, a `GET /HistomicsTK` to list the available algorithms could be useful.

One more thing, I noticed in the XML files.  Colors are described as `number_list` in the task spec or `double-vector` in the XML.  It would be much better ask for a color explicitly so that we can generate an actual color picker UI rather than a generic vector.  I'm not sure the extent to which this is possible with the current XML schema.

cdeepakroy:
@jbeezley Nice suggestion! I can add the route `GET /HistomicsTK/<algorithm>` if that makes things easier. I am assuming you do not need the input/output binding specs right?

One thing that made me lean towards creating a GET route that returns the xml is that the girder_worker task spec does not support parameter grouping. For example look at the xml spec of [this](https://github.com/KitwareMedical/TubeTK/blob/master/Applications/ConvertTubesToTubeTree/ConvertTubesToTubeTree.xml) CLI in TubeTK. It has two parameter groups - IO and Algorithm parameters. See the parameters groups (on the left) in the following UI that Slicer automatically generates for this xml.

converttubestotubetree

The grouping helps avoid cluttering the UI when there are a large number of parameters. Also, certain parameters are meant to be tweaked only by advanced users. So Advanced parameter groups should be hidden by default and shown only when the user explicitly asks for it by clicking some button.

If we rely on the girder_worker task spec for the UI then you would not have this information as it does not support grouping of parameters. Can we add this feature to the girder_worker task spec easily?

This is something we should discuss with more people maybe.

jbeezley:
That makes sense.  We could talk about an expanded task spec that brings these ideas into girder_worker as well.  For now, it shouldn't be a problem to parse the XML directly.

cdeepakroy:
@jbeezley Regarding the color, you make a good suggestion. I dont think both the slicer xml and girder_worker task spec have an explicit notion of color. We will have to negotiate with folks of both specs to add it.

cdeepakroy:
@jbeezley i have added a GET route that returns a list of relative routes to all CLIs

image

server/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -1,31 +1,42 @@
-import os
 import numpy as np
 import skimage.io
 import histomicstk as htk
 
-# Define Girder Worker globals for the style checker
+# Define Romanesco globals for the style checker
```
zachmullen: Aren't we using girder_worker rather than romanesco now?

server/SparseColorDeconvolution/SparseColorDeconvolution.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
+import skimage.io
+import histomicstk as htk
+
+# Define Romanesco globals for the style checker
```
zachmullen: and here too

server/__init__.py
```diff
@@ -1,136 +1,564 @@
...
+    xmlFile : str
+        Full path to xml file of the CLI
+    scriptFile : str
+        Full path to .py file containing the code of the clu
+
+    Returns
+    -------
+    function
+        Returns a function that runs the CLI using girder_worker
+    """
```
zachmullen: We should use sphinx docstrings like we do in the rest of girder. So this would be:

```
    """Generates a handler to run CLI using girder_worker

    :param restResource: The object of a class derived from girder.api.rest.Resource to which
        this handler will be attached
    :type restResource: `girder.api.rest.Resource`     
    :param xmlFile: Full path to xml file of the CLI
    :type xmlFile: str
    :param scriptFile: Full path to .py file containing the code of the clu
    :type scriptFile: str
    :returns: A function that runs the CLI using girder_worker
    :rtype: callable
    """

PR #48, opened 2016-03-11
Parameterizes tile coverage, adds exception

cooperlab:
Made the minimum tile coverage a parameter ('Coverage') for flexibility, and set a low default value to minimize cases that generate an empty sample.

Added an exception to catch the event of an empty sample.


brianhelba:
@cooperlab Per my email to you, please re-base and re-push this branch and re-open this PR. The build should succeed now.

PR #54, opened 2016-03-23
LoG seeding

cooperlab:
Added additional functions for nuclei seeding. cLoG implements the distance-transform constrained LoG filter. gLoG implements the generalized LoG filter that integrates responses over various scales and orientations of generalized elliptical LoG filters. The local maxima of the responses produced by cLoG and gLoG are candidates for nuclear seeds.


cooperlab:
Added two new Laplacian of Gaussian functions for nuclear seeding.

Issue #37, opened 2016-02-25
Develop front-end for DSA 


cdeepakroy:
The front end should have ways to:

- select data from girder
- viewer for visualizing whole-slide images and annotations/results
- select and run analysis CLIs exposed in the server directory of histomicsTK

cdeepakroy:
Below is a sample POST request that runs a CLI that does ColorDeconvolution

`api/v1/HistomicsTK/ColorDeconvolution/run?itemId=56279eb3a8487354347916c6&folderId=56279ecea8487354347916cd&stainColor_1=%5B0.65%2C%200.70%2C%200.29%5D&stainColor_2=%5B0.07%2C%200.99%2C%200.11%5D&stainColor_3=%5B0%2C%200%2C%200%5D`

and below is the response:

```
{
  "_accessLevel": 2,
  "_id": "56cf65fba8487359caad8382",
  "_modelType": "job",
  "args": [],
  "async": false,
  "celeryTaskId": "f09e40be-306d-49e8-8feb-8807122f1dfc",
  "created": "2016-02-25T20:37:15.682372+00:00",
  "interval": 0,
  "kwargs": {
    "auto_convert": true,
    "cleanup": true,
    "inputs": {
      "inputImageFile": {
        "format": "string",
        "host": "localhost",
        "id": "56279eb3a8487354347916c6",
        "mode": "girder",
        "name": "A.png",
        "port": 8080,
        "resource_type": "item",
        "token": "8kFFXZqZfajQU4yuVvKnz0hgqfEG4mw4vSRe71QO6ZWselKZXnZFjRDiEZlX3HXh",
        "type": "string"
      },
      "stainColor_1": {
        "data": "[0.65, 0.70, 0.29]",
        "format": "json",
        "mode": "inline",
        "type": "number_list"
      },
      "stainColor_2": {
        "data": "[0.07, 0.99, 0.11]",
        "format": "json",
        "mode": "inline",
        "type": "number_list"
      },
      "stainColor_3": {
        "data": "[0, 0, 0]",
        "format": "json",
        "mode": "inline",
        "type": "number_list"
      }
    },
    "jobInfo": {
      "headers": {
        "Girder-Token": "CTlsh1TOxaVtOGTQgX0QyC5FLlU9KF2hPv9N5t72z44iuFw7wreGnmVesa9c4AIq"
      },
      "logPrint": true,
      "method": "PUT",
[...]

PR #56, opened 2016-03-25
Initial web client

jbeezley:
Even though this is not yet a complete implementation, it is probably time to merge this branch in.  The primary intent of the code here is to:
* [parse](https://github.com/DigitalSlideArchive/HistomicsTK/blob/75e9c9cd091e4fe6878156890980140fcec12dee/web_client/js/schema/parser.js) Slicer XML specs
* [generate](https://github.com/DigitalSlideArchive/HistomicsTK/blob/75e9c9cd091e4fe6878156890980140fcec12dee/web_client/js/models/widget.js) intermediate json objects (backbone models)
* create a series of [panels](https://github.com/DigitalSlideArchive/HistomicsTK/blob/75e9c9cd091e4fe6878156890980140fcec12dee/web_client/js/views/controlsPanel.js) each containing multiple [control widgets](https://github.com/DigitalSlideArchive/HistomicsTK/blob/75e9c9cd091e4fe6878156890980140fcec12dee/web_client/js/views/controlWidget.js), supporting:
  * live updating of their respective models
  * data validation

All of that is implemented and tested with close to complete coverage.  There is other code in here to provide minimal infrastructure that remains untested.

Since I last demoed this to others, I have added a page that renders a more complete example of the different kind of widgets you can create.  This is currently hard coded to display at `/histomicstk#demo`.  The json it is rendering is stored in the repo [here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/75e9c9cd091e4fe6878156890980140fcec12dee/web_client/extra/widget_demo.json).  The spec is roughly equivalent the XML form.

I also made it so the background is rendered with geojs (still cats though).

Things that are missing:
* `Submit` only writes the parameter values to the console when valid rather than running the job
* A few parameter types (notably `directory` for the current analyses)
* Persistence of user provided values (through local storage)


cooperlab:
Looks good - timing is good as far as I'm concerned.


Lee Cooper, Ph.D.
lee.cooper@emory.edu
tel: (404) 712-0110


cdeepakroy:
@jbeezley great work! 

You should demo this in the next meeting. Then, we should probably discuss and sketch out how the front-end should look like in the next couple of meetings.

I tried to git checkout your banch and see the front-end through the url `http://localhost:8080/histomicstk#demo` but it doesnt show anything. Am i missing anything?

This is probably coming next. But, your parser doesn't seem to inspect the channel element of the parameter to find whether it is input or output. For output file we have to probably ask for the parent folder in girder and the name of the file.  All non-file outputs will be stored as `key = value` pairs in a `<name>.params` file. If the xml spec has any non-file outputs we have to ask for the name and parent folder of this file in girder too. Also in the parser you could look for the `fileExtensions` attribute of the parameter to filter the files in the file selection widget.

jbeezley:
@cdeepakroy You probably need to rebuild (`npm install` in girder's main directory).

Yes, I'll probably work on the hierarchy widget next to handle input/output as well as image/file/directory distinctions. Non-file outputs will probably need to be handled specially by the girder job.  Possibly, it could find the file somehow and put the parameters into the job model in the database.  We'll have to discuss how to handle it.

cdeepakroy:
@jbeezley i did `npm install` but it still does not show anything. Also updated my girder.

cdeepakroy:
@jbeezley Added large image to HistomicsTK dependencies.

LGTM. Please go ahead and merge this.

cdeepakroy:
Fixes #37

PR #61, opened 2016-03-31
Fix some tricky bugs with incompatable Python package versions

brianhelba:


brianhelba:
This should make everything compatible with large_image, and deal with some bugs in with SciPy too. Please don't change the versions of these packages any time in the near future without letting me know.

requirements_c.txt
```diff
@@ -4,7 +4,12 @@
 #
 #########################################################
 nimfa==1.2.3
-numpy==1.10.4
-openslide-python==1.1.0
```
cdeepakroy: There are still functions in core HistomicsTK that use openslide. Maybe leave it in there?

.travis.yml
```diff
@@ -47,6 +47,7 @@ before_install:
     - conda config --add channels https://conda.binstar.org/cdeepakroy
 
 install:
+    - conda install --yes libgfortran==1.0 numpy==1.10.2 scipy==0.16.0 scikit-image==0.12.3
```
cdeepakroy: Add openslide-python==1.1.0 to this?

I built it for conda and put it in this channel -- [https://conda.anaconda.org/cdeepakroy](https://conda.anaconda.org/cdeepakroy)

PR #63, opened 2016-04-04
Large image integration

jbeezley:
This adds a simple control widget to select a Girder item to display.  Either a "large_image" item or an item containing an image file can be selected, and it will appear in the visualization panel rendered by GeoJS.

* It requires the latest master of the large_image plugin.
* The control widget will only let you select items under your user in Girder rather than an arbitrary collection.

### DIFF
--- a/.install-openslide.sh
+++ b/.install-openslide.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+PREFIX="$CACHE/openslide-$OPENSLIDE_VERSION"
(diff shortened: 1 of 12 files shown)


brianhelba:
Lets discuss this in-person tomorrow before merging.

cdeepakroy:
@jbeezley Nice! I could get it to view a simple png image. But when i tried to view this [SVS image](https://data.kitware.com/#item/56b331718d777f5d6174259a) it throws an error.

large_image_integration

jbeezley:
You probably need to initialize it first by posting to `/item/{itemId}/tiles`.

brianhelba:
LGTM.

PR #69, opened 2016-04-07
Set default root path to /collections/TCGA

jbeezley:


brianhelba:
LGTM. Merging and deploying.

PR #73, opened 2016-04-14
Add new widget types for directories and output files

jbeezley:
For file outputs, I created a text `<input>` below the hierarchy widget for the user to type in the new item name.

screen shot 2016-04-14 at 2 04 25 pm

There is minimal validation done when submitting the modal.

screen shot 2016-04-14 at 1 57 22 pm

The main control widget now shows a path string rather than an object id.

screen shot 2016-04-14 at 1 58 07 pm

Finally, the submit button actually `POST`'s to the analysis endpoint.  I think it is submitting correctly, but I haven't been able to get the celery working running correctly locally yet.

### DIFF
--- a/plugin_tests/client/parser.js
+++ b/plugin_tests/client/parser.js
@@ -86,6 +86,7 @@ describe('XML Schema parser', function () {
                 slicerType: 'integer',
                 id: 'foo',
                 title: 'arg1',
+                channel: 'input',
                 description: 'An integer'
             });
         });
@@ -104,6 +105,7 @@ describe('XML Schema parser', function () {
                 slicerType: 'string',
                 id: 'foo',
                 title: 'arg1',
+                channel: 'input',
                 description: 'A description'
             });
         });
@@ -121,6 +123,7 @@ describe('XML Schema parser', function () {
                 type: 'boolean',
                 slicerType: 'boolean',
                 id: 'foo',
+                channel: 'input',
                 title: 'arg1',
                 description: 'A description'
             });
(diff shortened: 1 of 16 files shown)


zachmullen:
Minor nit, it looks like in your Select button, you put the text inside the `<i>` tag rather than as a sibling of it ;)

jbeezley:
Fixed.  Also, I change the button text to `Save`.

cdeepakroy:
@jbeezley To test, I git cloned your branch (directory-io), went to the front-end page at `http://localhost:8080/histomicstk#ColorDeconvolution`, and tried to set `Output Image of Stain 1`. But i could not see your new widget. See snapshot below:

new_widget

Do i have to do anything else to get to see the new widget?

jbeezley:
Did you rerun `npm install`?

cdeepakroy:
@jbeezley i could see the new widget after running npm install. If i hit submit, is it supposed to run the job, or that is not implemented yet?

jbeezley:
It should POST the analysis to the server.  When I tried, I got an internal server error, which I assumed was because my worker isn't setup correctly, but there could be something else going wrong.  I'll look into it more a bit later.

cdeepakroy:
I tried running it using the submit button but it dint seem to do anything --- no new job was initiated at `http://localhost:8080/#jobs/` and saw nothing on the girder or girder_worker command-line.

jbeezley:
I just fixed a bug that may have caused silent validation errors and added an error dialog to the form submission in case something goes wrong. You will need to`npm install` again (as always with client code updates).

cdeepakroy:
@jbeezley It now runs the CLI successfully. 

LGTM! Can be merged.

PR #74, opened 2016-04-18
Code to generate REST end point from standalone CLIs via docker

cdeepakroy:
@slee172  please review code.


.dockerignore
```diff
@@ -0,0 +1,23 @@
+**/*.pyc
+**/__pycache__/
+**/*#
+**/.#*
+**/*~
```
cdeepakroy: Change this to *#

PR #78, opened 2016-04-20
Modify Auto CLI REST endpoint code to run standalone CLIs using Docker in girder_worker

cdeepakroy:
This pull request
* modifies the code for automatic REST end point generation of slicer CLIs to
generate REST end points from a list of docker images
* converts all existing CLIs from inline python scripts (code passed as string to girder_worker task) to standalone python CLIs that can be run on command-line.
* creates a docker image for HistomicsTK called `dsarchive/histomicstk` with an entry point that 
  * **lists the CLIs the docker container has to offer** (developer has to list these in a file called `slicer_cli_list.json`) by running `docker run dsarchive/histomicstk --list_cli`, and 
  * **lets you run any of the CLIs in the list** by running `docker run dsarchive/histomicstk <cli-rel-path>`. 

Developers are free to write CLIs and put them in a their own git repository. All they have to do is write a Dockerfile, generate a docker image that supports the aformentioned interface (to list the CLI and run any CLI), and upload it to dockerhub. If they want to use the functionality in HistomicsTK they have to base their Dockerfile out of dsarchive/histomicstk using `FROM dsarchive/histomicstk`.


cdeepakroy:
Fixed #67

PR #82, opened 2016-04-25
Feature extraction

slee172:
- Update __init__.py
- Update FeatureExtraction.py


histomicstk/FeatureExtraction.py
```diff
@@ -65,25 +58,23 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
     CytoplasmSumCanny	CytoplasmMeanCanny
-
     Boundary
     --------
     Boundaries: x1,y1;x2,y2;...
-
-
-
     ##########     Features End     ##########
-
```
cdeepakroy: @slee172 can you add a Reference section with a paper for FSD features?

cdeepakroy: [Here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/histomicstk/SparseColorDeconvolution.py#L51-L57) an example to learn how to add a Reference section

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +144,157 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+def FSDs(X, Y, K, Intervals):
+    """
+    Calculated FSDs from boundary points X,Y. Boundaries are resampled to have
+    K equally spaced points (arclength) around the shape. The curvature is
+    calculated using the cumulative angular function, measuring the
+    displacement of the tangent angle from the starting point of the boundary.
+    The K-length fft of the cumulative angular function is calculated, and
+    then the elements of 'F' are summed as the spectral energy over
+    'Intervals'.
+
```
cdeepakroy: Add documentation/docstring for the inputs

histomicstk/__init__.py
```diff
@@ -62,4 +63,5 @@
            'SimpleMask',
            'SparseColorDeconvolution',
            'SubmitTorque',
-           'TilingSchedule')
+           'TilingSchedule',
+           'FeatureExtraction')
```
cdeepakroy: Add FeatureExtraction in alphabetical order so it is easy to find?

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +144,157 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+    Returns F - length(Intervals) vector containing spectral energy of
+    cumulative angular function, summed over defined 'Intervals'.
+    """
```
cdeepakroy: Here might be a good place to add a reference to paper describing FSD features.

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +144,157 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+    Output = iXY(iX, iY)
+
+    return Output
+
+
+def log2space(a, b, n):
+    """
+    Returns n logarithmically-spaced points from 2^a to 2^b.
+    """
+
```
cdeepakroy: I think you can use the [linspace](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.linspace.html) function for this -- 
`2**np.linspace(a, b, n, endpoint=True)` or 
`np.power(2, np.linspace(a, b, n, endpoint=True)`

histomicstk/__init__.py
slee172: ok.

histomicstk/FeatureExtraction.py
slee172: Good point. Thanks !!

histomicstk/FeatureExtraction.py
```diff
@@ -115,6 +112,24 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+        )
+        # calculate FSDs
+        FSD = FSDs(
+            Bounds[:, 0], Bounds[:, 1],
+            K, Interval
+        )
+        FSDGroup = np.append(FSDGroup, FSD)
 
     # add columns to dataframe
     df['X'] = CentroidX
```
cdeepakroy: Don't u want to add the X and Y coordinates as separate columns/features?

slee172: Can be stored in one column or feature, but I think a feature is usually one value. How about discuss with Lee?

cdeepakroy: CentroidX is an array of two values right?

slee172: CentroidX is a list of x points that indicate a center of each object, so it could have two values if there're two objects in an image. For example, if there is an image labeled with '1' and '2' for two different objects, CentroidX of the first object labeled with '1'  will be "51.3231" (for instance), while another CentroidX of the second object labeled with '2' will be "56.1515". As far as I know, one object will have only one value of CentroidX.

histomicstk/__init__.py
slee172: updated

histomicstk/FeatureExtraction.py
slee172: updated

histomicstk/FeatureExtraction.py
cdeepakroy: My bad! I now see that you have two separate features/columns for the x-coordinate (CentroidX) and y-coordinate (CentroidY). All good on this :)

histomicstk/FeatureExtraction.py
slee172: Thanks, the reference is added.

histomicstk/FeatureExtraction.py
slee172: added

histomicstk/FeatureExtraction.py
slee172: done

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +187,325 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+    df['HematoxlyinMeanGradMag'] = HematoxylinGradientGroup.MeanGradMag
+    df['HematoxlyinStdGradMag'] = HematoxylinGradientGroup.StdGradMag
+    df['HematoxlyinEntropyGradMag'] = HematoxylinGradientGroup.EntropyGradMag
+    df['HematoxlyinEnergyGradMag'] = HematoxylinGradientGroup.EnergyGradMag
+    df['HematoxlyinSkewnessGradMag'] \
+        = HematoxylinGradientGroup.SkewnessGradMag
+    df['HematoxlyinKurtosisGradMag'] \
+        = HematoxylinGradientGroup.KurtosisGradMag
+    df['HematoxlyinSumCanny'] = HematoxylinGradientGroup.SumCanny
+    df['HematoxlyinMeanCanny'] = HematoxylinGradientGroup.MeanCanny
```
cdeepakroy: @slee172 Instead of explicity setting each feature with `df[feature_name]=feature_value` statements, you could use a loop as follows:

```
for feature_name in HematoxylinIntensityGroup._fields:
    df['Hematoxylin' + feature_name] = getattr(HematoxylinIntensityGroup, feature_name)
```
This would make the code compact and a bit more cleaner in the sense that in the future you can add a feature within the feature group function and not have to change the root function.

You can do the same or something similar (using a loop) for other feature groups (FSD, Cytoplasm etc) as well.

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +187,325 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+        Entropy, Energy, Skewness, Kurtosis
+    )
+
+    return Output
+
+
+def IntensityFeatureGroup(I, Coords):
+    """
+    Get IntensityFeatures for nuclei and cytoplasms
+    """
```
cdeepakroy: Define the docstring like [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/histomicstk/ColorDeconvolution.py#L18-L26) for all inputs and like [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/histomicstk/ColorDeconvolution.py#L28-L41) for all the return values.

Making this a habit will make your functions easy to understand and eventually more used by people.

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +187,325 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+    Calculated FSDs from boundary points X,Y. Boundaries are resampled to have
+    K equally spaced points (arclength) around the shape. The curvature is
+    calculated using the cumulative angular function, measuring the
+    displacement of the tangent angle from the starting point of the boundary.
+    The K-length fft of the cumulative angular function is calculated, and
+    then the elements of 'F' are summed as the spectral energy over
+    'Intervals'.
+
+    Returns F - length(Intervals) vector containing spectral energy of
```
cdeepakroy: Define return values like [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/histomicstk/ColorDeconvolution.py#L28-L41) in the doc string.

Also add docstrings for inputs/parameters. 

Do this for all functions.

histomicstk/FeatureExtraction.py
cdeepakroy: I would advise you to get into the habit of first writing the docstring for the function before you start to write the logic. That way in your mind you will first think of 
* the inputs/parameters the function should take
* the outputs should return
* what the function does

cdeepakroy:
@slee172 nice work !!! Few minor comments !

cooperlab:
We just reviewed this in my office - looks very good. Is fairly specific to nuclear morphometry right now but will evolve towards a more general feature extraction library.


Sanghoon is going to change the interface to provide the color deconvolved intensity images instead of the color image.

histomicstk/FeatureExtraction.py
slee172: Thanks. That is what I wanted

histomicstk/FeatureExtraction.py
slee172: Thank you for the tips !!

histomicstk/FeatureExtraction.py
slee172: will be done.

histomicstk/FeatureExtraction.py
```diff
@@ -129,6 +177,410 @@ def FeatureExtraction(Label, I, K=128, Fs=6, Delta=8):
...
+        Skewness of gradient data. Value is 0 when all values are equal.
+    KurtosisGradMag : float
+        Kurtosis of gradient data. Value is -3 when all values are equal.
+    SumCanny : float
+        Sum of canny filtered gradient data.
+    MeanCanny : float
+        Mean of canny filtered gradient data.
+    Notes
+    -----
+    Return values are returned as a namedtuple.
```
cdeepakroy: @slee172 For all places where you return a named tuple, can you please add the docstring for the components of the tuple?

[Here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/histomicstk/ColorDeconvolution.py#L30-L41) is an example.

Also merge changes from the master repo using `git merge origin master` and resolve the conflicts. Currently it says that the code in your branch has some conflict with the code in the master repo. This could be because both modified the same file. You will have to go into the conflicted files and merge (it will show both your code and the code in the master for each conflict. You will have to select the right one and remove the wrong one) the changes one by one.

Rest all looks good ! Great job !

slee172: Hi Deepak,

I really appreciate your valuable comments all. Could you let me know what
the components of the tuple are? Does that mean types of values (e.g. float
or array-like)?

best,


Sanghoon Lee


2016-05-09 11:36 GMT-04:00 Deepak Roy Chittajallu <notifications@github.com>
:

histomicstk/FeatureExtraction.py
```diff
@@ -356,15 +356,15 @@ def IntensityFeatureGroup(I, Coords):
...
         Difference between mean and median.
-    MaxIntensity : float
+    MaxIntensity : array_like
         Max intensity data.
-    MinIntensity : float
+    MinIntensity : array_like
         Min intensity data.
-    StdIntensity : float
+    StdIntensity : array_like
         Standard deviation of intensity data.
```
cdeepakroy: Great! This is what i was suggesting.

histomicstk/FeatureExtraction.py
slee172: the conflict is now resolved.

cdeepakroy:
@slee172 Looks good to me ! Please merge. Thanks a lot for your contribution!

PR #92, opened 2016-05-19
Mixture model

cooperlab:
Adds functionality for generating smooth binary masks from probabilistic models using mixture modeling and graph cuts.

### DIFF
--- a/histomicstk/BinaryMixtureCut.py
+++ b/histomicstk/BinaryMixtureCut.py
@@ -1,0 +1,77 @@
+import numpy as np
+import pygco as gc
+
+
+def BinaryMixtureCut(Foreground, Background, I=None, Sigma=25):
+    """Performs a binary graph-cut optimization for region masking, given
+    foreground and background probabilities for image and
+
+    Parameters
+    ----------
+    I : array_like
+        An intensity image used for calculating the continuity term. If not
+        provided, only the data term from the foreground/background model will
+        drive the graph-cut optimization. Default value = None.
+    Foreground : array_like
+        Intensity image the same size as 'I' with values in the range [0, 1]
+        representing probabilities that each pixel is in the foreground class.
+    Background : array_like
+        Intensity image the same size as 'I' with values in the range [0, 1]
+        representing probabilities that each pixel is in the background class.
+    Sigma : double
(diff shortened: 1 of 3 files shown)


histomicstk/BinaryMixtureCut.py
```diff
@@ -0,0 +1,77 @@
...
+                 -np.log(Foreground + Small) / -np.log(Small)),
+                 axis=2)
+
+    # formulate pairwise label costs
+    Pairwise = 1 - np.eye(2)
+
+    # formulate vertical and horizontal costs if 'I' is provided
+    if I is not None:
+        H = np.exp(np.abs(I[:, 0:-1].astype(np.float) -
+                   I[:, 1:].astype(np.float)) / (2 * Sigma**2))
```
cdeepakroy: @cooperlab should this be:

image

cdeepakroy: The way you have it looks like weight would be higher when the gradient (or intensity difference) is high.

cdeepakroy: See equation 7 in http://www.csd.uwo.ca/~yuri/Papers/ijcv06.pdf

histomicstk/PoissonMixture.py
```diff
@@ -0,0 +1,89 @@
...
+        defined as 'None'. Default value = None.
+
+    Returns
+    -------
+    Tau : double
+        Optimal threshold for distinguishing foreground and background.
+    Foreground : array_like
+        An intensity image with values in the range [0, 1] representing
+        foreground probabiities for each pixel.
+    Foreground : array_like
```
slee172: @cooperlab this seems like to be 'Background'

cooperlab:
Docstring and edge-weight calculation fixed (that was an important minus sign!)

histomicstk/BinaryMixtureCut.py
cdeepakroy: @cooperlab should we also square the result of np.abs() for vertical and horizontal weights? Then it would give us guassianish fade out of the weight. And the sigma becomes a bit mote intuitive to set, since we know for a gaussian any intensity difference beyond `3 * sigma` pretty much will result in a weight of zero.

PR #93, opened 2016-05-20
Update BinaryMixtureCut.py

cooperlab:
Added missing square operator in edge weight numerator calculation.


cooperlab:
Fixed mathematical errors with edge weight.

Issue #86, opened 2016-05-10
Implement CLI for segment nuclei and generate annotations for them


cdeepakroy:

cdeepakroy:
@cooperlab I want to start working on this issue. Can you point me and @slee172 to the function to use for nuclei segmentation?

Also which of the two color deconvolution (standard/sparse) was working better so i could use that?

cooperlab:
@cdeepakroy use the standard color deconvolution.

The function is really a sequence: normalize color -> deconvolve to hematoxylin (H) & eosin (E) -> mask -> cLoG response -> max clustering -> feature extraction.

I will write an example script if I get some free time tomorrow.

cdeepakroy:
This is fixed by PRs #94 and #102.

Issue #65, opened 2016-04-06
Alpha expansions w/ graph coloring - implementation


cooperlab:
Refinement of label images using multi-label pygco graph cuts (alpha-expansions or alpha-beta swap). Graph coloring is used to reduce the computational burden so that # objects < # cells.

cooperlab:
Partially resolved by pull request #84 - implementation of (graph coloring and region adjacency graphs)

PR #94, opened 2016-05-23
ENH: Add CLI to segment nuclei and generate annotations of nuclei bounding boxes

cdeepakroy:


cdeepakroy:
Fixes #86

cooperlab:
Looks great - we can insert the label image refinement when I finish issue #65

cdeepakroy:
@brianhelba Can we bump the commit hash of girder, girder_worker, and large_image?

Bumping up large_image seems to have an issue with Pillow versions. See [here](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/132399102#L839)

slee172:
@cdeepakroy current MaxClustering at the master branch receives local_max_search_radius parameter as int, but it seems like that you're going to set it to double. Is it ok if updated the local_max_search_radius as double in FeatureExtraction.xml ?

cdeepakroy:
@slee172 I kept it as double to leave it to be a bit more general. You can use the same and we can change it back later if we want.

cdeepakroy:
@jbeezley Integrated changes from your PR to this #97 to test.

The following REST call generated from the swagger interface works:

> http://localhost:8080/api/v1/HistomicsTK/NucleiSegmentation/run?inputImageFile_girderItemId=573cc705a848737a396e529e&outputNucleiMaskFile_girderFolderId=573cc72ca848737a396e52a0&outputNucleiMaskFile_name=Easy1_seg.png&outputNucleiAnnotationFile_girderFolderId=573cc72ca848737a396e52a0&outputNucleiAnnotationFile_name=Easy1_nuclei.anot&foreground_threshold=160.0&local_max_search_radius=10.0&max_radius=7.0&min_nucleus_area=80.0&min_radius=4.0&stain_1=%22hematoxylin%22&stain_2=%22eosin%22&stain_3=%22null%22

See how stain_1, stain_2, and stain_3 are of type string-enumeration are passed.

After adding the code to catch the json.loads exception and print the param name + value + type, I tried to run this CLI via the front-end and the error i sent you was generated by parameter stain_1. Below is the error printed out by [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/nuclei_segmentation_cli/server/rest_slicer_cli.py#L458-L466) exception handling code:

```
Error: Parameter value is not in json.dumps format
  Parameter name = stain_1
  Parameter type = string-enumeration
  Value passed = hematoxylin
```

Can we print the REST call made when the submit button is pressed in your front-end code? Then we should be able to find where the problem is.

jbeezley:
If you open the network tab in Chrome's developer tools, you should see the request made.  I attached a screenshot of what I'm getting.  It looks to me like the strings are correctly JSON encoded.  I suspect you aren't getting the changes from my branch from some reason.

screen shot 2016-05-25 at 1 17 26 pm

cdeepakroy:
@jbeezley it works now. I had to restart girder and girder_worker. And redo npm install again. 

You can merge PR #97 now.

cdeepakroy:
Thanks for the help!

PR #96, opened 2016-05-24
First implementation of feature extraction CLI

slee172:
This is the initial implementation of feature extraction for CLI.

Added two files: FeatureExtraction.py and FeatureExtraction.xml
Modified server/slicer_cli_list.json to add feature extraction

FeatureExtraction.py reads an input image, stains, and some parameters (K, Fs, and Delta) and extracts features from the image. The features are set on a panda dataframe, and then finally stored on a hdf5 format. 

@cooperlab The input image is manually resized for the test at this time, but will be modified later.

@cdeepakroy The CLI works well on the command line, but needs to be tested on the web front.

### DIFF
--- a/server/FeatureExtraction/FeatureExtraction.py
+++ b/server/FeatureExtraction/FeatureExtraction.py
@@ -1,53 +1,105 @@
+from ctk_cli import CLIArgumentParser
+import histomicstk as htk
 import numpy as np
-import pandas
-import scipy
+import pandas as pd
+import scipy as sp
 import skimage.io
 import skimage.morphology
-
-import histomicstk as htk
-from ctk_cli import CLIArgumentParser
 
 import logging
 logging.basicConfig()
 
+stainColorMap = {
+    'hematoxylin': [0.65, 0.70, 0.29],
+    'eosin':       [0.07, 0.99, 0.11],
+    'dab':         [0.27, 0.57, 0.78],
+    'null':        [0.0, 0.0, 0.0]
+}
 
 def main(args):
 
+    #
     # Read Input Image
+    #
     print('>> Reading input image')
 
-    print '--- Image Name = ', args.inputImageFile
+    imInput = skimage.io.imread(args.inputImageFile)[:, :, :3]
 
-    Standard = skimage.io.imread(args.inputImageFile)[:, :, :3]
-    print '--- Original Image Size = ', Standard.shape
+    #
+    # Perform color normalization
+    #
+    print('>> Performing color normalization')
 
-    # Create stain matrix
-    print('>> Creating stain matrix')
-    W = np.array([args.stainColor_1, args.stainColor_2, args.stainColor_3]).T
+    # transform input image to LAB color space
+    imInputLAB = htk.RudermanLABFwd(imInput)
 
+    # compute mean and stddev of input in LAB color space
+    Mu = np.zeros(3)
+    Sigma = np.zeros(3)
+
+    for i in range(3):
+        Mu[i] = imInputLAB[:, :, i].mean()
+        Sigma[i] = (imInputLAB[:, :, i] - Mu[i]).std()
+
+    # perform reinhard normalization
+    imNmzd = htk.ReinhardNorm(imInput, Mu, Sigma)
+
+    #
     # Perform color deconvolution
+    #
     print('>> Performing color deconvolution')
-    UNS = htk.ColorDeconvolution(Standard, W)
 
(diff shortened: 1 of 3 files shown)


server/FeatureExtraction/FeatureExtraction.py
```diff
@@ -0,0 +1,59 @@
...
+    # Read Input Image
+    print('>> Reading input image')
+
+    print '--- Image Name = ', args.inputImageFile
+
+    Standard = skimage.io.imread(args.inputImageFile)[:, :, :3]
+    print '--- Original Image Size = ', Standard.shape
+
+    # Reduce image size for testing
+    Standard = Standard[50:200, 50:200]
```
cdeepakroy: @slee172 Remove this hard coded cropping. Run it on small image files for now. You can always crop a large image outside, save the cropped image, and use that for testing.

server/FeatureExtraction/FeatureExtraction.py
```diff
@@ -0,0 +1,59 @@
...
+    print('>> Performing color deconvolution')
+    UNS = htk.ColorDeconvolution(Standard, W)
+
+    # Constrained log filtering=== - generate R_{N}(x,y)
+    Nuclei = UNS.Stains[::2, ::2, 0].astype(dtype=np.float)
+    Mask = scipy.ndimage.morphology.binary_fill_holes(Nuclei < 160)
+
+    Response = htk.cLoG(Nuclei, Mask, SigmaMin=4*1.414, SigmaMax=7*1.414)
+    Label, Seeds, Max = htk.MaxClustering(Response.copy(), Mask, 10)
+    Filtered = htk.FilterLabel(Label, 4, 80, None).astype(dtype=np.int)
```
cdeepakroy: @slee172 do not hardcode any parameters (SigmaMin, SigmaMax, parameters to MaxClustering and FilterLabel). All these should be exposed as optional input parameters in the xml spec (you can use the values you have hardcoded here as default values for the parameters in the xml spec), so users can tweak it if needed.

Use [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/nuclei_segmentation_cli/server/NucleiSegmentation/NucleiSegmentation.xml#L74-L112) as an example.

server/FeatureExtraction/FeatureExtraction.py
slee172: @cdeepakroy OK. Good idea !!

server/FeatureExtraction/FeatureExtraction.xml
```diff
@@ -0,0 +1,71 @@
...
+      <default>6</default>
+    </integer>
+    <integer>
+      <name>Delta</name>
+      <label>Delta</label>
+      <description>Used to dilate nuclei and define cytoplasm region</description>
+      <longflag>delta</longflag>
+      <default>8</default>
+    </integer>
+  </parameters>
```
cdeepakroy: @slee172 Try grouping the parameters. 

For example you can:

* Put `stainColor_1`, `stainColor_2`, and `stainColor_3` into a group named `Color Deconvolution Parameters`. 
* Put `K`, `Fs`, and `Delta` into a group called `Feature Extraction Parameters`
* Put `inputImageFile` and `outputFile` into a group called `IO`
* Also expose nuclei segmentation parameters and create a separate group for them.

When there are a large number of parameters, grouping helps keep it modular and less overwhelming.

See [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/nuclei_segmentation_cli/server/NucleiSegmentation/NucleiSegmentation.xml#L74-L112) for an example.

Also, if you think a certain group of parameters is meant to be tweaked only by expert users, then you make that group by setting the `advanced` attribute of parameters to `true` as follows:
```
<parameters advanced="true">
</parameters>
```
This will make group panel closed/hidden in the GUI by default.

cdeepakroy:
@slee172 Few comments. Rest looks good to me.

server/FeatureExtraction/FeatureExtraction.py
slee172: done

server/FeatureExtraction/FeatureExtraction.xml
slee172: @cdeepakroy  Thank you so much. Now the parameters are grouped.

cdeepakroy:
@slee172 Looks good to me. Can you confirm that it is working well on command-line?

slee172:
@cdeepakroy works well excepting the issue of the local_max_search_radius because your  MaxClustering is not updated to the master branch yet.

PR #101, opened 2016-06-01
Haralick Feature Extraction

slee172:
Added Haralick feature extraction for 26 features.

The function receives an intensity image, distance of neighboring pixels (default 1 ), and number of gray level co-occurance matrix (defalut 256), and returns a 1 by 26 array indicating 26 features.
 
To first get a gray level co-occurance matrix from the intensity image, skimage.feature.greycomatrix is used, and then computed 13 features (Angular Second Moment, Contrast, Correlation, Sum of Squares, Inverse Difference Moment, Sum Average, Sum Variance, Sum Entropy, Entropy, Variance, Difference Entropy, Information Measures of Correlation 1 and 2) for each angle (0, 45, 90, and 135). And then, means and ranges of the 13 features for the angles are computed, thus generating 26 different features.

The calucation for the last feature named Maximal Correlation Coefficient was not added to the function because it could cause a computational instability. It is refered to MurphyLab (http://murphylab.web.cmu.edu/publications/boland/boland_node26.html)


histomicstk/FeatureExtractionHaralick.py
```diff
@@ -0,0 +1,106 @@
...
+    References
+    ----------
+    .. [1] Haralick, et al. "Textural features for image classification,"
+    IEEE Transactions on Systems, Man, and Cybernatics, vol. 6, pp: 610-621,
+    1973.
+    .. [2] Luis Pedro Coelho. "Mahotas: Open source software for scriptable
+    computer vision," Journal of Open Research Software, vol 1, 2013.
+    """
+    # sets the number of angles: 0 45 90 135. this must not be modified.
+    NumberofAngles = 4
```
cdeepakroy: Make NumberOfAngles as a parameter for the function

histomicstk/FeatureExtractionHaralick.py
```diff
@@ -0,0 +1,106 @@
+import skimage.feature
+import numpy as np
+
+
+def FeatureExtractionHaralick(I, Dst=1, Ng=256):
```
cdeepakroy: Maybe rename to ComputeHaralickFeatures

histomicstk/FeatureExtractionHaralick.py
slee172: Haralick feature uses 4 angles to compute means and ranges of the features. If we need to change the number of angles, it would not be haralick features I guess.

histomicstk/FeatureExtractionHaralick.py
slee172: Good suggestion. Thank you. !!

histomicstk/FeatureExtractionHaralick.py
cdeepakroy: @slee172 I was thinking people might want to do it just for horizontal and vertical adjacency. 

Also, the names of each of the features is not being returned which i think is important if we want to figure out which feature is important for a classification task or which feature correlates strongly with some genetic variable. Maybe return a pandas data frame.

Also, i see that these features are being computed for the whole image. We might also want to apply this for tissue/region classification and cell classification. So, I think it would be more generic if it takes an image and binary mask as input and compute the GLCM matrix for only those pairs of neighboring pixels that are within the mask (maybe with atleast one of the pixels in the mask). After that, you can compute all the features derived out of the GLCM matrix that you are currently computing.

CC: @cooperlab

slee172: @cdeepakroy Very good point !! I was also thinking about the binary mask to be used as a parameter because this haralick feature computation can be used for cell classification. My first thought was to add this function inside "for region in regionprops(Label)" in FeatureExtraction.py. For example,  
"FeaturesofHaralick[region.label-1, :] = \
            ComputeHaralickFeatures(In[region.bbox[0]:region.bbox[2], region.bbox[1]:region.bbox[3]]) " In this case, the ComputeHaralickFeatures uses the region of a nuclei as a parameter.
But, as you mentioned above, this function could be more independently used for other usages when adding a binary mask as a parameter. I think your approach would be better than my previous thought.

For the feature names, I'll update it soon.

For the angle selection, I have no idea of that yet. How about 1: horizontal and vertical 2: other two angles, 3: all of angles :)

cdeepakroy: @slee172 Maybe let us modeling the parameters similar to [matlab](http://www.mathworks.com/help/images/ref/graycomatrix.html).

See the GrayLimits, NumLevels, and Offset (this is probably the most general way to allow angle selection), and Symmetric parameters.

cdeepakroy:
@slee172 Maybe let us model the parameters similar to this [graycomatrix](http://www.mathworks.com/help/images/ref/graycomatrix.html#bu3j4f2-3) function in matlab.

See the following parameters:
* [Offsets](http://www.mathworks.com/help/images/ref/graycomatrix.html#input_argument_namevalue_graylimits): offsets along x-axis and y-axis to be used to determine neighboring pixels. This is a nice clean way to allow users to specify both angles and distances.
* [GrayLimits](http://www.mathworks.com/help/images/ref/graycomatrix.html#input_argument_namevalue_graylimits): Allows users to restrict the intensity range within which to compute the GLCM matrix. 
* [NumLevels](http://www.mathworks.com/help/images/ref/graycomatrix.html#input_argument_namevalue_numlevels): Allows users to bin the intensities. This reduces the susceptibility to noisy fluctuations in intensities of neighboring pixels.

slee172:
@cdeepakroy I think we need to create the graycomatrix function in a new python file called from the haralick feature extraction file with the parameters. Let's discuss it tomorrow's meeting.

slee172:
@cdeepakroy @jbeezley @cooperlab I got the error as belows when putting ComputeHaralickFeatures on travis :

10: /home/travis/build/DigitalSlideArchive/HistomicsTK/build/girder/plugins/HistomicsTK/histomicstk/ComputeHaralickFeatures.py:7: [C901] 'ComputeHaralickFeatures' is too complex (21)
10: def ComputeHaralickFeatures(I, Label, Offset, NumLevels=256,

Have you ever experienced on this?

cooperlab:
The continuous integration performs program analysis and will generate an exception if the cyclomatic complexity is too high.

You can avoid this by adding '# noqa: C901' as a comment two spaces after the function def. Try to reduce the complexity as much as possible before doing this.

slee172:
Thank you Lee. @cdeepakroy @jbeezley @cooperlab I found the error was caused when using more than two elif statements in the code. So, reduced them to less than 2 and finally passed the travis.

histomicstk/ComputeHaralickFeatures.py
```diff
@@ -0,0 +1,199 @@
...
+        90          [-D 0]
+        135         [-D -D]
+    NumLevels : integer
+        Specifies the number of gray level. NumLevels determines the size of
+        the gray-level co-occurrence matrix.
+        Default value = 256.
+    MaxGray : integer
+        Specifies a maximum grayscale value in I. Defalut value = 255
+    MinGray : integer
+        Specifies a minimum grayscale value in I. Defalut value = 0
```
cdeepakroy: @slee172 I would suggest clubbing MaxGray and MinGray into a single argument like GrayLimits or IntensityRange.

histomicstk/ComputeHaralickFeatures.py
```diff
@@ -0,0 +1,199 @@
+import numpy as np
+import pandas as pd
+import skimage.feature
+from skimage.measure import regionprops
+
+
+def ComputeHaralickFeatures(I, Label, Offsets=[[0, 1]], NumLevels=256,
+                            MaxGray=255, MinGray=0):
```
cooperlab: Binning the gray levels provides some robustness against noise or bias in intensity from one image to another. Maybe go with ~10 gray levels by default?

histomicstk/ComputeHaralickFeatures.py
slee172: Thanks !!

histomicstk/ComputeHaralickFeatures.py
slee172: Yes. binning strategies can be very useful for many tasks, such as feature extraction and segmentation. I think we need to talk a little about what is the optimal number of gray levels for the project tomorrow.

PR #103, opened 2016-06-06
Nuclei segmentation example

jbeezley:
This is kind of a WIP, but with the coming deadline I thought I would push it out now that there should be an end-to-end version working.

A couple of notes:
1. I made some changes to the docker image and pushed to jbeezley/histomicstk.  That should probably change back to dsarchive, but I don't think I have the credentials
2. The file names given in the UI must have the correct extensions: (png/jpg for the segmentation mask, anot for the annotations)

TODO:
1. Fix tests
2. Hook into stream events to update the jobs panel and annotation panels without reloading
3. When clicking on a job, open a new tab to the job view in girder.
4. Change the default style of the rectangle annotations to something other than black boxes


server/NucleiSegmentation/NucleiSegmentation.py
```diff
@@ -17,15 +18,33 @@
...
     #
     print('>> Reading input image')
 
-    imInput = skimage.io.imread(args.inputImageFile)[:, :, :3]
-
+    imInput = None
+    imageFile = args.inputImageFile
+    if os.path.isdir(imageFile):
+        dir = imageFile
+        for file in os.listdir(dir):
```
cdeepakroy: @jbeezley Why allow it to be a directory, loop through all files, try loading each one, and break at the first one that succeeds.

The front-end you developed already includes a nice widget to select an image file from any folder on girder. If the user selects a wrong format, we should maybe let the job fail. If you want to restrict input image files to simple png/jpg, we can add the fileExtenstions attribute like [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/server/NucleiSegmentation/NucleiSegmentation.xml#L28) for the inputImageFile parameter. Then your file selection GUI can restrict it to whichever file formats are mentioned in that list.

Dockerfile
```diff
@@ -43,6 +43,9 @@ RUN conda config --add channels https://conda.binstar.org/cdeepakroy && \
     conda clean -i -l -t -y && \
     rm -rf /root/.cache/pip/*
 
+# pregenerate font cache
+RUN python -c "from matplotlib import pylab"
+
```
cdeepakroy: Great! Did this get rid of those annoying fc-list warnings for each CLI?

server/NucleiSegmentation/NucleiSegmentation.py
jbeezley: Ideally, yes, but the hierarchy widget will only allow you to select items, not files.  When an item contains multiple files, you get a directory here.  This just my quick and dirty solution to avoid having to do a lot of work on the client side code.

Dockerfile
jbeezley: I think so.

server/NucleiSegmentation/NucleiSegmentation.py
cdeepakroy: Currently, after selecting a file for the inputImageFile in your earlier front-end and submitting the job, girder_worker takes care of copying the file from the girder database to the worker machine, and set args.inputImageFile to a local file path on the worker machine.

If we set it to a directory, may be girder_worker will try to copy the whole directory/folder from Girder which might take more time if it contains many images.

cdeepakroy: Ahh! I understand now.

cdeepakroy: Maybe create an issue, so we remember to address it after the demo.

server/__init__.py
```diff
@@ -21,5 +23,7 @@ def load(info):
 
     genRESTEndPointsForSlicerCLIsInDocker(
         info, 'HistomicsTK',
-        'dsarchive/histomicstk:v0.1.2'
+        'jbeezley/histomicstk'
```
cdeepakroy: I see that you already have owner rights for the dsarchive organization on Dockerhub.

Currently, Dockerhub automatically: 
* builds the dsarchive/histomicstk:latest image whenever a commit is merged into master
* builds the dsarchive/histomicstk:<release-tag> whenever a release is made for the HistomicsTK repo.

Once you merge these changes and are sure the end-to-end pipeline for NucleiSegmentation CLI (Select input image, parameters, set output file location and names through front-end, Submit job, Auto ingest the generated annotation file into girder, render the results) then create a new release and update the release tag here which will trigger the autobuild on DockerHub.

cdeepakroy:
I switched to your `annotation-rendering` branch, did npm install for girder, started girder+girder_worker_mongod, and executed the NucleiSegmentation CLI through the front-end. The job succeeded.

Then i used the file selector widget on the top-right of your front-end to loaded the image that was used as input to the NucleiSegmentation CLI. I was expecting that the annotation panel will show up and annotations auto-ingested from the annotation file produced by the CLI. But i do not see any annotation panel. Do i have to do anything else to see the annotations?

I also updated my large_image clone.

jbeezley:
You probably need to reload the page.  I need to add an explicit fetch on the item once the job finishes.

cdeepakroy:
@brianhelba @jbeezley Changed the name of NucleiSegmentation CLI to NucleiDetection

server/NucleiSegmentation/NucleiSegmentation.py
cdeepakroy: @zach @jbeezley What is the best/cleanest way to fix this?

In the REST end-point of a CLI, should i request the girder ID of an item or a file for an input argument of type image?

cdeepakroy: In the case of dicom images, i am guessing there will be an item on girder with all files of the dicom series. In the case ITK, we just give the path of the .mhd file to itkImageFileReader.

In the case of other formats such as nrrd or png all the image information is within one file.

cdeepakroy: CC: @jcfr

jbeezley: Items that are tagged by the large_image plugin contain a `largeImage.fileId` property that we should use.  The reason I couldn't use it for the purposes of this demo was because the analysis image couldn't handle the tiled image file produced by the large image plugin.

cdeepakroy: Added an issue on [girder_worker](https://github.com/girder/girder_worker/issues/46) to resolve this in a clean way.

jbeezley:
This is cleaned up for the most part.  I repaired the tests and improved the annotation rendering performance.  It should be able to handle several thousand annotations without too much trouble now.

Remind me if I forgot about anything we have discussed.

PR #105, opened 2016-06-15
Implementation of Membrane-detecting filter

slee172:
MembraneFilter.py is an implementation of a membrane-detecting filter using eigenvalue analysis based on hessian matrix.
- generates deviation, frobenius, eigenvalues, and thetas from Membraneness.py
- returns MxN intensity image of vessel filter response and MxN image of membrane angles in radians.

Updated **init**.py


cooperlab:
I think we need to change the naming of these filters - replacing 'vessel' with 'membrane'. These techniques will be used to analyze cell membranes and not vessels (although they were originally developed for angiography.) As long as the manuscript references are in the docstrings this should be fine.

cooperlab:
Please add a commit that also modifies "Vesselness.py"

slee172:
@cooperlab What is the best title of Vesselness.py you think?

cooperlab:
Membraneness. You will need to change the filename, function name and the docstring.

slee172:
Thanks !!

histomicstk/MembraneFilter.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # calculate blob score at multiple scales
+    for i in range(0, len(Sigmas)):
+        # blob score
+        Deviation, Frobenius, E, Thetas[:, :, i] = Membraneness(I, Sigmas[i])
+        # combine deviation and curvature into single score
+        R = np.exp(-0.5*(Deviation/Beta)**2) * \
+            (np.ones((M, N))-np.exp(-0.5*(Frobenius/C)**2))
+        # clear response where lambda_2 > 0
+        R[np.where(E[:, :, 1] < 0)] = 0
```
cdeepakroy: @slee172 @cooperlab  Maybe the entire computation of R should be wrapped inside the vesselness function.
 
Since the frangi expression/formula for R can be used for the enhancement of both blobs and vessels by tuning the parameters appropriately, maybe we should consider naming vesselness as `objectness` now or later.

Also, here is a nice generalized implement of this functionality in ITK which will be available once python-wrapped ITK is available for use in HistomicsTK:
* [itkHessionToObjectnessMeasureImageFilter](https://itk.org/Doxygen/html/classitk_1_1HessianToObjectnessMeasureImageFilter.html)
* [Blood vessel segmentation Example](https://itk.org/Doxygen/html/SphinxExamples_2src_2Nonunit_2Review_2SegmentBloodVesselsWithMultiScaleHessianBasedMeasure_2Code_8cxx-example.html#_a3)
* [Insight journal paper](http://www.insight-journal.org/browse/publication/175) with more details about the implementation. Here is the [pdf](file:///home/cdeepakroy/Downloads/IJ_175_Antiga-IJ-ObjectnessMeasure.pdf) of the paper. See the 2D example at the end.

histomicstk/Membraneness.py
```diff
@@ -4,10 +4,10 @@
 from Hessian import Hessian
```
cdeepakroy: @slee172 @cooperlab There are already a bunch of functions for this in scikit-image.
* [Hessian matrix](http://scikit-image.org/docs/dev/api/skimage.feature.html#skimage.feature.hessian_matrix)
* [Eigen values of the Hessian matrix](http://scikit-image.org/docs/dev/api/skimage.feature.html#skimage.feature.hessian_matrix_eigvals)
* [determinant of the hessian matrix](http://scikit-image.org/docs/dev/api/skimage.feature.html#skimage.feature.hessian_matrix_det)

Also numpy already has a way to compute [eigen values and vectors](http://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.eig.html#numpy.linalg.eig) for [multiple matrices at once](http://docs.scipy.org/doc/numpy/reference/routines.linalg.html#linear-algebra-on-several-matrices-at-once).

cdeepakroy:
@slee172 @cooperlab added a few comments.

I have a general question do discuss in our meeting: Do these membranes look like smoothed ridges or step-edges? Since membranes are at the cell boundary, i am guessing they should look like smoothed step-edges. In that scenario, would this vessel enhancement approach work? Maybe i need to take a close look at these membranes in some images.

histomicstk/Membraneness.py
cdeepakroy: @slee172 Reuse those functions as much as possible. They are most likely optimized underneath with C or Fortran libraries. Will make our code run faster.

cdeepakroy: I would remove Hessian.py and Eigenvalues.py and use those functions instead.

cooperlab: Let's discuss tomorrow - at some point we wondered whether or not we would stick with skimage. At this point it is pretty embedded in the code.

cooperlab:
Deepak - I can share some example data. We can discuss tomorrow but the ridges are fairly abrupt.

histomicstk/MembraneFilter.py
cooperlab: Let's discuss in the meeting tomorrow. I generally agree but it depends on how we want to present the functionality to developers.

slee172:
@cdeepakroy For hessian matrix, I found that there's no big difference between Hessian.py and hessian_matrix when scaling a test image (about 800 x 1000 size) on the Hessian.py regarding the running time. So, the decision would be made depending on what we choose for the code (using the package or using the code itself, I think there'll be pros and cons on it.) 

For numpy.linalg.eig, I also found there's another package named [scipy.linalg.eig ](http://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.eig.html) which is faster than numpy's one (I tested it), but the thing I'm wondering is that they just use a square array as their parameter. The parameter I'm trying to use for Eigenvalues.py was the all hessian matrix for the test image. Of course, we can use them already optimized as the frangi. Maybe we would need to find the best way to do this.

slee172:
@cdeepakroy  I got an error message as below when removing Hessian.py. Could you look this over?
 
[https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/145519351](url)

slee172:
@cooperlab @cdeepakroy This PR moved to #246

PR #107, opened 2016-06-21
Create examples as jupyter/ipython notebooks

cdeepakroy:


cdeepakroy:
@cooperlab As a proof of concept, i created an example notebook for nuclei segmentation. [Here](http://histomicstk.readthedocs.io/en/add-notebook-examples/examples/nuclei-segmentation.html) is how it renders on ReadTheDocs.

PR #112, opened 2016-07-01
Update MaxClustering.py

cooperlab:
Added type checking of input 'Mask' to avoid background masking error.


histomicstk/MaxClustering.py
```diff
@@ -48,6 +48,10 @@ def MaxClustering(Response, Mask, r=10):
            Biomedical Engineering,vol.57,no.4,pp.847-52, 2010.
     """
 
+    # check type of input mask
+    if ~(Mask.dtype == np.dtype('bool')):
```
zachmullen: Is this outer `~` meant to be logical negation? We should probably just use `!=` instead.

cdeepakroy:
LGTM!

PR #113, opened 2016-07-01
Update EmbedBounds.py

cooperlab:
Added shape checking for display and boundary image


cdeepakroy:
You may want to rename `Bounds` to something like 'ObjectBoundaryMask'. Otherwise LGTM!

PR #114, opened 2016-07-04
Update CondenseLabel.py

cooperlab:
Fixed output type of condensed label image to uint32. Integer type is required for compatibility with scipy label image processing algorithms.


cdeepakroy:
LGTM!

PR #119, opened 2016-07-07
Minimum-model nuclear segmentation

cooperlab:
Functions for minimum model nuclear segmentation and label image cleanup.


cdeepakroy:
@cooperlab 

This page shows how to instruct flake8 to ignore a whole file or a single line. You can also instruct it to ignore a specific error on a specific line. 

The error code for cyclomatic complexity seems C901.  So in this [line](https://github.com/DigitalSlideArchive/HistomicsTK/pull/119/files#diff-1bf3c88d1d8b11d968a6bb521b31d1ddR166) try replacing

`def ISBF(Mask, XStart, YStart, MaxLength):`  

with

`def ISBF(Mask, XStart, YStart, MaxLength):    # noqa: C901`

cooperlab:
Great thanks!


Lee Cooper, Ph.D.
lee.cooper@emory.edu
tel: (404) 712-0110

cdeepakroy:
Yes i think you need two spaces.

cooperlab:
The pyflakes exceptions worked - all checks passed.

histomicstk/MinimumModel.py
```diff
@@ -0,0 +1,795 @@
...
+        Close = ((Minima - Maxima) < 2).nonzero()
+        Maxima = np.delete(Maxima, Close)
+        Minima = np.delete(Minima, Close)
+
+        # remove pairs that do not have sufficient intensity transitions
+        if(Delta is not None):
+            if np.issubdtype(I.dtype, np.integer):
+                Range = Delta * 255.0
+            elif np.issubdtype(I.dtype, np.float):
+                Range = Delta * 1.0
```
cdeepakroy: `Range = Delta * 1.0` is redundant here.

Maybe it would be nice to normalize the image to [0, 1] range at the beginning of the function before the loop. You can also add the preferred intensity range as an optional parameter. After that you can directly use delta in the range [0, 1].

histomicstk/MinimumModel.py
```diff
@@ -0,0 +1,795 @@
...
+    Scores : array_like
+        A 1D array of horizontal coordinates of contour seed pixels for
+        tracing.
+
+    Notes:
+    ------
+    Can produce a large number of thin "halo" objects surrouding the objects
+    with higher scores. These can be removed by filtering object width in the
+    resulting label image.
+
```
cdeepakroy: When a portion of an object with lower score is overwritten by an object with higher score, maybe it would be better to remove/erase the object with lower score completely? What would be the use case that prevents us from doing this?

cooperlab: You could set a % overlap criteria - but it should be fairly high. Imagine a two partially overlapping contours that correspond to different cells with only small overlap. In that case a large portion of non-overlapping object would be deleted.

histomicstk/AreaOpenLabel.py
```diff
@@ -0,0 +1,58 @@
...
+    See Also:
+    ---------
+    CondenseLabel, ShuffleLabel, SplitLabel, MaxwidthOpenLabel
+    """
+
+    # copy input image
+    Opened = Label.copy()
+
+    # condense label image
+    if np.unique(Opened).size-1 != Opened.max():
```
cdeepakroy: Should we move this check inside CondenseLabel?

histomicstk/AreaOpenLabel.py
```diff
@@ -0,0 +1,58 @@
...
+            # extract object from label image
+            Template = Opened[Locations[i-1]]
+
+            # label mask of object 'i'
+            Template[Template == i] = 0
+
+    # condense to fill gaps
+    Opened = htk.CondenseLabel(Opened)
+
+    return Opened
```
cdeepakroy: We can do this using [regionprops](http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.regionprops) but this is probably faster. Wish they adapt `regionprops` in scikit-image to behave similar to the matlab one wherein we can request a selected subset of the properties to be computed.

cdeepakroy:
LGTM! We should plan on writing tests for all the major functions to uncover any bugs.

histomicstk/AreaOpenLabel.py
cooperlab: Yes we should. Condensing the label image is also not really needed - it's more of a convenience for implementation. I plan to re-write so that we don't need to call condense because it completely breaks the correspondence of values between inputs and outputs.

histomicstk/AreaOpenLabel.py
cooperlab: Yea - the scikit-image regionprops is fairly fast though. I think we are getting close to to having the functionality needed to replicate these features trivially.

histomicstk/MinimumModel.py
cooperlab: I'm hesitant to scale the input images unless we assume a standard value range. This can lead to very different behavior on distinct tiles from the same image. It seems like requiring the users to specify the input ranges every time is more work than asking them to scale the inputs in the occasional case where they are disregarding the range.

PR #122, opened 2016-07-12
Pin the pillow version for conda.

manthey:
Pillow 3.3.0 doesn't work in the travis environment without more work, so pin to 3.2.0.

### DIFF
--- a/requirements_c_conda.txt
+++ b/requirements_c_conda.txt
@@ -20,3 +20,6 @@ scipy==0.16.0
 
 scikit-learn==0.17.1
 pandas==0.18.1
+
+# We want Pillow 3.2.0.  Make sure conda doesn't change it on us.
+Pillow==3.2.0


cdeepakroy:
@manthey Thanks for this PR. LGTM!

On a general note, should we always pin to a specific version using `package==version` rather than ask for the latest version using `package>=version`? 

If we pin, i feel we have to keep looking for newer versions on all dependent packages which is heavy on maintenance. Or is the strategy that, we pin to a newer version only when we need a feature added by the newer version.

cdeepakroy:
@manthey None of the functions in HistomicsTK use PIL yet. Currently, i think only large_image has a dependency on Pillow. Should we pin in large_image instead?

manthey:
@jbeezley has commented on versions and what should be done.  See https://github.com/girder/girder/pull/1138 for a discussion on this.

manthey:
large_image already pins to PIL 3.2.0.  The problem is that HistomicsTK's travis file uses conda, and conda pulls PIL in as a subdependency (I'm not sure what triggers it), and that pulls in the latest version of PIL unless it is pinned.  So, it has to be pinned for travis.  We could just have it in the travis.yml file, but having it in the conda requirements file means that others using conda will get the version we want.

manthey:
@cdeepakroy Do you think would be better done just in the travis.yml file?  We could get the same result by adding the Pillow version to the conda install line.  Or shall I merge this in with the change to the requirements file?

cdeepakroy:
@manthey Let us merge this. I am guessing people using conda might run into the same error.

PR #123, opened 2016-07-13
Fix scrolling in control panels

jbeezley:
This reverts the change in 3794a403, which was intended to fix behavior observed in the demo video... When scrolling on the side panel, the image zooms when reaching a gap in the panels.  

That change had the consequence of disabling scrolling on the side panel entirely.  When attempting to fix this, I found I was unable to reproduce the behavior seen in the video.  If anyone can reproduce it with this change, let me know.

### DIFF
--- a/web_client/stylesheets/panel/panel.styl
+++ b/web_client/stylesheets/panel/panel.styl
@@ -1,8 +1,6 @@
 .h-panel-group
-  pointer-events none
 
   .h-panel
-    pointer-events initial
     background-color #ffffff
     opacity .95
     margin-top 5px


cdeepakroy:
@jbeezley I just tried it and i was able to reproduce the behavior seen in the demo. If you want, we can do a quick hangout in the afternoon so you can take a look.

jbeezley:
What browser and OS are you using?

cdeepakroy:
I am using
* Google chrome Version 50.0.2661.86 (64-bit)
* Ubuntu 14.04

jbeezley:
I couldn't manage to get Chrome < 51 installed on my Ubuntu 14.04 VM.  The issue didn't occur on chrome 51.0.2704.79.  I think it's either a bug in that particular version of chrome or you have an old stylesheet or jade template cached somewhere.

cdeepakroy:
@zachmullen / @manthey Can you please test it on your machine? If you don't see this problem then something is wrong with my machine.

manthey:
@cdeepakroy I don't see the issue on my machine (Chrome 51, Windows 7).  I'll try another machine, too.

manthey:
Chrome 51, Ubuntu 14.04 works for me properly, too.  I was using a mouse wheel for both Windows and Ubuntu.  I also tried a touchpad and a touch screen on Windows, all without any issue.

cdeepakroy:
Please feel free to merge then.

Maybe something is wrong with my machine then. I will try to figure out why.

jbeezley:
I'm not certain what version of is up at girder.neuro.emory.edu, but as we mentioned in the meeting the bug does exist there.  Modifying the CSS on that site by removing the `pointer-events: none` as is done in this branch seems to fix the issue.  I'll go ahead and merge once the CI tests pass.

PR #130, opened 2016-07-17
Update EmbedBounds.py

cooperlab:
Restricted check of size consistency on inputs 'I' and 'Bounds' to first two dimensions (width + height).


cdeepakroy:
LGTM!

PR #131, opened 2016-07-19
[WIP]Add docker images via settings

salamb:
@brianhelba @cdeepakroy 
The key is histomicstk.docker_images and the respective value is a list of dictionaries.
Due to the way settings are stored in girder, adding a docker image with a period causes a database issue.


server/__init__.py
```diff
@@ -5,14 +5,16 @@
 
 from .rest_slicer_cli import genRESTEndPointsForSlicerCLIsInDocker
 from .handlers import process_annotations
-
+from .constants import *
```
manthey: It is bad form to `import *`.  It would be clearer to `from . import constants`, and then use `constants.(value)`.  This makes it easier to figure out where functions and values come from when reviewing code.  (In python, explicit is better than implicit.)

server/constants.py
```diff
@@ -0,0 +1,67 @@
...
+    except subprocess.CalledProcessError as err:
+        # the image does not exist locally, try to pull from dockerhub
+
+        return None
+
+    return data
+
+
+SettingDefault.defaults[PluginSettings.DOCKER_IMAGES] = \
+    [{'dsarchive/histomicstk:v0-1-3': None}]
```
cdeepakroy: @zachmullen We want to add the list of docker images (for which end-points will be create and exposed) to  HistomicsTK's plugin settings. The presence of dots in the docker image string (ex: `dsarchive/histomicstk:v0.1.3`) seems to be an [issue](https://github.com/girder/girder/issues/139) for mongodb. Is there another work around other than using `v0-1-3` instead of `v0.1.3`.

zachmullen: I would recommend restructuring the document so that the image ID becomes a value rather than a key.

```
{
    'image': 'dsarchive/histomicstk:v0.1.3',
    'some_other_field': None
}
```

Does that work?

cdeepakroy: That would work. But after this we want to cache (maybe in `some_other_field` above) the CLI list and xml-spec of each CLI in the docker image. I feel the retrieval of the cached information for a docker image will become O(n).

zachmullen: If you need to do a search in mongo by that image, you should just add an index on that image key, and it will make your lookups O(log n).

zachmullen: As a general rule, it's a good practice not to allow runtime user-specified data to be keys in mongo, but instead use a static key and use the runtime data as a value. This makes things like indexing possible and also avoids the problem of constraints on key names.

One notable exception is how girder handles metadata on folders and items, but we made a conscious decision there to violate this rule in the interest of facilitating arbitrary key/value search.

salamb: I think the best solution would be to have image ids(dockers hash on the image) be the keys and the value be a a dictionary with keys: name and xml. Even though it will be o(n) time to confirm matching image names, we can check for duplicate and updated images in 0(1) time .

zachmullen: If you anticipate having enough of these cache documents that this matters, I would suggest breaking it out of the settings and making these things documents in their own collection.

cdeepakroy: I think let us maybe use what zach first suggested for now

```python
[
    {
       'image': 'dsarchive/histomicstk:v0.1.3',
       'clis': { 
                  "ColorDeconvolution" : {"type": "python", "xml-spec": blah},
                  "NucleiDetection": {"type": "python", "xml-spec": blah} 
                }
     },
    {
       'image': 'cooperlab/nucleicuts:v0.1.0',
       'clis': { 
                  "ColorNormalization" : {"type": "python", "xml-spec": blah},
                  "NucleiSegmentation": {"type": "python", "xml-spec": blah} 
                }
     }
]
```

The `number of docker images` is not going to be that big for a distant future. We can think of optimizing when we face the bottleneck. Maybe later we can compute hash out of the image name and use it as a key.

salamb:
@cdeepakroy @brianhelba  @manthey 
example of  data stored in the the histomicstk setting 
{
  "0f9021276df2628de8c5af50e69cb091724abf46f268c3b0179635b662156d9f": {
    "Gaussian-itk": {
      "type": "python",
      "xml": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<executable>\n </executable>\n\n"
    },
    "docker_image_name": "salamb/histomicstk_modules:gaussian"
  }
}

In the future image id will also be stored. Need to decide on a policy for when a newer image or different version is available with the same name.

server/__init__.py
```diff
@@ -5,25 +5,36 @@
...
-    genRESTEndPointsForSlicerCLIsInDocker(
-        info, 'HistomicsTK',
-        'dsarchive/histomicstk:v0.1.3'
-    )
+    genRESTEndPointsForSlicerCLIsInDocker(info, resource, dockerCache)
 
-    events.bind('data.process', 'HistomicsTK', process_annotations)
+events.bind('data.process', 'HistomicsTK', process_annotations)
+events.bind('model.setting.validate', 'histomicstk_modules',
+            docker_resource.validateSettings)
```
cdeepakroy: The indentation of lines 38-40 seems wrong here? They should be within the `load(info)` function

server/__init__.py
```diff
@@ -5,25 +5,36 @@
...
+    info['apiRoot'].HistomicsTK = resource
+
+    dockerCache = DockerCache(docker_resource.getDockerImages())
+    dockerImages = dockerCache.getDockerImg()
 
-    genRESTEndPointsForSlicerCLIsInDocker(
-        info, 'HistomicsTK',
-        'dsarchive/histomicstk:v0.1.3'
-    )
+    genRESTEndPointsForSlicerCLIsInDocker(info, resource, dockerCache)
```
cdeepakroy: I would suggest creating another function named `genRESTEndPointsForSlicerCLIsInDockerCache` rather than modifying the existing one. That way we will have one function that works from images in the local or global docker registry. And another which caches them.

server/docker_resource.py
```diff
@@ -0,0 +1,343 @@
...
+    cause a job to generate the cli handler and generate the rest endpoint
+    asynchronously.Deleted images will result in the removal of the rest api
+    endpoint though docker will still cache the image unless removed manually
+    (docker rmi image_name)
+    """
+
+
+    def __init__(self):
+        super(DockerResource, self).__init__()
+        self.resourceName = 'HistomicsTK'
```
cdeepakroy: resourceName is hard-coded.

We are planning on moving the whole REST end-point generation functionality into girder or girder_worker so it can be used for other projects. So we have to make this docker caching infrastructure generic. Maybe you can have a DockerCache base class that can be specialized for each project. Or maybe there is a way better than that.

server/docker_resource.py
```diff
@@ -0,0 +1,343 @@
...
+        self.route('PUT', ('add_docker_image',), self.appendImage)
+        self.route('DELETE', ('docker_image',), self.deleteImage)
+
+    @access.admin
+    @describeRoute(
+        Description('Remove a docker image from histomicstk')
```
cdeepakroy: histomicstk hard-coded.

server/__init__.py
salamb: What do you mean by global registry? The way we are executing containers, any images we run must be available locally.

cdeepakroy: I meant the prior version of `genRESTEndPointsForSlicerCLIsInDocker` docker pulls from which registry docker is configured (default: Dockerhub) to use on that machine. docker registry. So maybe it is better to let that function stay.

And put your stuff in another function that works with the DockerCache into a separate function tentatively called `genRESTEndPointsForSlicerCLIsInDockerCache`

server/docker_resource.py
```diff
@@ -0,0 +1,343 @@
...
+    endpoint though docker will still cache the image unless removed manually
+    (docker rmi image_name)
+    """
+
+
+    def __init__(self):
+        super(DockerResource, self).__init__()
+        self.resourceName = 'HistomicsTK'
+        self.route('PUT', ('add_docker_image',), self.appendImage)
+        self.route('DELETE', ('docker_image',), self.deleteImage)
```
cdeepakroy: Rename to delete_docker_image?

manthey: Actually, the two end points should have the same name, since the method distinguishes them.  I.e.,
```
self.route('PUT', ('docker_image',), self.appendImage)
self.route('DELETE', ('docker_image',), self.deleteImage)
```

cdeepakroy: Also maybe add these end-points in a sub-route. Maybe `/docker_image_manager/`?

server/docker_resource.py
```diff
@@ -0,0 +1,343 @@
...
+        for name in names:
+
+            if not currentSettings.deleteImage(name):
+                raise RestException('%s does not exist' % name)
+        ModelImporter.model('setting').set(PluginSettings.DOCKER_IMAGES,
+                                           currentSettings.raw())
+
+    @access.admin
+    @describeRoute(
+        Description('Add a to the list of images to be loaded by histomics tk')
```
cdeepakroy: histomicstk hard-coded

server/docker_resource.py
```diff
@@ -0,0 +1,343 @@
...
+
+                    }
+                docker_image_name:name
+                }
+            }
+        """
+
+        self.data = cache
+
+    def getDockerImg(self):
```
cdeepakroy: Rename to `getDockerImages`.

server/docker_resource.py
```diff
@@ -0,0 +1,343 @@
...
+        else:
+            return False
+
+    def addImage(self, name):
+        imageKey = self._getHashKey(name)
+        self.data[imageKey] = {}
+
+        self.data[imageKey][self.imageName] = name
+
+    def addCLI(self, img_name, cli_name, type, xml):
```
cdeepakroy: This makes me think that it would be nice have another class called `DockerImage` which manages a list of CLIs with methods to addCLI, deleteCLI, getCLIXML etc. And DockerCache manages a list of `DockerImage` objects with methods to `addDockerImage, deleteDockerImage, updateDockerImage` etc.

cdeepakroy:
@zachmullen Can you please QA this? 

This PR by Bilal is meant to add a way to cache the Slicer CLI xml specs of each docker image into the plugin settings of HistomicsTK (or whichever plugin is using this functionality).

When we move this REST end point generation functionality into girder or girder_worker this docker caching functionality should also go with it. So i want it should be as generic as possible without histomicstk hard-coded.

server/__init__.py
```diff
@@ -5,25 +5,36 @@
 
 from .rest_slicer_cli import genRESTEndPointsForSlicerCLIsInDocker
 from .handlers import process_annotations
-
+from .constants import PluginSettings
+from .docker_resource import validateSettings, DockerResource,DockerCache
```
zachmullen: style: add space after comma

server/__init__.py
```diff
@@ -5,25 +5,36 @@
...
     'webroot.mako'
 )
 
 
 def load(info):
+    """
+    images = ModelImporter.model('setting').getDefault(
+            constants.PluginSettings.DOCKER_IMAGES)
+    ModelImporter.model('setting').set(PluginSettings.DOCKER_IMAGES, images)
+    """
```
zachmullen: What's the purpose of this comment block? Can we remove it?

salamb: It is removed in my current version

zachmullen:
It feels a bit wrong to store a set of multiple XML blobs inside a setting value. We have already built a separate REST API to manage these documents, which is a good idea given the complexity of the document. I think it makes sense to store these documents in a new database collection rather than dumping them all into a single nested document in the setting collection.

salamb:
So create a model called DockerImage save instances to the database?

zachmullen:
That's what I would recommend, yes. It would allow us to easily add indexes specific to this set of documents, which is the closest thing we have to a schema at this point.

salamb:
@cdeepakroy  @manthey @zachmullen @brianhelba 
PTAL

In the next PR cached data will be saved as part of a model and new rest endpoints will be generated immediately as opposed to restarting girder for changes to take affect

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+
+    for dimg in dockerImages:
+        # check if the docker image exists
+
+        id = localDockerImageExists(dimg)
+        if id is None:
+            id = pullDockerImage(dimg)
+        if id is None:
+            raise ValueError("The image %s does not exist locally or "
+                             "on the default registry" % dimg)
```
cdeepakroy: Lines 873-878 that check if the docker image exists locally and tried to pull if not available can be wrapped into a function say `getDockerImage(image_string, pull=True)`

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+        # get CLI list
+        cliListSpec = localDockerImageCLIList(dimg)
+        if cliListSpec is None:
+            raise ValueError("Could not get the cli list for the img %s most "
+                             "likely the docker image entrypoint does "
+                             "not accept the argument --list_cli" % dimg)
+        # pprint.pprint(cliListSpec)
```
cdeepakroy: You can move lines 881-884 that `check for None and raise an exception` inside the function `localDockerImageCLIList` to reduce cyclomatic complexity.

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+        # check if the docker image exists
+
+        id = localDockerImageExists(dimg)
+        if id is None:
+            id = pullDockerImage(dimg)
+        if id is None:
+            raise ValueError("The image %s does not exist locally or "
+                             "on the default registry" % dimg)
+        # get CLI list
+        cliListSpec = localDockerImageCLIList(dimg)
```
cdeepakroy: Maybe `getDockerImageCLIList` would be a better name.

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+            raise ValueError("Could not get the cli list for the img %s most "
+                             "likely the docker image entrypoint does "
+                             "not accept the argument --list_cli" % dimg)
+        # pprint.pprint(cliListSpec)
+
+        cliListSpec = json.loads(cliListSpec)
 
+        # Add REST end-point for each CLI
+        for cliRelPath in cliListSpec.keys():
+            cliXML = localDockerImageclixml(dimg, cliRelPath)
```
cdeepakroy: Maybe `getDockerImageCLIXML` is a better name.

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+            if cliXML is None:
+                raise ValueError("Could not get the cli for the img %s cli %s"
+                                 " most likely the docker image entrypoint does"
+                                 " not accept the argument %s --xml" %
+                                 (dimg, cliRelPath, cliRelPath))
```
cdeepakroy: Again you can move lines 892-896 (that check if `cliXML is None` and raise an exception) into the function `getDockerImageCLIXML`. That function should report an error if one occurs inside it. Also it would reduce cyclomatic complexity without sacrificing any readability.

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+                    cliRelPath, e)
+                print "cli handler"
+                continue
+
+            cliSuffix = os.path.normpath(cliRelPath).replace(os.sep, '_')
+
+            cliRunHandlerName = 'run_' + cliSuffix
+            setattr(restResource, cliRunHandlerName, cliRunHandler)
+            restResource.route('POST',
+                               (cliRelPath, 'run'),
```
cdeepakroy: Just to keep in mind (no need to change now), to get around name conflicts across docker images, maybe we can prefix this route with a massaged version (maybe use hash if there is no generic way to massage) of the docker image string

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+                fname = os.path.split(exc_tb.tb_frame.f_code.co_filename)[1]
+                print(exc_type, fname, exc_tb.tb_lineno)
+                continue
+
+            cliGetXMLSpecHandlerName = 'get_xml_' + cliSuffix
+            setattr(restResource,
+                    cliGetXMLSpecHandlerName,
+                    cliGetXMLSpecHandler)
+            restResource.route('GET',
+                               (cliRelPath, 'xmlspec',),
```
cdeepakroy: Again just to keep in mind (no need to change now), to get around name conflicts across docker images, maybe we can prefix this route with a massaged version (maybe use hash if there is no generic way to massage) of the docker image string

server/rest_slicer_cli.py
```diff
@@ -826,14 +829,163 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
+    @access.user
+    @describeRoute(
+        Description('Get list of relative routes to all CLIs')
+    )
+    def getCLIListHandler(self, *args, **kwargs):
+        return cliList
+
+    getCLIListHandlerName = 'get_cli_list'
+    setattr(restResource, getCLIListHandlerName, getCLIListHandler)
+    restResource.route('GET', (), getattr(restResource, getCLIListHandlerName))
```
cdeepakroy: Again just to keep in mind (no need to change now), to get around name conflicts across docker images, maybe we can prefix this route (second argument) with a massaged version (maybe use hash if there is no generic way to massage) of the docker image string

server/rest_slicer_cli.py
```diff
@@ -867,24 +1019,25 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
     for dimg in dockerImages:
```
cdeepakroy: In lines 999-1006 you do 

```python
    if isinstance(dockerImages, list):
        for img in dockerImages:
            if not isinstance(img, str):
                raise Exception('dockerImages must either be a single '
                                'docker image string or a list of docker '
                                'image strings')
    else:
        dockerImages = [dockerImages]
```

I feel this should not be done here. The output of docker images obtained from 
```python
dockerImages = dockerCache.getDockerImageList()
```
should already have been checked for errors inside. So, move that sanity check somewhere inside dockerCache.getDockerImageList() or elsewhere within DockerCache where list of docker image strings is provided as input to DockerCache.

Also, let us assume that `dockerCache.getDockerImageList` returns a list even if there is a single docker image.

server/rest_slicer_cli.py
```diff
@@ -867,24 +1019,25 @@ def genRESTEndPointsForSlicerCLIsInDocker(info, restResource, dockerImages):
...
 
         # get CLI list
-        cliListSpec = subprocess.check_output(['docker', 'run',
-                                               dimg, '--list_cli'])
+        cliListSpec = dockerCache.getCLIDict(dimg)
```
cdeepakroy: rename to `getCLIList`? 

Implementing it as a Dict was a convenience to make it easy to find CLIs. I went back and forth between making it a list of dicts vs a dict.

cdeepakroy: Maybe rename to `getCLIListSpec`

server/rest_slicer_cli.py
```diff
@@ -936,3 +1090,79 @@ def getCLIListHandler(self, *args, **kwargs):
 
     # return restResource
     return restResource
+
+
+def localDockerImageExists(imageName, pullIfNotExist=False):
```
cdeepakroy: rename to `getDockerImage`

server/rest_slicer_cli.py
```diff
@@ -936,3 +1090,79 @@ def getCLIListHandler(self, *args, **kwargs):
...
+        data = subprocess.check_output(
+            ['docker', 'run', imageName, '--list_cli'])
+        return data
+    except subprocess.CalledProcessError:
+        # the image does not exist locally, try to pull from dockerhub
+
+        return None
+
+
+def localDockerImageclixml(img, cli):
```
cdeepakroy: rename to `getDockerImageCLIXML`

server/rest_slicer_cli.py
```diff
@@ -936,3 +1090,79 @@ def getCLIListHandler(self, *args, **kwargs):
...
+    except subprocess.CalledProcessError:
+        if pullIfNotExist:
+            # the image does not exist locally, try to pull from dockerhub
+            # none is returned if it fails
+            data = pullDockerImage(imageName)
+            return data
+        return None
+
+
+def localDockerImageCLIList(imageName):
```
cdeepakroy: rename to `getDockerImageCLIList`

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+                self.data = json.loads(cache)
+        else:
+            self.data = {}
+
+    def getDockerImageList(self):
+
+        return [str(imgDict[DockerCache.imageName])
+                for (imgHash, imgDict) in iteritems(self.data)]
+
+    def imageAlreadyLoaded(self, name):
```
cdeepakroy: rename to `isImageAlreadyLoaded`?

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+        self.data[imageKey][DockerCache.cli_dict] = {}
+
+    def addCLI(self, img_name, cli_name, type, xml):
+
+        cliData = {}
+        cliData[DockerCache.type] = type
+        cliData[DockerCache.xml] = xml
+
+        imageKey = self._getHashKey(img_name)
+        self.data[imageKey][DockerCache.cli_dict][cli_name] = cliData
```
cdeepakroy: Maybe change this function as follows:

```python
    def addCLI(self, img_name, cliDict, xml):

        cliDict = cliDict.copy()
        cliDict[DockerCache.xml] = xml

        imageKey = self._getHashKey(img_name)
        self.data[imageKey][DockerCache.cli_dict][cli_name] = cliData
```

That way less refactoring will be needed it we decide to add more fields in addition to `type` for each CLI in slicer_cli_list.json

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+
+    def raw(self):
```
cdeepakroy: give it a more meaningful name. say `getData`?

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+            return None
+
+        if cli in imageData[DockerCache.cli_dict]:
+
+            return imageData[DockerCache.cli_dict][cli][DockerCache.xml]
+        else:
+            print('no cli named %s' % imgName)
+            return None
+
+    def getCLIDict(self, imgName):
```
cdeepakroy: rename to `getCLIListSpec`?

server/docker_resource.py
cdeepakroy: or `hasImage`

cdeepakroy: We might also find a method called `hasCLI` useful. You can add it in the next PR in the DockerImageCache class that manages a list of CLIs

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+                                           imgData[DockerCache.imageName])
+            if DockerCache.cli_dict not in imgData:
+                raise DockerImageDataError('There is no dictionary of cli ')
+            for (cliName, cliData) in iteritems(imgData[DockerCache.cli_dict]):
+                if DockerCache.type not in cliData:
+                    raise DockerImageDataError('The cli %s does not have a key'
+                                               ' for the type of cli' % cliName)
+                if DockerCache.xml not in cliData:
+                    raise DockerImageDataError('The cli %s does not have a key '
+                                               'for the xml of cli' % cliName)
```
cdeepakroy: I think all of this validation can be improved by writing a JSON template (or xml template for xml) and validating against that. But this is good for now.

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+        self.requireParams(('name',), params)
+        name = params['name']
+        name = json.loads(name)
+
+        if isinstance(name, list):
+            for img in name:
+
+                if not isinstance(img, six.string_types):
+                    raise RestException('%s was not a valid string.' % img)
+            else:
```
cdeepakroy: else seems improperly indented.

cdeepakroy: Also maybe it is better to check the sanity of all docker image strings in the list with a for loop raising an error if anything is incorrect. And then delete them in another loop.

server/docker_resource.py
cdeepakroy: or just `data`

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+        for name in names:
+
+            if not currentSettings.deleteImage(name):
+                raise RestException('%s does not exist' % name)
+        # need to remove each cli attribute
+        # need to remove the clli list attribute
+        # need to remove the rest route
+        print "after delete", currentSettings.raw()
+        ModelImporter.model('setting').set(PluginSettings.DOCKER_IMAGES,
+                                           currentSettings.raw())
```
cdeepakroy: Should we also run a `docker rmi` command to reclaim space occupied by the docker image? Can be in a later PR if you want.

server/docker_resource.py
```diff
@@ -0,0 +1,433 @@
...
+                    'docker images to be loaded ', required=True)
+            .errorResponse('You are not a system administrator.', 403)
+            .errorResponse('Failed to set system setting.', 500)
+    )
+    # TODO check the local cache and cloud for different images of same name
+    # TODO how to handle newer images(take the latest or require a confirmation)
+    # TODO use image id to confirm equivalence need v2 manifest schema on cloud
+    # TODO how to handle duplicate clis
+    # TODO create the new REST endpoints
+    def appendImage(self, params):
```
cdeepakroy: Maybe rename to `addImage`

server/docker_resource.py
zachmullen: Does the server require docker images, or just the worker? Sorry, bit out of the loop on this.

server/image_worker.py
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env python
```
cdeepakroy: Maybe the contents of this file should be moved into docker_resource.py. That way all the caching stuff is in one file. Not sure if i would use the functions in this file for anything else.

cdeepakroy:
@salamb Amazing work! Just added a few comments.

server/docker_resource.py
salamb: Each docker image may have multiple cli's. The server exposes each cli as a rest endpoint. Originally the server would query for all cli's available and their xml specs upon loading the histomicstk plugin. This requires multiple executions of docker run and delayed the loading of the  histomicstk plugin. The actual cli was executed as an async job when the rest endpoint is being used.

server/docker_resource.py
salamb: @cdeepakroy 
I will refactor the name, but it is returning the json structure defined for each image which is why I though calling it a dict will be more appropriate.

server/docker_resource.py
salamb: @cdeepakroy 
I am considering setImages just because calling this REST endpoint removes previous images and attempts to set only the ones passed in as a param as the new available docker images

salamb:
@cdeepakroy 
PTAL
For now I kept the async job functions in a separate file. I will move them to docker_resource when I refactor dockerCache

server/docker_resource.py
cdeepakroy: @salamb shall we do it as mentioned above?

In the future, we might add more fields other than type to the dict of a CLI. Essentialy the whole dict of a CLI needs to be caches no matter what fields are in it.

cdeepakroy: Rest LGTM!

cdeepakroy:
@salamb Just one comment on DockerCache.addCLI. Once you resolve that rest LGTM!

server/docker_resource.py
salamb: @cdeepakroy 
Since Docker cache was going to be refactored, I was going to make that change in the next PR.

cdeepakroy: Ok. We can merge this PR then.

PR #133, opened 2016-07-23
Minimum model fix

cooperlab:


cooperlab:
@cdeepakroy check this out when you get the chance. There were errors with min/max detection preventing segmentation from completing.

histomicstk/MinimumModel.py
```diff
@@ -158,23 +158,57 @@ def SeedContours(I, Delta=0.3):
...
+            MaxPos = 0
+            TranPos = 0
+
+            # iterate through maxima, identifying relevant minima for each
+            while MaxPos < Maxima.size:
+                Index = np.nonzero(Minima > Maxima[MaxPos])[0]
+                if Index.size:  # minima found beyond current maxima
+
+                    # get position of next minimum in array 'Minima'
+                    MinPos = np.min(np.nonzero(Minima > Maxima[MaxPos])[0])
```
cdeepakroy: Looks like this can replaced with `MinPos = np.min(Index)`. 

To keep it more simple i think you can set `MinPos = Index[0]` since output indices of np.nonzero() are already in ascending order.

cdeepakroy:
LGTM!

cooperlab:
@cdeepakroy conflicts on this branch need to be merged on command line. I can't push because master is protected though.

PR #134, opened 2016-07-26
Fix input images types in UI generation

jbeezley:
The changes in #103 expect image id's passed into the workers to be files rather than items.  This PR makes the corresponding change in the UI code.  The result is still a little bit hacky because I can't return the svs file produced by the the large image plugin for the end to end demo.  Instead, I have to use the original png.  Once the workers can read svs, we should fix [this line](https://github.com/DigitalSlideArchive/HistomicsTK/compare/input-image-ui?expand=1#diff-fca41eca09d6e677748f05dabcc6839eR65).

I also made two other small changes:

1. Input images now default to the currently displayed image.
2. Fixed a couple of small bugs in the header rendering.


cdeepakroy:
@manthey Can you please QA this?

manthey:
LGTM.  I'll update with master.  Since @jbeezley is away for a while, after travis is happy, we should merge it in.

PR #135, opened 2016-07-27
Update CondenseLabel.py

cooperlab:
Removed dependency on skimage. Corrected error where input 'Label' is 1x1.


cdeepakroy:
LGTM!

PR #136, opened 2016-07-28
Updated FeatureExtraction.py

slee172:
FeatureExtraction.py is modified.

1 Arbitary sized arrays can be used as parameters.

2 Stand matrix is removed.

3 ValueError is fixed.

4 ZeorDivisionError is fixed.

5 Boundary value error is fixed.


histomicstk/features/FeatureExtraction.py
```diff
@@ -21,11 +21,6 @@ def FeatureExtraction(Label, In, Ic, W, K=128, Fs=6, Delta=8):
         A T x T intensity image for Nuclei.
```
cdeepakroy: Works only on square images? If not this should be changed.

histomicstk/features/FeatureExtraction.py
```diff
@@ -21,11 +21,6 @@ def FeatureExtraction(Label, In, Ic, W, K=128, Fs=6, Delta=8):
...
     Ic : array_like
         A T x T intensity image for Cytoplasms.
```
cdeepakroy: Works only on square images? If not this should be changed.

cdeepakroy:
@slee172 
* Few minor comments in the docstring. 
* Integrate changes from master into your branch.

Rest LGTM!

histomicstk/features/FeatureExtraction.py
slee172: It supports M by N images. My mistake. Modified to M by N.

PR #138, opened 2016-07-28
Automated provisioning

zachmullen:


zachmullen:
- [x] Install girder stack
- [x] Install rabbitmq as task broker
- [x] Install girder worker
- [x] Create admin user and worker robot user
- [x] Install and run service scripts for girder and the worker
- [x] Install HistomicsTK, DSA, and large_image plugins into girder
- [x] Enable all required plugins
- [x] Configure system settings for the worker and homepage text
- [x] Create filesystem assetstore
- [ ] Ensure existence of TCGA collection in girder
- [ ] Upload sample image file?

cdeepakroy:
@zachmullen If possible, it would be nice if you could put in the instructions that build ITK with python wrapping within the deployment scripts.

Right now ITK is available for use in CLIs that are run within docker via this [base image](https://github.com/DigitalSlideArchive/base_docker_image/blob/master/Dockerfile). But if we want to use ITK while running outside docker, we are required to build ITK on our machine and add it to python path.

cdeepakroy:
Going forward, we will start using ITK within the core functions of HistomicsTK whenever needed. 

If we want to let developers use this infrastructure to setup a common development environment, then i think it would be nice if it installs of python wrapped ITK as well. Maybe make it optional to be used only to setup development environment?

zachmullen:
Ok, I'll make it into its own role in ansible so we can easily choose whether or not to deploy it.

cooperlab:
I think that's a good solution to maintain the possibility of a fast & lightweight deployment.

manthey:
Since the Vagrant file is using trusty as the base, we'll need to build our own OpenJPEG, then build libtiff and OpenSlide to use the correct versions.  It is possibly that if we switched to xenial as the base we could avoid having to do that, but I haven't tested that (I've tested building our own OpenJPEG on xenial, and that works.

zachmullen:
Is OpenSlide something that should be in its own role also, or do our girder plugins' server side code depend on it?

manthey:
The server doesn't strictly depend on it.  We need it to read certain files (including the ones from the TCGA archive), but one could avoid those files.

cooperlab:
OpenSlide has a couple possible important uses - providing uniform access key image metadata like magnification (regardless of vendor) and enabling access to unadulterated pixel data.

ansible/group_vars/all
```diff
@@ -5,4 +5,5 @@ histomicstk_version: master
 large_image_version: master
 dsa_version: master
 girder_worker_version: master
+itk_version: master
```
cdeepakroy: Should we peg this to some latest stable release of ITK? I presume the master keeps changing rapidly.

zachmullen: Probably. We can do the same for others also, if people have suggested tags please let me know.

cdeepakroy: Is it possible to auto-get the latest release like [here](http://julienrenaux.fr/2013/10/04/how-to-automatically-checkout-the-latest-tag-of-a-git-repository/) so we don't have to track/watch those repositories?

Should we get the master for all repos within DigitalSlideArchive (since they are closely dependent) and peg to latest release tags for external repositories (ex: ITK) only?

zachmullen:
Everything is here now except an "upload a sample image" step, which I'm not sure exactly how to handle at this point.

Be warned, the ITK build step takes multiple hours on my laptop inside vagrant. We may be able to optimize it a bit by turning off more components, but I wasn't sure exactly which configure definitions to use, so for now it's just `-DBUILD_TESTING:BOOL=OFF -DITK_WRAP_PYTHON:BOOL=ON`.

manthey:
Do we care about ansible warnings?  For instance, I get `[WARNING]: Consider using get_url or uri module rather than running curl` as I run this.

zachmullen:
Good point, I'll switch to using get_url so we don't download the file if we already have it.

manthey:
Very good.  The ITK build time is tedious -- is there a way to ask not to build it during the vagrant up/provision command?  Or, do you need to log into the VM and run ansible manually to skip the ITK build?

I'd like to include a few sample images, so one could run at least one of the HistomicTK analyses without any further downloads or work.  @cdeepakroy, @cooperlab, and @dgutman are there some good sample images we should use (and have appropriate license terms for us to distribute them in this way)?  We could, at least, include a folder in the TCGA collection called 'demo' or 'test' and place the images we use for testing (Easy1.png and L1.png).

manthey:
This morning, I get an error: 
```
WARNING: The following packages cannot be authenticated!
  mongodb-org-shell mongodb-org-server mongodb-org-mongos mongodb-org-tools
  mongodb-org
```
though it looks like the GPG key for mongo was added properly.

cooperlab:
David M.,


This is a great idea - I can pick a couple that have some interesting cells. All TCGA images are public and unrestricted.


We can either have them run at install or have DSA download and insert pre-run results with the images.


Lee Cooper, Ph.D.
lee.cooper@emory.edu
tel: (404) 712-0110

zachmullen:
> Very good. The ITK build time is tedious -- is there a way to ask not to build it during the vagrant up/provision command?

Sure, we just exclude it from the list of roles in `vagrant.yml`.

> WARNING: The following packages cannot be authenticated!
  mongodb-org-shell mongodb-org-server mongodb-org-mongos mongodb-org-tools
  mongodb-org

I'm seeing that as well, must have been a remote change in the last day or so. I'll look into it.

zachmullen:
It seems the mongo failure is just a mistake by the package managers, hopefully they will have it resolved soon. I'll try again later today and see.

dgutman:
So I think it would be nice to have a script that pulls test data.. but ideally the data would live outside of the VM per se... like in a "local" directory on the machine that is actually hosting the vagrant file..

zachmullen:
By default, vagrant maps the top-level directory into the VM under the directory `/vagrant`, so we could have a step that does a local import from within that directory to expose files from the host in girder.

manthey:
LGTM

We should add the test data script and linkage in another PR.

cdeepakroy:
@zachmullen 

I just tried:

* [download and install vagrant](https://www.vagrantup.com/downloads.html)
* `pip install ansible`
* `git clone git@github.com:DigitalSlideArchive/HistomicsTK.git`
* git checkout ansible-provisioning
* vagrant up
* vagrant ssh

I was able to generate the vagrant box and ssh into it smoothly.

Where in the vagrant box are all the repos (HistomicsTK, girder, large_image etc) clones?

zachmullen:
Pretty much everything gets placed under the path specified by the `root_dir` variable, which right now is set as `/opt/histomicstk`.

I'll merge this and commit the sample image stuff on a subsequent PR.

PR #141, opened 2016-07-29
Fix import issues post reorganization into sub-packages

cdeepakroy:
* import stuff in the correct order in `__init__.py` files (ex: `ColorDeconvolution` calls `ComplementStainMatrix `so it must be imported before `ColorDeconvolution` in `color_deconvolution/__init__.py`)
* import sub-packages in `__init__.py` files to allow nested calls (ex: histomicstk.preprocessing.color_normalization.ReinhardNorm)
* use relative imports for importing stuff in the same package (ex: see [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/reorg-core-functions/histomicstk/preprocessing/color_deconvolution/ColorDeconvolution.py#L2)) and use absolute imports for everything else.


cdeepakroy:
@zachmullen Can you please take a quick look at the `__init__.py` files?

CC @cooperlab

zachmullen:
Looks OK to me. The shadowing would only be a concern to me if you care a lot about potentially breaking API compatibility in the future if you ever want to restructure modules. That could be alleviated by renaming the python files to use snake_case, which is more conventional anyway.

cooperlab:
Thanks @zachmullen - conversion to snake case is next on the list.

cdeepakroy:
Thanks Zach.

@cooperlab @slee172 Please go through the changes and comments i have placed in `__init__.py` files (for how to do imports and expose package contents for public use) and voice your questions here.

Once we all agree that things are ok, we can merge this.

cooperlab:
@cdeepakroy LGTM. Not tricky but will have to watch the order moving forward. Thank you.

cdeepakroy:
Also, forgot to mention that i had to move `SimpleMask.py` from `histomicstk.segmentation` into `histomicstk.utils` to resolve a cyclic dependency. 

`Sample` in `histomicstk.utils` was using `SimpleMask` and other functions in `histomicstk.segmentation` were using `histomicstk.utils`.

slee172:
@cdeepakroy LGTM !!

PR #142, opened 2016-07-30
Collisions in SplitLabel.py

cooperlab:
Fixed issue with sequential writing of objects to output causing collisions.

### DIFF
--- a/histomicstk/segmentation/label/SplitLabel.py
+++ b/histomicstk/segmentation/label/SplitLabel.py
@@ -64,7 +64,7 @@ def SplitLabel(Label, Connectivity=8):
 
         # relabel if necessary
         if(Count > 1):
-            Template[Template == 1] = i
+            Template[L == 1] = i
             for i in np.arange(2, Count+1):
                 Template[L == i] = Total + 1
                 Total += 1


cdeepakroy:
@cooperlab  LGTM!

But there is a function called [skimage.measure.label](http://scikit-image.org/docs/dev/api/skimage.measure.html#label) in scikit-image which is similar to bwlabel in matlab. Can't we use it instead of SplitLabel and CondenseLabel?

cooperlab:
I'm not sure if skimage.measure.label provides the same functionality. SplitLabel and CondenseLabel are processing existing label images to make corrections. There may not be a corresponding binary image to operate directly from in some cases.

Would also prefer to avoid skimage where possible. Most of what we need is in scipy and I think we can eventually make up the rest.

cdeepakroy:
For the label mask generation use case, it looks like [scipy.ndimage.measurements.label](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.ndimage.measurements.label.html) and  [skimage.measure.label](http://scikit-image.org/docs/dev/api/skimage.measure.html#label) are equivalent. So we can use either.

We can always generate a binary mask from a label mask and pass it to one of these two functions to regenerate the label mask. 
```python
L = scipy.ndimage.measurements.label(L > 0)
```
I think these functions will internally do whatever SplitLabel (make sure two different connected components do not get the same label) and CondenseLabel (shift labels to remove gaps) does. Also, i think they use an optimized [linear-time union-find algorithm](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_label.py#L57-L63).

cooperlab:
I don't think we can rely on connected component labeling to implement what SplitLabel and CondenseLabel do. The label image input may not have been produced by connected component labeling in the first place, and relabeling it based on connectivity may merge objects. Consider the case of a label image containing clumped nuclei that have been split by watershed or similar post-processing. If you threshold this (Label > 0) and label using connectivity you will inevitably merge objects.

cdeepakroy:
Ok understood. Good to merge then!

PR #143, opened 2016-07-30
Fixes seeding issues

cooperlab:
Fixed the generation of seeds in areas with flat response values. These regions contain contiguous regions of local maxima and now each is assigned a single seed.

Also fixes the placement of seeds in non-convex objects, ensuring that they are located within the object.


histomicstk/segmentation/nuclear/MaxClustering.py
```diff
@@ -98,12 +97,35 @@ def MaxClustering(Response, Mask, r=10):
...
+    Seeds = np.round(Seeds).astype(np.uint32)
+
+    # capture maxima for each connected region
+    Maxima = spm.maximum(Response, Label, np.arange(1, Label.max()+1))
+
+    # handle seeds lying outside non-convex objects
+    Fix = np.nonzero(Label[Seeds[:, 0].astype(np.uint32),
+                           Seeds[:, 1].astype(np.uint32)] !=
+                     np.arange(1, Label.max()+1))[0]
+    Locations = spm.find_objects(Label)
```
cooperlab: @cdeepakroy I couldn't find a way to get the bounding box for a single object in a label image. The list of objects 'Fix' is a small subset and so it is wasteful to locate the bounds of all objects.

cdeepakroy:
LGTM!

PR #152, opened 2016-08-02
Refactoring feature extraction

slee172:
- __init__.py is updated

- FeatureExtraction.py is modified by using these four files.

- Four files are added

ComputeMorphometryFeatures.py
: receives a region property from regioprops, and returns 13 morphometry features. Added three morphometry features more: MajorMinorAxisRatio, MajorAxisCoordsX, and MajorAxisCoordsY.

ComputeFSD.py
: receives a labeled mask, number of boundary sampling points, number of frequency bins, and returns fourier descriptors.

ComputeGradientFeatures.py
: receives an intensity image and coordinates of an object from the image, and returns 8 gradient features.

ComputeIntensityFeatures.py
: receives an intensity image and coordinates of an object from the image, and returns 9 intensity features.


histomicstk/features/ComputeFSDs.py
```diff
@@ -0,0 +1,179 @@
...
+            2, np.linspace(0, np.log2(K)-1, Fs+1, endpoint=True)
+        )
+    ).astype(np.uint8)
+
+    FSDGroup = FSDs(Bounds[:, 0], Bounds[:, 1], K, Interval)
+
+    return FSDGroup
+
+
+def InterpolateArcLength(X, Y, L):
```
cdeepakroy: prefix internal (ones we dont want to expose to public) function name with _ to declare it as private

histomicstk/features/ComputeFSDs.py
```diff
@@ -0,0 +1,179 @@
...
+    Parameters
+    ----------
+    Label : array_like
+        A M x N label mask.
+    K : Number of points for boundary resampling to calculate fourier
+        descriptors. Default value = 128.
+    Fs : Number of frequency bins for calculating FSDs. Default value = 6.
+
+    Returns
+    -------
```
cdeepakroy: Should return a pandas data frame.

histomicstk/features/ComputeFSDs.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def ComputeFSDs(Label, K=128, Fs=6):
+    """
+    Calculates `Fourier shape descriptors` from a label mask.
+
+    Parameters
+    ----------
+    Label : array_like
+        A M x N label mask.
```
cdeepakroy: Does this work only for binary masks (i.e. labeled mask with a singe object)?

slee172: Yes. Can we talk a little with this?

cooperlab: We should be cautious about repeating calculations. Finding contours from a mask can be expensive and should be done once. Perhaps we can generate contours prior to feature extraction and pass them as a parameter?

cdeepakroy: I think if we modify the function to accept a labeled mask with multiple object and compute features for all the objects then we will end up calling find_boundaries only once for the whole labeled mask

slee172: OK. I'll take a look into and test it.

cooperlab: @slee172 Sanghoon - we can make each function accept a label image, and then perform the necessary looping over objects in the main function body. You can create a helper function to actually perform the calculation on each single object. This will make the code clean & manageable.

slee172: Yes. That is what I'm going to do. Will enable the ComputeFSDs.py (for example) to receive a labeled image and to perform the loops for each object.

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -0,0 +1,93 @@
...
+        'Circularity',
+        'Eccentricity',
+        'EquivalentDiameter'
+        'Extent',
+        'MajorAxisLength',
+        'MinorAxisLength',
+        'MajorMinorAxisRatio',
+        'Perimeter',
+        'Solidity',
+    """
```
cdeepakroy: @cooperlab @slee172 Lee and sanghoon Can you please take a look? I refactored morphometry features.

I removed CentroidsX, CentroidsY as these are absolution locations and are not invariant to spatial transforms. [Local centroid](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_regionprops.py#L420-L422) might be informative but even that would not be properly normalized unless it is computed relative to some oriented bounding box.

I add [EquivalentDiameter](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_regionprops.py#L397-L398)

cooperlab: I was going to suggest that we remove centroids. They certainly will be used (for calculating spatial features) but are not features themselves. Some segmentation algorithms also already provide their own seed points.

histomicstk/features/ComputeFSDs.py
cooperlab: I assume we will catch all these private functions when we re-name to snake_case. No harm in doing some now though as you encounter them.

histomicstk/features/FeatureExtraction.py
```diff
@@ -73,374 +78,104 @@ def FeatureExtraction(Label, In, Ic, K=128, Fs=6, Delta=8):
...
 
-    for i in range(0, Fs):
-        df['FSD' + str(i+1)] = FSDGroup[:, i]
+    # initialize panda dataframe
+    df = pd.DataFrame()
 
-    for f in HematoxylinIntensityGroup._fields:
-        df['Hematoxylin' + f] = getattr(HematoxylinIntensityGroup, f)
+    fmorph = ComputeMorphometryFeatures(Label)
+    df = pd.concat([df, fmorph], axis=1)
```
cdeepakroy: We can join the dataframes returned by all the feature group functions using pandas.concat at once as shown here.

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -0,0 +1,126 @@
...
+        ellipse that has the same second-moments as the region. Eccentricity
+        of an ellipse is the ratio of the focal distance (distance between
+        focal points) over the major axis length. The value is in the
+        interval [0, 1). When it is 0, the ellipse becomes a circle.
+
+    EquivalentDiameter : float
+        The diameter of a circle with the same area as the object.
+
+    Extent : float
+        Ratio of area of the object to its axis-aligned bounding box.
```
cdeepakroy: @cooperlab I think this needs to be compute relative to the tightest enclosing bounding box for most invariance.

cooperlab: @cdeepakroy meaning a box that can tilt? I agree. Consider an elliptical cell at 45 degrees. Is that what regionprops is providing?

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -1,73 +1,93 @@
...
+        'MajorMinorAxisRatio',
+        'Perimeter',
+        'Solidity',
+    ]
+
+    rprops = regionprops(im_label)
+
+    numFeatures = len(feature_list)
+    numLabels = len(rprops)
+    fdata = pd.DataFrame(np.zeros(numLabels, numFeatures),
```
slee172: need to be np.zeros((numLabels, numFeatures)),

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -1,73 +1,93 @@
...
+                         columns=feature_list)
+
+    for i in range(numLabels):
+
+        # compute Area
+        fdata.at[i, 'Area'] = rprops[i].area
```
slee172: area to Area

histomicstk/features/ComputeMorphometryFeatures.py
cdeepakroy: [This](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_regionprops.py#L402-L404) is what skimage.measure.regionprops computes.

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -1,73 +1,93 @@
...
+
+        # compute Circularity
+        numerator = 4 * np.pi * rprops[i].Area
```
slee172: Area to area

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -0,0 +1,126 @@
...
+        # compute Extent
+        fdata.at[i, 'Extent'] = rprops[i].extent
+
+        # compute MajorAxisLength and MinorAxisLength
+        fdata.at[i, 'MajorAxisLength'] = rprops[i].major_axis_length
+        fdata.at[i, 'MinorAxisLength'] = rprops[i].minor_axis_length
+
+        # compute MajorMinor axis ratios
+        fdata.at[i, 'MajorMinorAxisRatio'] = \
+            rprops[i].major_axis_length / rprops[i].minor_axis_length
```
slee172: Maybe if statement can be added to avoid ZeroDivisionError: float division by zero at this line

histomicstk/features/ComputeFSDs.py
```diff
@@ -0,0 +1,253 @@
...
+        descriptors. Default value = 128.
+    Fs : int, optional
+        Number of frequency bins for calculating FSDs. Default value = 6.
+    Delta : int, optional
+        Used to dilate nuclei and define cytoplasm region. Default value = 8.
+
+    rprops : output of skimage.measure.regionprops, optional
+        rprops = skimage.measure.regionprops( im_label ). If rprops is not
+        passed then it will be computed inside which will increase the
+        computation time.
```
cdeepakroy: @slee172 @cooperlab Optionally passing in output of regionprops to speed up computational time.

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,156 @@
...
+        Intensity image
+
+    num_hist_bins: int
+        Number of bins used to computed the intensity histogram of an object.
+        Histogram is used to energy and entropy features.
+
+    rprops : output of skimage.measure.regionprops, optional
+        rprops = skimage.measure.regionprops( im_label ). If rprops is not
+        passed then it will be computed inside which will increase the
+        computation time.
```
cdeepakroy: @slee172 @cooperlab Optionally passing in output of regionprops to speed up computational time.

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,156 @@
...
+        A labeled mask image wherein intensity of a pixel is the ID of the
+        object it belongs to. Non-zero values are considered to be foreground
+        objects.
+
+    im_intensity : array_like
+        Intensity image
+
+    num_hist_bins: int, optional
+        Number of bins used to computed the intensity histogram of an object.
+        Histogram is used to energy and entropy features. Default is 10.
```
cdeepakroy: @cooperlab @slee172 Should we set the number of bins using something like [this](http://stats.stackexchange.com/a/862)

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,156 @@
...
+        Difference between mean and median intensities of object pixels.
+
+    StdIntensity : float
+        Standard deviation of the intensities of object pixels
+
+    IqrIntensity: float
+        Inter-quartile range of the intensities of object pixels
+
+    MadIntensity: float
+        Median absolute deviation of the intensities of object pixels
```
cdeepakroy: Added [Inter-quartile range](https://en.wikipedia.org/wiki/Interquartile_range) and [median absolute deviation](https://en.wikipedia.org/wiki/Median_absolute_deviation) as more robust measure of spread.

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,156 @@
...
+        Minimum intensity of object pixels.
+
+    MaxIntensity : float
+        Maximum intensity of object pixels.
+
+    MeanIntensity : float
+        Mean intensity of object pixels
+
+    MedianIntensity : float
+        Median intensity of object pixels
```
cdeepakroy: Added MedianIntensity

histomicstk/features/ComputeMorphometryFeatures.py
cdeepakroy: Should we let it generate NaN in such cases or should we set to zero or in this case to some high value?

histomicstk/features/ComputeGradientFeatures.py
```diff
@@ -0,0 +1,68 @@
...
+        Diff = np.sqrt(Gx**2 + Gy**2)
+
+    if not len(C):
+        C = canny(I)
+
+    pixOfInterest = Diff[Coords[:, 0], Coords[:, 1]]
+    # compute gradient features
+    MeanGradMag = np.mean(pixOfInterest)
+    StdGradMag = np.std(pixOfInterest)
+    EntropyGradMag = entropy(pixOfInterest)
```
cdeepakroy: @slee172 Entropy should be computed from a pdf (normalized intensity histogram in this case). See [here](http://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.entropy.html)

histomicstk/features/ComputeFSDs.py
cdeepakroy: We have to implement our own lighter version of regionprops that only computes a small subset of the properties like `pixel_intensities`, `pixel_coords`, `pixel_linear_indices`, `centroid` and `bounding_box`. These are the ones that are needed in most of our feature computation functions to get to the objects easily.

histomicstk/features/ComputeFSDs.py
```diff
@@ -0,0 +1,253 @@
+import collections
+import numpy as np
+import pandas as pd
+from skimage.measure import regionprops
+from skimage.segmentation import find_boundaries
+
+
+def ComputeFSDs(im_label, K=128, Fs=6, Delta=8, rprops=None):
```
cdeepakroy: Rename to ComputeFSDFeatures to be consistent with other functions

histomicstk/features/ComputeFSDs.py
```diff
@@ -0,0 +1,253 @@
...
+        # get bounds of dilated nucleus
+        min_row, max_row, min_col, max_col = \
+            _GetBounds(rprops[i].bbox, Delta, sizex, sizey)
+        # grab label mask
+        lmask = (
+            im_label[min_row:max_row, min_col:max_col] == rprops[i].label
+        ).astype(np.bool)
+        # find boundaries
+        Bounds = np.argwhere(
+            find_boundaries(lmask, mode="inner").astype(np.uint8) == 1
```
cdeepakroy: @slee172 I think you can speedup by calling call `find_boundaries` just once above the loop for the whole label mask. And then get the boundaries of each object inside the loop by an `and` operation of the patches (of size equal to the object bounding box) of label mask and the binary mask returned by of find_boundaries.

histomicstk/features/FeatureExtraction.py
```diff
@@ -73,374 +78,98 @@ def FeatureExtraction(Label, In, Ic, K=128, Fs=6, Delta=8):
...
 
-    for f in EosinTextureGroup._fields:
-        df['Cytoplasm' + f] = getattr(EosinTextureGroup, f)
+    fint_nuclei = ComputeIntensityFeatures(Label, In, rprops=regions)
+    fint_nuclei.columns = ['Nucleus' + col for col in fint_nuclei.columns]
+    df = pd.concat([df, fint_nuclei], axis=1)
 
-    for f in EosinGradientGroup._fields:
-        df['Cytoplasm' + f] = getattr(EosinGradientGroup, f)
+    fint_cytoplasm = ComputeIntensityFeatures(Label, Ic, rprops=regions)
```
cdeepakroy: I think i need to pass in the cytoplasm mask here.

cdeepakroy: This is tricky, the same pixel can be considered to be in the cytoplasm of two nuclei. So it would not be possible to come up with one labeled mask. I now see why you were passing in coords earlier.

cdeepakroy: I think we have to make another function called `ComputeNeighborhoodIntensityFeatures` that takes in a labeled mask and computes intensity features in a ring like neighborhood around each object

histomicstk/features/FeatureExtraction.py
```diff
@@ -73,374 +78,98 @@ def FeatureExtraction(Label, In, Ic, K=128, Fs=6, Delta=8):
...
-    # calculate hematoxlyin features, capture feature names
-    HematoxylinIntensityGroup = IntensityFeatureGroup(In, Nuclei)
-    HematoxylinTextureGroup = TextureFeatureGroup(In, Nuclei)
-    HematoxylinGradientGroup = GradientFeatureGroup(In, Nuclei)
-    # calculate eosin features
-    EosinIntensityGroup = IntensityFeatureGroup(Ic, Cytoplasms)
-    EosinTextureGroup = TextureFeatureGroup(Ic, Cytoplasms)
-    EosinGradientGroup = GradientFeatureGroup(Ic, Cytoplasms)
+            np.logical_xor(Mask, binary_dilation(Nucleus, Disk))
+        )
```
cdeepakroy: I think it is the repeated call to binary_dilate that is hogging time. Maybe we dilate the whole label mask once to get a binary mask of cytoplasm of all cells. And the loop over the number of objects and use a KDtree to find the cytoplasm pixels of each nucleus.

histomicstk/features/ComputeIntensityFeatures.py
cooperlab: That's an interesting technique. The number of bins probably needs to be standardized in multiple calls across different images in a dataset, so it should be explicitly set by the user and consistent across an analysis.

histomicstk/features/ComputeGradientFeatures.py
cooperlab: Yes - good catch.

histomicstk/features/ComputeFSDs.py
cooperlab: Agreed - @cdeepakroy what submodule should this reside in?

histomicstk/features/FeatureExtraction.py
cooperlab: @cdeepakroy I think 'neighborhood' is a much better description for general applications. We deal with brain where there is no clear boundary between cells, but that may not be the case in images of other tissues, especially if the membrane is stained.

cooperlab: We didn't really talk about this yet - but the basic idea is to 1. dilate each nucleus 2. remove pixels corresponding to any nucleus from this dilated mask and 3. calculate statistics from the eosin channel (from color deconvolution) in this region.

This also brings up the point of how we will handle other images besides hematoxylin and eosin. The feature extraction will eventually need to be aware of what stains are coming, and what the expected localization is (nuclear, membrane, non-nuclear or some combination).

histomicstk/features/ComputeGradientFeatures.py
slee172: Oh, I see. Will be modified

histomicstk/features/FeatureExtraction.py
slee172: Maybe we need to create another file to be used in finding the region of cytoplasm?

cooperlab: @cdeepakroy is working on this now. It probably should not be stored in features submodule though.

histomicstk/features/ComputeMorphometryFeatures.py
slee172: how about scailing from 0 to 1. than set it to 1 ?

cooperlab: NaN and inf values are fine. I usually filter these out of the results because they correspond to degenerate objects.

From a software perspective not sure what is best. Just so we don't go setting off hundreds of thousands of warning messages during a run.

histomicstk/features/ComputeIntensityFeatures.py
slee172: good to know that.

slee172:
_InterpolateArcLength in ComputeFSDFeatures.py is modified to increase the efficiency.

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,156 @@
...
+    feature_list = [
+        'MinIntensity',
+        'MaxIntensity',
+        'MeanIntensity',
+        'MedianIntensity',
+        'MeanMedianDifferenceIntensity',
+        'StdIntensity',
+        'IqrIntensity',
+        'MadIntensity',
+        'Skewness',
```
slee172: ComputeGradientFeature.py also computes the Skewness. Would it be renamed to SkewIntensity ?

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,156 @@
...
+        'Kurtosis',
```
slee172: Kurtosis is as well..

histomicstk/segmentation/label/ComputeNeighborhoodMask.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+    # generate labeled neighborhood mask
+    im_neigh_label = np.zeros_like(im_label)
+
+    im_neigh_label[neigh_rind, neigh_cind] = im_label[
+        closest_obj_rind[neigh_rind, neigh_cind],
+        closest_obj_cind[neigh_rind, neigh_cind]]
+
+    return im_neigh_label
```
cdeepakroy: @cooperlab @slee172 Can you please take a look at this function?

I am using this to generate a labeled cytoplasm mask.

histomicstk/features/FeatureExtraction.py
```diff
@@ -1,12 +1,15 @@
...
 import numpy as np
 import pandas as pd
-from scipy.stats import entropy, skew, kurtosis
 from skimage.feature import canny
 from skimage.measure import regionprops
-from skimage.segmentation import find_boundaries
-from skimage.morphology import disk, dilation
+from skimage.morphology import disk, binary_dilation
+
+from .ComputeFSDFeatures import ComputeFSDFeatures
```
cooperlab: FSDs are shape features and should be calculated under shape/morphometry

histomicstk/features/FeatureExtraction.py
```diff
@@ -1,12 +1,15 @@
...
+from .ComputeGradientFeatures import ComputeGradientFeatures
+from .ComputeIntensityFeatures import ComputeIntensityFeatures
+from .ComputeMorphometryFeatures import ComputeMorphometryFeatures
```
cooperlab: I think we can go with 'shape' instead of 'morphometry'. Although morphometry strictly means shape, in pathology it also is used to describe textural and intensity characteristics.

histomicstk/segmentation/label/ComputeNeighborhoodMask.py
cdeepakroy: Below is an example of what this function does:

image

histomicstk/features/FeatureExtraction.py
```diff
@@ -73,374 +81,99 @@ def FeatureExtraction(Label, In, Ic, K=128, Fs=6, Delta=8):
...
 
-    for f in EosinTextureGroup._fields:
-        df['Cytoplasm' + f] = getattr(EosinTextureGroup, f)
+    fint_nuclei = ComputeIntensityFeatures(Label, In, rprops=regions)
+    fint_nuclei.columns = ['Nucleus' + col for col in fint_nuclei.columns]
+    df = pd.concat([df, fint_nuclei], axis=1)
 
-    for f in EosinGradientGroup._fields:
-        df['Cytoplasm' + f] = getattr(EosinGradientGroup, f)
+    cyto_mask = htk_label.ComputeNeighborhoodMask(Label, neigh_width=Delta)
```
cdeepakroy: Here is how i am generating the cytoplasm mask.

histomicstk/segmentation/label/ComputeNeighborhoodMask.py
cooperlab: Looks good to me.

histomicstk/features/FeatureExtraction.py
cdeepakroy: What bothers me with shape is that it does not include size and we have size features inside.

histomicstk/features/ComputeIntensityFeatures.py
cdeepakroy: Thanks for the catch.

histomicstk/features/FeatureExtraction.py
cooperlab: True - area alone is not shape. Neither is perimeter. But together they provide shape cues. MajorAxisCoordsX/Y also just indicate orientation, not shape.

I don't think there are enough size features that exist to create a separate category, and shape seems to be the closest category (containing features that depend only on the binary mask of an object).

histomicstk/features/ComputeGradientFeatures.py
```diff
@@ -0,0 +1,68 @@
...
+
+    References
+    ----------
+    .. [1] Daniel Zwillinger and Stephen Kokoska. "CRC standard probability
+    and statistics tables and formulae," Crc Press, 1999.
+    """
+
+    if not len(Diff):
+        Gx, Gy = np.gradient(I)
+        Diff = np.sqrt(Gx**2 + Gy**2)
```
cdeepakroy: @slee172 You may want to use [scipy.ndimage.gaussian_gradient_magnitude](http://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_gradient_magnitude.html#scipy.ndimage.gaussian_gradient_magnitude). A bit of smoothing will give us better gradients.

histomicstk/features/ComputeMorphometryFeatures.py
```diff
@@ -19,18 +19,24 @@ def ComputeMorphometryFeatures(im_label):
...
+        object/label listed below.
+
+    Notes
+    -----
+
+    List of morphometry features computed by this function:
+    - 'Area',
+    - 'Circularity',
+    - 'Eccentricity',
+    - 'EquivalentDiameter'
```
slee172: , should be added

histomicstk/features/ComputeIntensityFeatures.py
slee172: @cdeepakroy Could we use two features for gradient as well? And, don't see the MAD in the code, how about just using this : numpy.median(abs(x - numpy.median(x)))

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -0,0 +1,157 @@
...
+        'MeanIntensity',
+        'MedianIntensity',
+        'MeanMedianDifferenceIntensity',
+        'StdIntensity',
+        'IqrIntensity',
+        'MadIntensity',
+        'SkewnessIntensity',
+        'KurtosisIntensity',
+        'HistogramEnergy',
+        'HistogramEntropy',
```
slee172: @cdeepakroy Name also used in ComputeGradientFeatures.py. How about using HistEnergyIntensity and HistEntropyIntensity? So, HistEnergyGradMag and HistEntropyGradMag in ComputeGradientFeatures.py

cooperlab: @cdeepakroy @slee172 - what do you think about adding a prefix or suffix to the features to indicate what group they come from?

histomicstk/features/ComputeGradientFeatures.py
slee172: it took more time than the previous one. We need to choose which one is more important (efficiency or effectiveness).

histomicstk/features/ComputeIntensityFeatures.py
slee172: It could be very good if there are any standard prefix or suffix terms for image analysis, especially features.

slee172:
Code in the ComputeGradientFeatuers.py and FeatureExtraction.py is refactored

histomicstk/features/FeatureExtraction.py
```diff
@@ -60,6 +64,7 @@ def FeatureExtraction(Label, In, Ic, K=128, Fs=6, Delta=8):
         - MeanGradMag, StdGradMag, SkewnessGradMag, KurtosisGradMag,
         - EntropyGradMag, EnergyGradMag,
         - SumCanny, MeanCanny
+
```
slee172: @cooperlab @cdeepakroy I think we need to update docstring in the refactored FeatureExtraction.py

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -28,9 +28,6 @@ def ComputeIntensityFeatures(im_label, im_intensity,
         passed then it will be computed inside which will increase the
         computation time.
 
-    Coords : array_like
-        A N x 2 coordinate list of a region.
-
```
slee172: These lines seem to be removed

histomicstk/features/ComputeIntensityFeatures.py
```diff
@@ -107,7 +107,7 @@ def ComputeIntensityFeatures(im_label, im_intensity,
 
         # get intensities of object pixels
         pixelIntensities = np.sort(
-            im_intensity[rprops[i].Coords[:, 0], rprops[i].Coords[:, 1]]
```
slee172: changed Coords to coords

histomicstk/features/ComputeIntensityFeatures.py
cdeepakroy: @cooperlab I like the idea of a prefix. 

When we start doing machine learning, we will look at the features of all objects in the form of a table. Then, sorting the columns in alphabetical order will nicely keep features of the same group together.

cooperlab: OK. Eventually we also need to add information about what input the features were calculated on (hematoxylin, eosin, etc.)

histomicstk/features/ComputeIntensityFeatures.py
cdeepakroy: Thanks!

histomicstk/features/ComputeIntensityFeatures.py
cdeepakroy: We can indicate which channel the features are computed by a prefix as we are doing it now. See [here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/c1795bf5a32769a54ecdc8a78b79b25bb8ea9b11/histomicstk/features/FeatureExtraction.py#L90-L91).

slee172: Don't we add the Hematoxylin and Eosin? So, Hematoxylin.Intensity.Min or Intensity.Min.Hematoxylin ... (Or Nucleus.Intensity.Min / Intensity.Min.Nucleus .. )?

cdeepakroy: I am using `Nucleus.Intensity.Min`  instead of 'Hematoxylin.Intensity.Min' as i dint want to spell out the nuclei stain (thinking there could be other stains that highlight nuclei where i could be wrong). Also by looking at the name `Nucleus.Intensity.Min` it becomes apparent to non-pathology data science folks (who may not know stain types well) that it is a feature associated with the nucleus.

Similarly, I am using `Cytoplasm.Intensity.Min`  instead of 'Eosin.Intensity.Min'.

slee172: I agree with that.

histomicstk/features/ComputeGradientFeatures.py
cdeepakroy: It probably took more time because it applies smoothing before computing the gradient. A little bit of smoothing `sigma = 1` or `sigma=2` would give us some resilience against any noise.

histomicstk/features/ComputeIntensityFeatures.py
slee172: Seems like FSD goes shape. Or Texture? Could you recommend me?

histomicstk/features/ExtractNuclearFeatures.py
```diff
@@ -0,0 +1,119 @@
...
+    fdata = pd.concat([
+        fmorph,
+        ffsd,
+        fint_nuclei,
+        fint_cytoplasm,
+        fgrad_nuclei,
+        fgrad_cytoplasm
+    ], axis=1)
+
+    return fdata
```
cdeepakroy: @cooperlab @slee172 Can you please take a look?

Renamed FeatureExtraction to ExtractNuclearFeatures as it is specific to nuclei and cleaned up docstring. 

I like it that this function looks so simple and clean now.

histomicstk/features/ComputeGradientFeatures.py
```diff
@@ -96,32 +103,32 @@ def ComputeGradientFeatures(im_label, im_intensity,
...
+        fdata.at[i, 'Gradient.Mag.HistEnergy'] = np.sum(prob**2)
 
         bw_canny = cannyG[rprops[i].coords[:, 0], rprops[i].coords[:, 1]]
-        fdata.at[i, 'SumCanny'] = np.sum(bw_canny)
+        canny_sum = np.sum(bw_canny)
+        fdata.at[i, 'Gradient.Canny.Sum'] = canny_sum
 
-        fdata.at[i, 'MeanCanny'] = fdata.at[i, 'SumCanny'] / \
-            len(pixelGradients)
+        fdata.at[i, 'Gradient.Canny.Mean'] = canny_sum / len(pixelGradients)
```
slee172: @cdeepakroy This Gradient.Canny.Mean goes to 0, need to be changed to float or ..

slee172:
@cdeepakroy seems to update functions fist, and then update __init__.py last.

histomicstk/features/ComputeGradientFeatures.py
cdeepakroy: Fixed!

histomicstk/features/ExtractNuclearFeatures.py
```diff
@@ -0,0 +1,173 @@
...
+
+from histomicstk.segmentation import label as htk_label
+
+
+def ExtractNuclearFeatures(im_label, im_nuclei, im_cytoplasm=None,
+                           fsd_k=128, fsd_freq_bins=6, cyto_width=8,
+                           morphometry_features_flag=True,
+                           fsd_features_flag=True,
+                           intensity_features_flag=True,
+                           gradient_features_flag=True
```
cdeepakroy: @cooperlab @slee172 Added flags to toggle on/off individual feature groups

cdeepakroy: Also cytoplasm features will be computed only if cytoplasm channel is passed.

server/ComputeNucleiFeatures/ComputeNucleiFeatures.xml
```diff
@@ -25,6 +25,46 @@
...
+      <default>true</default>
+    </boolean>
+    <boolean>
+      <name>cytoplasm_features</name>
+      <label>Cytoplasm Features</label>
+      <channel>input</channel>
+      <longflag>cytoplasm</longflag>
+      <description>Compute Intensity and Gradient features from the cytoplasm channel</description>
+      <default>true</default>
+    </boolean>
```
cdeepakroy: @cooperlab @slee172 Renamed FeatureExtraction CLI to ComputeNucleiFeatures and added boolean flags to toggle on/off the computation of each feature group

cdeepakroy:
@cooperlab @slee172 I think the refactoring of feature extraction is pretty much done. Can you take a look to see there are any outstanding issues?

slee172:
@cdeepakroy @cooperlab  need to test on the server. I'll do this this evening and let you know before tomorrow morning.

cooperlab:
Looks great. The only thing I see is that FSDs can be grouped with morphometry/shape. This gives us a solid version for a first histomics release.

cdeepakroy:
Do we want the ability to toggle FSD features independently from other morphometry features?

cooperlab:
I don't think we need to toggle them independently. Are you still planning to make a base class eventually?

cdeepakroy:
I see us providing the toggle feature only at the group level. 

The individual feature toggling will probably not be used. I feel it just adds onto the number of decisions to make.

In the case of FSD, i feel they are a bit more advanced shape features. Maybe in the future we will add more shape features (ex: [ray features](http://www.kev-smith.com/papers/SMITH_ICCV09.pdf)). Maybe we keep each such feature group in a separate function and provide the ability to toggle them. We can prefix their feature names with `shape.` so that when all the features we compute get to a tabular format, all shape features/columns will be seen together.

cooperlab:
OK - good enough for now. We will be revisiting this submodule to accept metadata from Girder in the future on stain type and localization.

PR #158, opened 2016-08-08
Add python coverage tests to those run by travis.

manthey:
Enabled python code coverage measurement.  We want to add more tests, and this will help determine where they are needed.  We can add a service like codecov to make it easier to see what has been covered, if developers would find that useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,9 +90,9 @@ install:
 script:
     - mkdir -p $build_path/girder_testing_build
     - cd $build_path/girder_testing_build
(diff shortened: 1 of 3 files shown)


cdeepakroy:
LGTM! Can you also add reporting using codecov? That will give us a nice meter when we start writing unit tests.

manthey:
@cdeepakroy I've added codecov.  I think we'll get a badge on the front page README when this is merged in, though I'm not sure how to check that without merging the PR.

cdeepakroy:
@manthey Thank you for doing this. Let's merge.

PR #160, opened 2016-08-10
Add Haralick features to ComputeNucleiFeatures 

slee172:
ComputeHaralickFeatures.py is added to ComputeNucleiFeatures.py

### DIFF
--- a/histomicstk/features/ComputeHaralickFeatures.py
+++ b/histomicstk/features/ComputeHaralickFeatures.py
@@ -3,7 +3,7 @@ import pandas as pd
 from skimage.measure import regionprops
 from .graycomatrixext import graycomatrixext
 from .graycomatrixext import _default_num_levels
-
+from .graycomatrixext import _default_offsets
 
 def ComputeHaralickFeatures(im_label, im_intensity, offsets=None,
                             num_levels=None, gray_limits=None, rprops=None):
@@ -32,21 +32,7 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None,
         its neighbor. Note that the first dimension corresponds to
         the rows.
 
-        Because this offset is often expressed as an angle, the
-        following table lists the offset values that specify common
-        angles for a 2D image, given the pixel distance D.
-
-        AngleXY  |  OFFSET
-        -------  |  ------
-        0        |  [0 D]
-        45       |  [-D D]
-        90       |  [-D 0]
-        135      |  [-D -D]
-
-        Default
-            [1] for 1D,
-            [[1, 0], [0, 1]] for 2D,
-            [[1, 0, 0], [0, 1, 0], [0, 0, 1] for 3D and so on
+        See `histomicstk.features.graycomatrixext` for more details.
 
     num_levels : unsigned int, optional
         An integer specifying the number of gray levels For example, if
@@ -257,9 +243,12 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None,
 
     # offsets
     if offsets is None:
+
         # set default offset value
-        offsets = np.identity(num_dims)
+        offsets = _default_offsets(im_intensity)
+
     else:
+
         # check sanity
         if offsets.shape[1] != num_dims:
             raise ValueError(
--- a/histomicstk/features/ComputeHaralickFeatures.py
+++ b/histomicstk/features/ComputeHaralickFeatures.py
@@ -8,7 +8,7 @@ from .graycomatrixext import _default_offsets
 
 def ComputeHaralickFeatures(im_label, im_intensity, offsets=None,
                             num_levels=None, gray_limits=None, rprops=None):
-    """
+    r"""
     Calculates 26 Haralick texture features for each object in the given label
     mask.
 
--- a/histomicstk/features/ComputeHaralickFeatures.py
+++ b/histomicstk/features/ComputeHaralickFeatures.py
@@ -1,7 +1,8 @@
 import numpy as np
 import pandas as pd
 from skimage.measure import regionprops
-from graycomatrixext import graycomatrixext
+from .graycomatrixext import graycomatrixext
+from .graycomatrixext import _default_num_levels
 
 
 def ComputeHaralickFeatures(im_label, im_intensity, offsets=None,
@@ -12,12 +13,12 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None,
     Parameters
     ----------
     im_label : array_like
-        A labeled mask image wherein intensity of a pixel is the ID of the
+        An ND labeled mask image wherein intensity of a pixel is the ID of the
         object it belongs to. Non-zero values are considered to be foreground
         objects.
 
     im_intensity : array_like
-        Intensity image
+        An ND single channel intensity image
 
     offsets : array_like, optional
         A (num_offsets, num_image_dims) array of offset vectors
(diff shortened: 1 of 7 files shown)


histomicstk/features/ComputeNucleiFeatures.py
```diff
@@ -167,6 +180,26 @@ def ComputeNucleiFeatures(im_label, im_nuclei, im_cytoplasm=None,
 
         feature_list.append(fgrad_cytoplasm)
 
+    # compute nuclei haralick features
+    if haralick_features_flag:
+
+        fharalick_nuclei = ComputeHaralickFeatures(im_label, im_nuclei,
```
cdeepakroy: @cooperlab [ComputeHaralickFeatures](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/histomicstk/features/ComputeHaralickFeatures.py) has the following parameters:

    offsets : array_like, optional
        A (num_offsets, num_image_dims) array of offset vectors
        specifying the distance between the pixel-of-interest and
        its neighbor. Note that the first dimension corresponds to
        the rows.
        Because this offset is often expressed as an angle, the
        following table lists the offset values that specify common
        angles for a 2D image, given the pixel distance D.
        AngleXY  |  OFFSET
        -------  |  ------
        0        |  [0 D]
        45       |  [-D D]
        90       |  [-D 0]
        135      |  [-D -D]
        Default
            [1] for 1D,
            [[1, 0], [0, 1]] for 2D,
            [[1, 0, 0], [0, 1, 0], [0, 0, 1] for 3D and so on
    num_levels : unsigned int, optional
        An integer specifying the number of gray levels For example, if
        `num_levels` is 8,  the intensity values of the input image are
        scaled so they are integers between 1 and 8.  The number of gray
        levels determines the size of the gray-level co-occurrence matrix.
        Default: 8 for numeric image, 2 for binary/logical image
    gray_limits : array_like, optional
        A two-element array specifying the desired input intensity range.
        Intensity values in the input image will be clipped into this range.
        Default: [0, 1] for binary/logical image, [0, 255] for numeric image

What do you think would be good parameters for nuclei classification? 

Not sure if gray_limits [0, 255] is good for nuclei. The intensity range of nuclear/cytoplasm pixels might be much more narrow than [0, 255].

num_levels is set by default to log_base_2( max_intensity=256 ). 

what offsets should we use? By default it uses `offsets=[ [0, 1], [1, 0] ]` for 2D. I feel we should at least add the 45 degree offset.

CC: @slee172

cooperlab: Deepak - dealing with the varied ranges of input images is a troubling problem. Either we stretch the ranges of inputs - which creates huge problems for inter-image comparability - or we adopt some assumptions and ask users to follow best practices in analysis & development:

1. int type intensity images range from 0-255
2. float type intensity images range from 0-1
3. color conversions, deconvolution etc. attempt to preserve these ranges
4. filter responses may deviate from these ranges - these are not strictly intensity images

Color normalization should be used always - this will map intensity values to a reasonable range where subsequent operations like segmentation and feature extraction can make some reasonable assumptions.

cdeepakroy: Ok i like the idea of following that convention about intensity ranges (I will change the graycomatrixext function accordingly).

If we keep [0, 255] as the default, should we increase `num_levels` to something like 32. If we set it to 8, i feel all nuclear pixels will be restricted to few levels making glcm matrix very sparse. 

@slee172 Maybe this is a small experiment we may want to run - compute glcm matrix for say 100 randomly selected nuclei with different values of num_levels (8, 16, 32, 64, 128), and look at the sparsity (1 - fraction of non-zero values) of the glcm. The less sparse the better. And that will tell us what would be a good default value.

cooperlab: 8 is too few - 16 or 32 is more appropriate.

slee172: @cdeepakroy graycomatrixext function seems to restrict the size of nuclei box or mask to num_levels, so it occurs an error when any size of box is less than the num_levels. Do we put a condition out of the function or inside the function? And, could you inform me how to check the sparsity of the glcm?

slee172:
Set num_levels to 32, and modified doc string following graycomatrixext function.

histomicstk/features/ComputeHaralickFeatures.py
```diff
@@ -157,51 +226,57 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None, num_levels=8,
...
+    if num_levels is None:
+        num_levels = _default_num_levels(im_intensity)
 
     # check for consistent shapes between 'I' and 'Label'
     if im_intensity.shape != im_label.shape:
         raise ValueError("Inputs 'I' and 'Label' must have same shape")
 
-    # determine if image is grayscale or RGB
-    if len(im_intensity.shape) != 2:  # color image
-        raise ValueError("Inputs 'I' should be a grayscale image")
```
cdeepakroy: This function will work on ND single channel images. User of this function is supposed to make sure that he passes in a single-channel image.

histomicstk/features/ComputeHaralickFeatures.py
```diff
@@ -157,51 +212,60 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None, num_levels=8,
...
+
+    else:
+
+        # check sanity
+        if offsets.shape[1] != num_dims:
+            raise ValueError(
+                'Dimension mismatch between input image and offsets'
+            )
+
+    num_offsets = offsets.shape[0]
```
cdeepakroy: Moved the code that sets offset outside the loop. Also using a common function to set default values for `offsets`, `num_levels`, and `gray_limits`. If we decide to use a different default value in the future, we will only have to change things at one place.

histomicstk/features/ComputeHaralickFeatures.py
```diff
@@ -339,57 +374,8 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None, num_levels=8,
...
-        fdata.at[i, 'Haralick.Entropy.Range'] = \
-            np.ptp(ldata.loc[:, 'Haralick.Entropy'])
-        fdata.at[i, 'Haralick.Variance.Range'] = \
-            np.ptp(ldata.loc[:, 'Haralick.Variance'])
-        fdata.at[i, 'Haralick.DifferenceEntropy.Range'] = \
-            np.ptp(ldata.loc[:, 'Haralick.DifferenceEntropy'])
-        fdata.at[i, 'Haralick.IMC1.Range'] = \
-            np.ptp(ldata.loc[:, 'Haralick.IMC1'])
-        fdata.at[i, 'Haralick.IMC2.Range'] = \
-            np.ptp(ldata.loc[:, 'Haralick.IMC2'])
```
cdeepakroy: Simplified this.

histomicstk/features/ComputeHaralickFeatures.py
```diff
@@ -62,89 +57,149 @@ def ComputeHaralickFeatures(im_label, im_intensity, offsets=None, num_levels=8,
...
 
-    Haralick.IMC1.Range : float
-        Range the first information measures of correlation of object
-        pixels.
+            IMC2 = [1 - \exp(-2(HXY2 - HXY))]^{1/2}, \qquad where \\
 
-    Haralick.IMC2.Range : float
-        Range the second information measures of correlation of object
-        pixels.
+            HXY2 = -\sum_{i,j=0}^{levels-1} p_x(i) p_y(j) \log(p_x(i) p_y(j))
```
cdeepakroy: Added clearer documentation for all features. You can put latex equations in the docstring like this.

cooperlab: Nice! Will definitely use this in the future.

histomicstk/features/ComputeHaralickFeatures.py
slee172: OK.

histomicstk/features/ComputeHaralickFeatures.py
slee172: Good.

histomicstk/features/ComputeHaralickFeatures.py
slee172: Good !

PR #164, opened 2016-08-12
Reorder the openslide task so it is before installing pip's libtiff.

manthey:
If the python libtiff package is installed before we build our own openjpeg, libtiff, and openslide, then we can't read some images.  For example, https://data.kitware.com/#item/57a3519e8d777f126827a678 , won't work.


manthey:
@zachmullen Should we explicitly add a test?  My method is to upload the file I referenced and click on the "Use this file for a large image" icon.  It should immediately show you the image in the viewer if everything is working.

zachmullen:
LGTM, we should maybe add deployment testing later as a separate topic.

Issue #166, opened 2016-08-12
Improving routing to include form and image information


manthey:
Include image bounds and rotation.  Include form values

jbeezley:
How should the URL history be handled?  Should all changes in the form and image bounds push a new browser history entry or should it just replace the current URL?

I think the implementation of this depends pretty heavily on what its primary purpose is.  For example, if we just want a permalink to the current state of the page, replacing the URL is sufficient.  If we just want to preserve the state while navigating (or between sessions), a localStorage approach would work.  If we want the user to be able to back out of recent changes, we need to push history, but what is the resolution of the save points?

manthey:
I know we want this for permalinks.  I don't think we discussed other uses yet.

jbeezley:
Okay, I can start by replacing the URL state without pushing to history, and we can iterate from there.

manthey:
I'm closing this.  We don't currently url encode form data, but we do encode image information.  If we want other state, we can add new issues.

PR #168, opened 2016-08-15
Include displayed image bounds in route parameters

jbeezley:
This implements part of #166.  I'll do the form values in a different PR.


manthey:
Can you round the coordinates and rotation angle to a few (2?) decimal places.  It keeps the route shorter and looks less messy.

Convert the rotation from radians to degrees for the route.  This will make it easier to manually set it to 90 degrees, for instance.

Also, I'd rather see the coordinates in the order (left, top, right, bottom, rotation), rather than (left, right, top, bottom, rotation).

jbeezley:
@manthey PTAL

web_client/js/views/visualization.js
```diff
@@ -58,27 +58,47 @@ histomicstk.views.Visualization = girder.View.extend({
...
+            return;
+        }
+        var bounds = query.split(',').map(function (v) { return +v; });
+        this._map.rotation(bounds[4] * Math.PI / 180);
+        this._map.bounds({
+            left: bounds[0],
+            top: bounds[1],
+            right: bounds[2],
+            bottom: bounds[3]
+        }, null);
```
manthey: By passing `null` as a second parameter here and in the `var bounds = this._map.bounds(undefined, null);` call below, we are in the map gcs coordinates, rather than the interface coordinate system.  The difference is that the interface coordinate system is 0,0 is left, top and positive values are right, bottom, while the map gcs is right-handed with 0,0 left, top, with positive values to the right and *negative* values to the left.  Just removing the `null` (not padding a second parameter or passing undefined), we can stay with the interface coordinate system and have the coordinates in image pixels.

jbeezley: That's what I originally tried, but the interface coordinates with `+axis=esu` behaves strangely (or I just haven't grokked it).  In this case, calling `map.bounds()` returns as the distance in pixels below the bottom of the viewport rather than below the top of the viewport.  I think we could correct for this by subtracting the image height.  Is this something that should be changed in geojs?

manthey: I'm not sure what is going on there, and it will require more time to delve into it that I have right now.  Can we just invert the sign of the vertical parameters in the route (and, obviously invert it again when loading from the route), leaving the `null` parameter for now.

jbeezley: Sure, I can do that.

manthey:
LGTMWTIH

PR #171, opened 2016-08-16
Import contents of sample_images folder into TCGA collection

zachmullen:
Whatever files exist in the directory will be imported into the
collection during the ansible deployment process. This is an
idempotent operation.


zachmullen:
This is now ready, PTAL.

manthey:
LGTM.

PR #175, opened 2016-08-17
Fixes #167. Add dev role for vagrant box

### LINKED ISSUE #167: Add ansible task to make full development environment
It should install all of the appropriate requirements-dev.txt modules for girder our used plugins, conda, and the requirements listed in requirements_c_conda.txt.

Once installed, we should be able to make a build directory, run cmake (with appropriate flags), make, and ctest and have all of the HistomicsTK unit tests run.  I'm not sure if we should include this in the ansible task.

zachmullen:
This installs conda requirements and all dev requirements for girder.


manthey:
Is there a simple way to run the units tests for Girder and our plugins?  It seems like I need to activate a virtualenv for conda, but doing so doesn't include all of the pip-installed modules (or I'm doing it wrong).  I also need to be sudo, I think.

For instance, this doesn't work because of the conda environment (and I'd rather avoid the sudo)
```
sudo su
mkdir /opt/histomicstk/build
cd /opt/histomicstk/build
source ../miniconda/bin/activate ../miniconda
cmake ../girder
make
ctest -VV
```

I'm probably just missing something obvious.

zachmullen:
Pip has a fallthrough option for the virtualenv, does conda also?

manthey:
As far as I can tell, conda doesn't have any fallthrough option.  The only way I can figure out how to build in the conda environment is to pip install all of the appropriate packages in that environment.  Once I've done that, running `cmake` and `ctest` in the conda environment works except for (a) girder packaging tests (it fails at creating the correct virtualenv inside the conda env), and (b) the large_image tiles test (girder worker is using a different broker than the tiles test was expecting)

zachmullen:
Hm, I'm not sure how to work around that.

manthey:
One way would be to flog `PYTHONPATH`: 
```
export PYTHONPATH='/opt/histomicstk/miniconda/lib/python2.7/site-packages:/opt/histomicstk/miniconda/lib/python2.7/site-packages/ctk_cli-1.3-py2.7.egg'
```
and then the conda-installed modules should be available to the system python.  Very ugly, but probably functional (I'll test it after I do a fresh `vagrant up`).

manthey:
Setting the `PYTHONPATH` works, though I had to pip install the HistomicsTK `requirements_dev.txt` and `requirements_c.txt` files.  The docker test fails, probably due to the girder_worker broker.

zachmullen:
Heh, good catch, I completely forgot to install rabbit. I'll make that into a separate role, I can pretty much copy and paste it right out of COVALIC's.

zachmullen:
Wait, actually there is one, I just named it differently. I'll investigate.

manthey:
The large_image plugin's test explicitly asks for the mongo broker, but the HistomicsTK docker test doesn't.  I haven't looked into it further than that.

zachmullen:
Oh, didn't realize that. The default value now is `amqp://guest@localhost`, which is what the worker is configured to use. Rabbitmq is light and very easy to install, could we change large_image to use the default broker?

manthey:
Yes.  We should just change the large_image test to install rabbit and use the defaults.

The HistomicsTK travis test copies a girder_worker config file which mentions a mongo broker.  That seems wrong now.

zachmullen:
I've added the installation of the two missing requirements files. Is there anything else we want to do on this topic? Should we try to handle the PYTHONPATH somehow as part of the provisioning?

manthey:
I feel for ease of developers, we either need to alter PYTHONPATH or have a small utility script that would cmake / ctest appropriately.  In my ideal world, I'd love to see builds not require root permission (short of chown vagrant:vagrant /opt/histomicstk, I'm not sure what would be a good way to do this.

zachmullen:
@manthey I created a testing script that works; the tests expecting the mongo celery boker still fail, and it still requires root user to run for the moment, but it's pretty simple to run: `sudo ~/run_tests.sh`.

I think we should address the celery broker configuration problems and the requirement of root on a separate topic.

manthey:
I'm currently getting a failure when it tries to install fontello fonts: 
```
==> default: ncat:ext_css\" (concat) task\u001b[24m", "File clients/web/static/built/girder.ext.min.css created.", "", "\u001b[4mRunning \"fontello:ext_font\" (fontello) task\u001b[24m", "Verify paths...\u001b[32mOK\u001b[39m", "Creating session...\u001b[32mOK\u001b[39m", "Fetching archive...\u001b[31mFatal error: Cannot read property 'statusCode' of undefined\u0007\u001b[39m"], "warnings": []}
```
This has nothing to do with your most recent change, but it stops the provisioning.

ansible/roles/dev/templates/run_tests.sh.j2
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cd {{ root_dir }}/build
+export PYTHONPATH='{{ root_dir }}/miniconda/lib/python2.7/site-packages:{{ root_dir }}/miniconda/lib/python2.7/site-packages/ctk_cli-1.3-py2.7.egg:$PYTHONPATH'
+cmake ../girder
+make -j2
+ctest -j3 -V
```
manthey: Since this is a convenience script, I'll recommend letting command line parameters through to the ctest command.  That is, the last line would be changed to:
```
if [ $# -eq 0 ]; then
  ctest -j3 -V
else
  ctest "$@"
fi
```
So that it is easy to run a single tests as `ctest -VV -R mytest_regex`.

zachmullen:
PTAL

manthey:
LGTM

PR #181, opened 2016-08-23
color_normalization: Add unit tests and do PEP 8 renaming

cdeepakroy:
Also added a utility function `histomicstk/preprocessing/color_conversion/compute_lab_mean_std` to compute the mean and stddev of an RGB image in LAB space. These are needed as input for reinhard normalization.


cdeepakroy:
@cooperlab Can you please take a look?

cooperlab:
LGTM - my only comment is that 'compute_lab_mean_std' is a bit long.

cdeepakroy:
What would you suggest? Below are other alternatives:
* `lab_mean_std`
* `compute_lab_stats`

cooperlab:
'lab_mean_std' - I don't have a strong preference though. Don't change it if it's a PITA.

PR #183, opened 2016-08-30
Update girder version to fix installation

zachmullen:
This was a breakage in an upstream npm package related to fetching our fontello icon set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ addons:
 before_install:
     - CACHE="$HOME/.cache" OPENSLIDE_VERSION="3.4.1" source .install-openslide.sh
 
-    - GIRDER_VERSION=d910e905c4196e3046a129d0edb84e5c7e301596
+    - GIRDER_VERSION=c68d9c63fae8f5ffc24fe802f2b7333ad973273f
     - GIRDER_WORKER_VERSION=f834d4d3701df7f6f3f64fcd7d3f22d15b3b1db2
     - LARGE_IMAGE_VERSION=60c32b11abb4d5f0d1696221a7e81ab21c611980
     - main_path=$PWD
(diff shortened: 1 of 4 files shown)


zachmullen:
The `server_HistomicsTK.docker` test is failing due to timeouts, any idea what's going wrong there?

manthey:
I'm getting the same failure locally.  I'll look into it.

manthey:
The failure is caused by how we check if a job is updated.  Before, whenever a job was updated, save was called on the job model, and girder emitted `model.job.save` and `model.job.save.after` events.  Now, no such events are emitted.  Instead of hooking to these, we can bind to the `jobs.job.update` event.  Strictly, this event is only emitted when update is called via a REST endpoint (and not via the `updateJob` method), so this is not friendly to clients that don't use rest.

I can change the event that is bound, but it might also be desirable to refactor the jobs plugin to either (a) emit the `jobs.job.update` event regardless of update code path, and / or (b) emit the `model.job.save` and `model.job.save.after` events on updates.

zachmullen:
Ah, I see. Would it be suitable to just move the emission of `jobs.job.update` into the model layer rather than the REST layer?

manthey:
Yes, along with the `defaultPrevented` code that is also in the REST layer.

zachmullen:
Roger, I'll do that now.

manthey:
From a utility perspective, it would be nice to have a `jobs.job.update.after` event, too, so that a client that just wants the current state of the job doesn't have to combine the update parameters with the current job status.

zachmullen:
The tricky part here is that, unlike with most of our model layer events, we don't necessarily have the complete document available for these events, since part of this optimization was that we don't fetch or push the full document. So long as downstream handlers for this event understand that constraint, we should be fine.

zachmullen:
(At the moment, it's only the `log` field that might not be accurate.)

manthey:
Primarily, if there is an event which is sure to have the current job status, I think that would satisfy most client's requirements.

zachmullen:
Your changes LGTM, merge if you are OK with this in general.

PR #186, opened 2016-09-06
Reorganize analysis menu

jbeezley:
This fixes the UI to work with the latest rest API.  Currently, the submenus are ordered `image name -> version -> cli name`, but I can change that pretty easily.

screen shot 2016-09-06 at 3 45 16 pm


cdeepakroy:
I confirm that this works both inside and outside (on ubuntu 14.04) the vagrant box

Issue #188, opened 2016-09-08
Upgrade Python in ansible or install modules to validate SSL


manthey:
I get the following error during vagrant provisioning:

```
==> default: fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to validate the SSL certificate for deb.nodesource.com:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine or you can install the `urllib3`, `pyopenssl`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible"}
```

This is the SNI problem with Python < 2.7.9.

manthey:
It looks like we could just `pip install urllib3[secure]`.  See: http://urllib3-docs.readthedocs.io/security.html#installing-urllib3-with-sni-support-and-certificates (thanks @danlamanna).

danlamanna:
Alternatively, if you want to sidestep Python altogether: https://github.com/girder/girder/pull/1561/commits/05fceeb224430695189bf765b051e40685eb0518

cdeepakroy:
I got the same error. Should we add urlib3[secure] to requirements.txt?

manthey:
@cdeepakroy It needs to happen *before* we do the 'Download nodejs install script' step.  I just tried that and it failed for some reason (possibly because I messed something up).  @danlamanna's method of avoiding using python for that step might be a better solution.

manthey:
Fixed in PR #191.

PR #191, opened 2016-09-08
Fix node ppa key by using it directly.

manthey:
I tried pip installing urllib3[secure], but that still had issues. This is the same approach taken by girder/girder@05fceeb.  It works in my local test.

Feedback or a different approach welcome.


manthey:
Addresses issue #188.

cdeepakroy:
I independently confirm that this works.

PR #198, opened 2016-09-12
TraceBounds: Add unit tests

slee172:
Just for TraceBounds unit test. Not added cython version yet.


plugin.cmake
```diff
@@ -53,6 +53,11 @@ add_python_test(glcm
     PLUGIN HistomicsTK
 )
 
+add_python_test(segmentation_label
+    SUBMODULE TraceBoundsTest
+    PLUGIN HistomicsTK
+)
+
```
cdeepakroy: Can you try this:

add_python_test(segmentation_label
    SUBMODULE TraceBoundsTest
    DBNAME core_segmentation_trace_bounds
    PLUGIN HistomicsTK
)

I am guessing the test name is getting too long. With `DBNAME` you can override and give it a shorter name.

slee172: It works !!

slee172:
@cdeepakroy @manthey Could you take a look at setup.py? Is this right way to add ext_modules?

histomicstk/segmentation/label/trace_boundary.py
```diff
@@ -0,0 +1,167 @@
+import numpy as np
+import ctypes
+
+from histomicstk.segmentation.label.isbf import isbf
```
cdeepakroy: Change this to `from .isbf import isbf`. That should resolve the TravisCI error. In general, if your are import from a file in the same folder use relative import like this. Otherwise, use absolute import.

histomicstk/segmentation/label/__init__.py
```diff
@@ -12,6 +12,7 @@
 from .LabelPerimeter import LabelPerimeter
 from .ShuffleLabel import ShuffleLabel
 from .TraceBounds import TraceBounds
+from .trace_boundary import trace_boundary
```
slee172: @cdeepakroy What do you think of how to compare the original ISBF algorithm in TraceBounds with the cython version of ISBF algorithm? I just added the same function on it to compare them.

plugin_tests/segmentation_label_test.py
```diff
@@ -0,0 +1,62 @@
...
+                         [0, 1, 1, 1, 1, 0],
+                         [0, 1, 1, 1, 1, 0],
+                         [0, 0, 1, 1, 0, 0],
+                         [0, 0, 0, 0, 0, 0]], dtype=np.bool)
+
+        # test isbf with 4 connectivity
+        res_X, res_Y = label.TraceBounds(mask, 4)
+
+        # cython_X, cython_Y = label.tracebounds_cython(mask, 4)
+        cython_X, cython_Y = label.TraceBounds(mask, 4)
```
slee172: Got an error when using isbf on histomicstk. segmentation.label directory, so just wrote the same line on it temporally.

histomicstk/segmentation/label/tracebounds_cython.py
```diff
@@ -0,0 +1,166 @@
+import ctypes
+import numpy as np
+from .isbf import isbf
```
slee172: @cdeepakroy I wrote this here, but got an import error.

histomicstk/segmentation/label/trace_boundary.py
slee172: OK. will use the relative import.

histomicstk/segmentation/label/isbfcpp.h
```diff
@@ -0,0 +1,42 @@
...
+    int** matrix90;
+    int** matrix180;
+    int** matrix270;
+
+public:
+    isbfcpp();
+    int length();
+    void nth_from_last(int n, int &x, int &y);
+    bool addList(int x, int y);
+    void roateMatrix(int rows, int cols, int **input, int **output);
```
slee172: Fixed.

histomicstk/segmentation/label/__init__.py
cdeepakroy: I thought you already compared both and showed us that trace_boundary isbf is much faster. 

We should not have two functions which do the same thing in the toolkit. 

If the pure python one is slower, we should just drop it and keep the cython one instead.

slee172: @cdeepakroy OK. I see, so don't we need this unit test ?

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,344 @@
...
+{
+    addList(startX, startY);
+
+    nrows = rows;
+    ncols = cols;
+
+    matrix00 = new int*[nrows];
+    for (int i = 0; i < nrows; i++) {
+      matrix00[i] = new int[ncols];
+    }
```
cdeepakroy: You can use `std::vector< std::vector<int> >` to create a matrix. See [here](http://www.cplusplus.com/reference/vector/vector/)

cdeepakroy: [Here](http://thispointer.com/creating-a-matrix-using-2d-vector-in-c-vector-of-vectors/) is another example.

histomicstk/segmentation/label/trace_boundary.py
slee172: @cdeepakroy got an import error when using the relative. Could you see [this error](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/161689528#L2722)

histomicstk/segmentation/label/isbfcpp.cpp
slee172: Wow. Good to know that. I have used the vector as a return array between cython and cpp codes. But, try to use it in the cpp code.

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    // copy mask to matrix00
+    for(int i=0; i< nrows; i++){
+        for(int j=0; j< ncols; j++){
+            matrix00[i][j] = mask[i*ncols+j];
+        }
+    }
```
cdeepakroy: You can initialize an std::vector from c array like [this](http://stackoverflow.com/questions/2434196/how-to-initialize-stdvector-from-c-style-array). Generalize this to 2D.

histomicstk/segmentation/label/isbfcpp.h
```diff
@@ -0,0 +1,42 @@
...
+#include <iostream>
+#include <string>
+#include <vector>
+
+struct node
+{
+    int val_x;
+    int val_y;
+    node *next;
+};
```
cdeepakroy: Use `std::list` instead of this. See [here](http://www.cplusplus.com/reference/list/list/push_back/)

histomicstk/segmentation/label/isbfcpp.h
```diff
@@ -0,0 +1,42 @@
...
+    int** matrix180;
+    int** matrix270;
+
+public:
+    isbfcpp();
+    int length();
+    void nth_from_last(int n, int &x, int &y);
+    bool addList(int x, int y);
+    void rotateMatrix(int rows, int cols, int **input, int **output);
+    std::vector<int> getList(int nrows, int ncols, int *size, int *mask, int startX, int startY, float inf);
```
cdeepakroy: maybe rename to `trace` or `traceBoundary`

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,344 @@
...
+        }
+        angle = M_PI;
+      }
+
+      // initialize cX and cY
+      // cX and cY indicate directions for each ISBF
+      vector<int> cX(1);
+      vector<int> cY(1);
+      cX.at(0)=0;
+      cY.at(0)=0;
```
cdeepakroy: `cX, cY` do not seem to change size. Why declare them as vectors instead of scalars?

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,344 @@
...
+              DY = 1;
+          }
+          else{
+              if(h[0][0] == 1){
+                  if(h[0][1] == 1){
+                      // inner corner at front
+                      cX.at(0) = 0;
+                      cY.at(0) = -1;
+                      cX.resize(2,(int)-1);
+                      cY.resize(2,(int)0);
```
cdeepakroy: My bad! `cx, cy` change in size here. 

[std::vector::pushback](http://www.cplusplus.com/reference/vector/vector/push_back/) instead of resize. pushback is like list-append in python.

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,344 @@
...
+                      DX = 0;
+                      DY = -1;
+                  }
+                  else{
+                      // inner-outer corner at front-left
+                      cX.at(0) = -1;
+                      cY.at(0) = -1;
+                      DX = 0;
+                      DY = 1;
```
cdeepakroy: Should `DY = -1`? See [here](http://histomicstk.readthedocs.io/en/latest/_modules/histomicstk/segmentation/label/TraceBounds.html)

histomicstk/segmentation/label/isbfcpp.cpp
cdeepakroy: See if you can use vector for inputs (ex: mask) too

histomicstk/segmentation/label/isbfcpp.h
```diff
@@ -0,0 +1,42 @@
...
+    int** matrix00;
+    int** matrix90;
+    int** matrix180;
+    int** matrix270;
+
+public:
+    isbfcpp();
+    int length();
+    void nth_from_last(int n, int &x, int &y);
+    bool addList(int x, int y);
```
cdeepakroy: Use std::vector for x and y coords of the boundary. 

Then you wont need the `addList` (use pushback function of std::vector) and `nth_from_last` (you can just index in the vector and get any element) functions.

plugin_tests/segmentation_label_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_isbf(self):
+
+        mask = np.array([[0, 0, 0, 0, 0, 0],
+                         [0, 0, 1, 1, 0, 0],
+                         [0, 1, 1, 1, 1, 0],
+                         [0, 1, 1, 1, 1, 0],
+                         [0, 1, 1, 1, 1, 0],
+                         [0, 0, 1, 1, 0, 0],
+                         [0, 0, 0, 0, 0, 0]], dtype=np.bool)
```
cdeepakroy: Try to use a test case with all cases:

* left neighbor
* inner-outer corner at left-rear
* inner corner at front
* inner-outer corner at front-left
* front neighbor
* outer corner

to make sure everything works.

histomicstk/segmentation/label/isbfcpp.cpp
slee172: Used the input mask as 2D vector.

histomicstk/segmentation/label/isbfcpp.cpp
slee172: Moved all pointers to vector/list

histomicstk/segmentation/label/isbfcpp.h
slee172: Renamed to traceBoundary

histomicstk/segmentation/label/isbfcpp.cpp
slee172: Used pushback(), so now the vector has one or two values for each cX and cY.

histomicstk/segmentation/label/isbfcpp.cpp
slee172: Weired. It didn't effect the test result, but fixed the typo. I think we need to change the test dataset that covers every case.

histomicstk/segmentation/label/isbfcpp.h
slee172: Good suggestion. Removed addList, nth_from_last functions. Instead, used the vector index.

histomicstk/segmentation/label/isbf.pyx
```diff
@@ -0,0 +1,38 @@
...
+cdef extern from "passInf.h":
+    pass
+
+cdef extern from "isbfcpp.h":
+    cdef cppclass isbfcpp:
+        isbfcpp() except +
+        int length()
+        void nth_from_last(int, int &x, int &y)
+        bool addList(int, int)
+        void roateMatrix(int, int, int **input, int **output)
```
cdeepakroy: typo rotateMatrix

slee172: did not push yet. It will be updated soon.

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,247 @@
...
+      }
+
+      // initialize cX and cY which indicate directions for each ISBF
+      std::vector<int> cX(1);
+      std::vector<int> cY(1);
+
+      if (h[1][0] == 1) {
+          // 'left' neighbor
+          cX.at(0) = -1;
+          cY.at(0) = 0;
```
cdeepakroy: You can just do cX[0] = -1 instead of cX.at(0).

slee172: changed !!

plugin_tests/segmentation_label_test.py
slee172: Done.

slee172:
@cdeepakroy Could you see [this](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/162035268#L2719)? Still got the import error.

cdeepakroy:
@manthey Can you please take a look? Adding [python setup.py build_Ext --inplace](https://github.com/DigitalSlideArchive/HistomicsTK/blob/1fe2ef9fc17b37bbc882c7f7ab5a3fd848cf80a1/.travis.yml#L91) seems to produce this [error](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/162035268#L2770).

cdeepakroy:
@manthey Nevermind, Looks like it got resolved.

slee172:
@cdeepakroy I put the test cases for isbf algorithm. Could you take a look into it?

histomicstk/segmentation/label/passInf.h
```diff
@@ -0,0 +1,5 @@
+#ifndef INFINITY
+
+#define INFINITY 0
+
+#endif
```
cdeepakroy: We probably do not need a separate file for this? 

Can you define it in the isbfcpp.h or the .pyx file itself?

slee172: I don't see a directly way to define it on .pyx, but I think it's possible to define it on isbfcpp.h. I'll try it.

plugin_tests/segmentation_label_test.py
```diff
@@ -0,0 +1,155 @@
...
+
+        # refenece left neighbor
+        rx = [1, 1, 1]
+        ry = [2, 1, 2]
+
+        # test left neighbor
+        m_left_neighbor = np.array([[0, 0, 0, 0],
+                                    [0, 1, 1, 0],
+                                    [0, 0, 0, 0],
+                                    [0, 0, 0, 0]], dtype=np.bool)
```
cdeepakroy: Since you use `DX = 1, DY = 0` as initial direction in your tracing algorithm, this would be front neighbor i think?

cdeepakroy: Check other like wise

slee172: Yes. It could be front neighbor if the starting point is (1,1)

histomicstk/segmentation/label/passInf.h
slee172: Done.

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,255 @@
...
+    // copy mask to matrix00
+    for(int i=0; i< nrows; i++){
+        for(int j=0; j< ncols; j++){
+            //matrix00[i][j] = mask[i*ncols+j];
+            matrix00[i][j] = mask[i][j];
+        }
+    }
+
+    // rotate matrix for 90, 180, 270 degrees
+    matrix270 = rotateMatrix(nrows, ncols, matrix00);
```
cdeepakroy: returning a vector from rotateMatrix and assigning it will result in a copy. You may want to pass in a reference to the output matrix as a function argument to speed up.

cdeepakroy: also maybe rename rotateMatrix to `rotateMatrix90` or `rot90` to be clear.

histomicstk/segmentation/label/trace_boundary.py
```diff
@@ -0,0 +1,166 @@
...
+    Returns:
+    --------
+    X : array_like
+        A 1D array of horizontal coordinates of contour seed pixels for
+        tracing.
+    Y : array_like
+        A 1D array of the vertical coordinates of seed pixels for tracing.
+
+    See Also:
+    ---------
```
cdeepakroy: Remove See Also if you don't have anything to put there.

cdeepakroy:
@slee172 Few more comments. Rest LGTM!

histomicstk/segmentation/label/isbfcpp.cpp
slee172: Used the function argument in a reference and the function name is changed to rot90.

histomicstk/segmentation/label/isbfcpp.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    // allocate memory for return value
+    std::vector <std::vector<int> > boundaries(2, std::vector<int>(sizeofX));
+
+    std::vector<int> bX (boundary_listX.begin(), boundary_listX.end());
+    std::vector<int> bY (boundary_listY.begin(), boundary_listY.end());
+
+    boundaries[0] = bX;
+    boundaries[1] = bY;
```
slee172: OK. Used the assign

histomicstk/segmentation/label/trace_boundary.py
slee172: Done.

cdeepakroy:
@slee172 Did you put a print statement for each case and check if all print statements are being executed by your test?

slee172:
It was tested at my local machine.

PR #205, opened 2016-09-16
Make sure we are on girder client 1.3.1.

manthey:
Girder work installs a version of girder client.  This script then installs a different version, which prevents some functions from working.


cdeepakroy:
LGTM!

Issue #206, opened 2016-09-20
Disable caching when pip installing via ansible


manthey:
We need to flush or ignore the pip cache when installing via ansible.  Otherwise, if Pillow was installed *prior* to installing openslide, then when we ask for Pillow to be installed we get a version which is cached and references the no-longer-existing old jpeg libraries instead of the newly generated jpeg libraries.

manthey:
@zachmullen Do you think we should just disable pip caching in general, or should we flush the pip cache?

@dgutman This is the root of the jpeg-version problem on the two installs you did.

zachmullen:
I'd say we should just flush it prior to doing the pip install.

dgutman:
Aah .. makes sense..
And definitely a bit of package management hell...

manthey:
Resolved in PR #207.

PR #207, opened 2016-09-20
Once we have built and installed openslide, purge the pip cache.

manthey:
Once we have built and installed openslide, purge the pip cache and explicitly uninstall Pillow if it is installed.

Ask pip for the location of the pip cache and remove that directory.  I feel this is brittle and heavy-handed.

If Pillow is installed, it will look for the old jpeg libraries rather than the ones we have just built.  If the cache is not cleared, pip will just reinstall Pillow from the cache and still look for the wrong libraries.

This addresses issue #206.

@zachmullen I've put this in at the end of the OpenSlide role, as that is when the pip cache tends to have something invalid in it.  I feel there should be a cleaner way to do this, and welcome suggestions or a completely different approach.

### DIFF
--- a/ansible/roles/openslide/tasks/main.yml
+++ b/ansible/roles/openslide/tasks/main.yml
@@ -122,3 +122,13 @@
 - name: Run ldconfig
   command: ldconfig
   sudo: yes
+
+- name: Purge pip cache
+  command: bash -c "if PIP_CACHE=`python -c 'import pip.locations, sys;sys.stdout.write(pip.locations.USER_CACHE_DIR)'`; then rm -r $PIP_CACHE; fi"
+  sudo: yes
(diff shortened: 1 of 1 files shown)


manthey:
@zachmullen I think clearing the pip cache and reinstalling libraries that are affected by changing libtiff and OpenJpeg is preferred to always ignoring the cache.

We need to clear the cache, as otherwise we have the danger that sometime later someone asks for a different Pillow version and gets the one with the bad library binding.  By reinstalling the pypi modules that are installed already, we avoid having stale modules right away (in case the openslide role is used by itself), and allow the pip cache to henceforth do its job at reducing web fetches.

manthey:
That's a fair characterization.

PR #212, opened 2016-10-02
Update unit tests for trace boundary

slee172:
Unit tests for trace boundary is updated. 

- Removed isbfcpp.cpp and isbfcpp.h
- Added trace_boundary_cpp.cpp and trace_boundary_cpp.cpp, which include the cython version of Moore neighbor tracing algorithm and ISBF algorithm
- Modified setup.py
- Modified plugin.cmake
- Updated segmentation_lable_test.py (unit tests for moore algorithm is added) 
- Modified trace_boundary.py


plugin_tests/segmentation_label_test.py
```diff
@@ -153,3 +155,157 @@ def test_trace_boundary(self):
...
+                               [0, 0, 0, 0, 0],
+                               [0, 0, 0, 0, 0]], dtype=np.bool)
+        x = 2
+        y = 2
+
+        by, bx = trace_boundary(m_neighbor, Connectivity=8,
+                                XStart=y, YStart=x)
+
+        np.testing.assert_allclose(rx, bx)
+        np.testing.assert_allclose(ry, by)
```
cdeepakroy: @slee172 Add a test where the object is at the boundary of the image and see if it works.

setup.py
```diff
@@ -80,9 +80,11 @@
...
       tests_require=test_requirements,
       ext_modules = cythonize(Extension(
-           "histomicstk.segmentation.label.isbf",
-           sources=["histomicstk/segmentation/label/isbf.pyx",
-                    "histomicstk/segmentation/label/isbfcpp.cpp"],
+           "histomicstk.segmentation.label.trace_object_boundary",
+           sources=[
+                "histomicstk/segmentation/label/trace_object_boundary.pyx",
+                "histomicstk/segmentation/label/trace_boundary_cpp.cpp"
+           ],
```
cdeepakroy: There seems to be a way to specify sources in the pyx file instead of setup.py. See [here](http://docs.cython.org/en/latest/src/userguide/wrapping_CPlusPlus.html#specify-c-language-in-setup-py). It will make setup.py much more cleaner.

histomicstk/segmentation/label/trace_boundary.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 import ctypes
-from .isbf import isbf
+from .trace_object_boundary import trace_object_boundary
 
 
 def trace_boundary(Mask, Connectivity=4, XStart=None, YStart=None,
```
cdeepakroy: Make trace_boundary work on a label mask (with mutliple objects) instead of a binary mask (with one object).

Currently i have to do the following for this:

```python
def get_label_contours(imLabel):
    
    label_id_list = np.unique(imLabel.flatten())

    obj_contours = {}

    for i in range(len(label_id_list)):

        label_id = label_id_list[i]
        
        # create binary mask for current label object
        # BUG: had to pad to make it work on objects at image boundary
        imMask = np.pad(imLabel == label_id, (1, 1), mode='constant')

        # get object boundary coords as an N X 2 array
        X, Y = trace_boundary(imMask)       

        coords = np.vstack((X, Y)).T - 1

        # add object contour to dictionary
        obj_contours[label_id] = coords  
        
    return obj_contours
```

It would be much faster to have that loop over object in cython or c++ file.

histomicstk/segmentation/label/trace_boundary_cpp.cpp
```diff
@@ -1,18 +1,18 @@
...
-isbfcpp::isbfcpp()
+trace_boundary_cpp::trace_boundary_cpp()
 {
 }
 
-void isbfcpp::rot90(int nrows, int ncols, std::vector <std::vector<int> > matrix,
+void trace_boundary_cpp::rot90(int nrows, int ncols, std::vector <std::vector<int> > matrix,
                     std::vector <std::vector<int> > &matrix270,
                     std::vector <std::vector<int> > &matrix180,
                     std::vector <std::vector<int> > &matrix90)
```
cdeepakroy: It is more modular to have rot90 take a matrix and output only one matrix that is rotated by 90 degrees.

plugin_tests/segmentation_label_test.py
slee172: It does not work because the Moore neighbor tracing needs to be back at the start point. For example, if there are two points (0,0), (0,1) and the start point (0,0), it will be back when beginning so that (0,-1) will cause an error. Also, isbf algorithm also has a condition to be considered. I think we should use a padding array before using it.

slee172:
@cdeepakroy Could you see [here](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/166226630#L884)? I think this is because of numpy.get_include(), so I added it to ext_modules in setpy.py as [here](http://cython.readthedocs.io/en/latest/src/reference/compilation.html), but it does not work. Do we need to go back and use Extension in setup.py if it does not work?

histomicstk/segmentation/label/trace_boundary.py
cdeepakroy: Also putting the loop over objects into cython or c++ will speed up things a lot.

cooperlab: @cdeepakroy @slee172 I think this is a good suggestion. Merge the functions and add a check to detect whether an image is label or binary.

cdeepakroy: Also, i tried the above function `get_label_contours` on a label mask with  over 1000 objects that @dgutman sent me with TraceBounds and trace_boundary.

When i used TraceBounds it took 160 seconds. When i used trace_boundary instead it tool 2400 seconds.

I think the main reason for this is the [rot90](https://github.com/DigitalSlideArchive/HistomicsTK/blob/0a7156c81f28f872916e7a2b29cb894897eabada/histomicstk/segmentation/label/trace_boundary_cpp.cpp#L15) which computes rotated versions of the input image is being called for each and every object. We only need to call this once.

Also, another reason is that the loop over objects is in python.

cooperlab: I think this is a great place to pickup speed - just pre-calculate 3 rotations at the beginning (90, 180, 270) and use these in subsequent calculations.

histomicstk/segmentation/label/trace_boundary_cpp.cpp
slee172: Done.

histomicstk/segmentation/label/trace_label.py
```diff
@@ -0,0 +1,12 @@
...
+from trace_object_boundary import trace_object_boundary
+
+
+def trace_label(imLabel, Connectivity=8):
+
+    imLabel = np.ascontiguousarray(imLabel, dtype=ctypes.c_int)
+
+    output = trace_object_boundary(imLabel, Connectivity)
+
+    return output
```
cdeepakroy: I don't think there is a need for trace_label anymore because trace_object_boundary is doing the same thing. So remove it.

histomicstk/segmentation/label/trace_boundary_cpp.h
```diff
@@ -0,0 +1,33 @@
...
+               std::vector <std::vector<int> > input,
+               std::vector <std::vector<int> > &output);
+    std::vector <std::vector<std::vector<int>> > trace_boundary(std::vector <std::vector<int> > imLables, int connectivity);
+    std::vector <std::vector<int> > isbf(int nrows, int ncols, std::vector <std::vector<int> > mask,
+      std::vector <std::vector<int> > mask_90, std::vector <std::vector<int> > mask_180,
+      std::vector <std::vector<int> > mask_270, int startX, int startY, float inf);
+    std::vector <std::vector<int> > moore(int nrows, int ncols, std::vector <std::vector<int> > mask,
+      std::vector <std::vector<int> > mask_90, std::vector <std::vector<int> > mask_180,
+      std::vector <std::vector<int> > mask_270, int startX, int startY, float inf);
+    ~trace_boundary_cpp();
```
cdeepakroy: We don't need a class as we are not using any object oriented features in this case. There are no member variables even.

plugin_tests/segmentation_label_test.py
slee172: Yes. So, I added some codes for padding [here](https://github.com/DigitalSlideArchive/HistomicsTK/pull/212/files#diff-20a5ab8844c79071c983e4a2e1d30fd0R100).

histomicstk/segmentation/label/trace_boundary_cpp.h
cdeepakroy: Just all the functions as pure C-like functions. Use a class only if you want any of object oriented features that come with it.

histomicstk/segmentation/label/trace_boundaries_opt.h
```diff
@@ -0,0 +1,21 @@
...
+           std::vector <std::vector<int> > &output);
+std::vector <std::vector<std::vector<int>> > trace_boundary(std::vector <std::vector<int> > imLables, int connectivity);
+std::vector <std::vector<int> > isbf(int nrows, int ncols, std::vector <std::vector<int> > mask,
+  std::vector <std::vector<int> > mask_90, std::vector <std::vector<int> > mask_180,
+  std::vector <std::vector<int> > mask_270, int startX, int startY, float inf);
+std::vector <std::vector<int> > moore(int nrows, int ncols, std::vector <std::vector<int> > mask,
+  std::vector <std::vector<int> > mask_90, std::vector <std::vector<int> > mask_180,
+  std::vector <std::vector<int> > mask_270, int startX, int startY, float inf);
+
+#endif
```
slee172: @cdeepakroy Made some commits for it. Could you see this? I'm not sure this is what you expect.

histomicstk/segmentation/label/trace_boundaries.pyx
```diff
@@ -0,0 +1,21 @@
...
+  vector[vector[vector[int]]] trace_boundary(vector[vector[int]] label, int)
+
+
+def trace_boundaries(np.ndarray[int, ndim=2, mode="c"] imLabel not None, connectivity=4):
+
+    cdef vector[vector[vector[int]]] output
+
+    output = trace_boundary(imLabel, connectivity)
+
+    return output
```
slee172: @cdeepakroy Your comments reduces the lines in pyx a lot !! Thanks

setup.py
```diff
@@ -55,7 +55,7 @@
 ]
 
 # cython extensions
-ext_compiler_args = ["-std=c++11"]
+ext_compiler_args = ["-std=c++11", "-O2"]
```
cooperlab: Any reason not to -O3?

setup.py
cdeepakroy: Read the last part of [this](http://docs.cython.org/en/latest/src/userguide/wrapping_CPlusPlus.html#specify-c-language-in-setup-py).

Seems like the options and sources can also be passed directly from the .pyx file as follows. This will make setup.py simpler and also we wont have to change the setup.py file each time we cythonize a new function. 

Here is a simplified setup.py file:

```python
from distutils.core import setup
from Cython.Build import cythonize

setup(
    name = "rectangleapp",
    ext_modules = cythonize('*.pyx'),
)
```
And in the .pyx source file, write this into the first comment block, before any source code, to compile it in C++ mode and link it statically against the Rectangle.cpp code file:

```python
# distutils: language = c++
# distutils: sources = Rectangle.cpp
```

histomicstk/segmentation/label/trace_boundaries.pyx
```diff
@@ -0,0 +1,21 @@
...
+
+cdef extern from "trace_boundaries_opt.h":
+  vector[vector[vector[int]]] trace_boundary(vector[vector[int]] label, int)
+
+
+def trace_boundaries(np.ndarray[int, ndim=2, mode="c"] imLabel not None, connectivity=4):
+
+    cdef vector[vector[vector[int]]] output
+
+    output = trace_boundary(imLabel, connectivity)
```
cdeepakroy: Are you padding to trace boundaries of object at image edges correctly? If not, you can do a 1-pixel padding for `imLabel` all round right before calling this function.

cdeepakroy: And then subtract the returned/output x and y coords by 1.

setup.py
slee172: cython automatically uses an optimization flag when compiling. So, will not have speedup for this. And, couldn't find any big difference between -O2 and -O3 at this time.

histomicstk/segmentation/label/trace_boundaries.pyx
slee172: Added a padding for each object using min-max of coordinates, and then finally added window offset [here](https://github.com/DigitalSlideArchive/HistomicsTK/pull/212/files#diff-22e72ae0b92c8b76a2c9bd57f2ff509bR85) Would it be better using it out of the function? If it is, I will add them before the function and then removed the previous work.

histomicstk/segmentation/label/trace_boundaries_opt.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+    // start ----- find unique id of labels
+
+    // 2d to 1d array
+    std::vector<int> im1D;
+    im1D.reserve(nrows_img * ncols_img);
+    for(int i = 0; i < nrows_img; i++) {
+       const std::vector<int>& v = imLabels[i];
+       im1D.insert(im1D.end(), v.begin(), v.end());
+    }
```
cdeepakroy: The creation of im1D is not needed to generate label_set.

Instead you can just do an iterator on imLabel and insert the elements into label_set directly.

histomicstk/segmentation/label/trace_boundaries.pyx
cdeepakroy: I think what you did is fine!

histomicstk/segmentation/label/trace_boundaries_opt.cpp
slee172: Removed 1D conversion of 2D matrix, and the label numbers are directly inserted into label_set. Thanks !!

setup.py
cdeepakroy: @slee172 Did you look at this?

PR #217, opened 2016-10-10
Move CLI generation to slicer_cli_web

jbeezley:


jbeezley:
I suppose I should add some tests for the main body view to get the coverage back up.

cdeepakroy:
@manthey Can you please QA this?

web_client/js/views/controlWidget.js
```diff
@@ -1,175 +0,0 @@
...
-    },
-
-    initialize: function () {
-        this.listenTo(this.model, 'change', this.render);
-        this.listenTo(this.model, 'destroy', this.remove);
-        this.listenTo(this.model, 'invalid', this.invalid);
-
-        // Handle image param types when no value is provided by using
-        // the currently open image.
-        if (this.model.get('type') === 'image' && !this.model.get('value')) {
```
manthey: Are we okay with getting rid of using the loaded image as a fallback?

jbeezley: That functionality was moved to the HistomicsTK [body view](https://github.com/DigitalSlideArchive/HistomicsTK/blob/bfa18b4bb28d3f56abadf7b302652e25bd9c4538/web_client/js/views/body.js).  I thought of it more as an issue of routing that belongs in the application rather than in the library here.

manthey: I see.

manthey:
LGTM.

Issue #219, opened 2016-10-13
Using std::next() results in a cython build error


cdeepakroy:
Running `python setup.py build_ext --inplace` or `conda develop --build_ext .` to build the cython code produces the following error:

> gcc -fno-strict-aliasing -I/Users/cdeepakroy/anaconda/envs/emory/include -arch x86_64 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/cdeepakroy/anaconda/envs/emory/lib/python2.7/site-packages/numpy/core/include -I/Users/cdeepakroy/anaconda/envs/emory/include/python2.7 -c histomicstk/segmentation/label/isbfcpp.cpp -o build/temp.macosx-10.5-x86_64-2.7/histomicstk/segmentation/label/isbfcpp.o -std=c++11 -stdlib=libstdc++ -mmacosx-version-min=10.6
> histomicstk/segmentation/label/isbfcpp.cpp:217:25: error: no member named 'next' in namespace 'std'
>         int fx2 = *std::next(boundary_listX.begin(), 1);
> histomicstk/segmentation/label/isbfcpp.cpp:219:25: error: no member named 'next' in namespace 'std'
>         int fy2 = *std::next(boundary_listY.begin(), 1);
> histomicstk/segmentation/label/isbfcpp.cpp:220:25: error: no member named 'prev' in namespace 'std'
>         int lx1 = *std::prev(boundary_listX.end());
> histomicstk/segmentation/label/isbfcpp.cpp:221:25: error: no member named 'prev' in namespace 'std'
>         int ly1 = *std::prev(boundary_listY.end());
> histomicstk/segmentation/label/isbfcpp.cpp:222:25: error: no member named 'prev' in namespace 'std'
>         int lx2 = *std::prev(boundary_listX.end(), 2);
> histomicstk/segmentation/label/isbfcpp.cpp:223:25: error: no member named 'prev' in namespace 'std'
>         int ly2 = *std::prev(boundary_listY.end(), 2);
> histomicstk/segmentation/label/isbfcpp.cpp:224:25: error: no member named 'prev' in namespace 'std'
>         int lx3 = *std::prev(boundary_listX.end(), 3);
> histomicstk/segmentation/label/isbfcpp.cpp:225:25: error: no member named 'prev' in namespace 'std'
>         int ly3 = *std::prev(boundary_listY.end(), 3);
> 8 errors generated.
[...]

PR #220, opened 2016-10-13
Specify compiler flag mmacosx-version-min to 10.9 to use libc++ 

cdeepakroy:
Building cython code that uses c++11 features on osx seems to result in build errors. See issue #219.

According to [this](http://stackoverflow.com/questions/36960587/cython-build-cant-find-c11-stl-files-but-only-when-called-from-setup-py) stackoverflow, the default of mmacosx-version-min to be 10.5 which was making it use libstdc++ instead of libc++. Hence it doesnt seem to recognize c++11 features. 

This PR sets mmacosx-version-min to 10.9 to fix it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,8 @@ setup(name='histomicstk',
            sources=["histomicstk/segmentation/label/isbf.pyx",
                     "histomicstk/segmentation/label/isbfcpp.cpp"],
            include_dirs=[numpy.get_include()],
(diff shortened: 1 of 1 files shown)


cdeepakroy:
CC: @cooperlab

cooperlab:
Thanks Deepak.

slee172:
This is what I wanted later. LGTM !!

PR #221, opened 2016-10-14
Crop images on the right and bottom so that only valid portions of the image are shown.

manthey:


web_client/js/views/visualization.js
```diff
@@ -284,7 +285,21 @@ histomicstk.views.Visualization = girder.View.extend({
...
-            tileRounding: Math.ceil
+            tileRounding: Math.ceil,
+            tilesAtZoom: function (level) {
+                var scale = Math.pow(2, opts.maxLevel - level);
+                return {
+                    x: Math.ceil(opts.sizeX / opts.tileWidth / scale),
+                    y: Math.ceil(opts.sizeY / opts.tileHeight / scale)
+                };
+            },
+            tilesMaxBounds: function (level) {
```
jbeezley: This callback is not getting called in the client test.  Are there any special conditions necessary for geojs to use it?

web_client/js/views/visualization.js
```diff
@@ -116,9 +114,12 @@ histomicstk.views.Visualization = girder.View.extend({
             ingcs: '+proj=longlat +axis=esu',
             gcs: '+proj=longlat +axis=enu',
             maxBounds: bounds,
+            // to appropriate set min and max, we need to know the tile size
```
jbeezley: to set min and max appropriately

web_client/js/views/visualization.js
manthey: I failed to update the version of large_image used by travis, so the latest version of geojs wasn't being used.

PR #234, opened 2017-01-06
When we install, some of our requirements were lowering library versions

manthey:
This could cause problems if we expect to work with specific versions.  For instance, we can work with Pillow 3.4 or 4.0 (but not if installed via conda), so don't revert to an earlier version.

### DIFF
--- a/requirements_c_conda.txt
+++ b/requirements_c_conda.txt
@@ -18,5 +18,6 @@ scipy==0.16.0
 scikit-learn==0.17.1
 pandas==0.18.1
 
-# We want Pillow 3.2.0.  Make sure conda doesn't change it on us.
-Pillow==3.2.0
+# Don't have conda install Pillow; we need a version that is linked against
+# our own libraries, and conda may not do that.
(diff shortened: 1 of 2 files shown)


requirements_c_conda.txt
```diff
@@ -18,5 +18,6 @@ scipy==0.16.0
 scikit-learn==0.17.1
 pandas==0.18.1
 
-# We want Pillow 3.2.0.  Make sure conda doesn't change it on us.
-Pillow==3.2.0
+# Don't have conda install Pillow; we need a version that is linked against
+# our own libraries, and conda may not do that.
+# Pillow==3.2.0
```
jbeezley: Should Pillow be added to `requirements.txt`?

manthey: We depend on large_image, which requires Pillow.

We can also list it in HistomicsTK if you think it should be there as well.

jbeezley: It looks like we don't import pillow here, so leaving it out is probably best.

PR #235, opened 2017-01-10
Girder 2 support

jbeezley:
This updates the HistomicsTK plugin to work with Girder 2.  As discussed in our last meeting, the web client (which is the majority of the changes here) is not finished, but there is interest for getting this merged anyway so we can start to deploy the entire DSA stack on Girder 2.  The endpoints and server-side triggers should be working correctly in this branch.

This contains one temporary commit which will be amended when the following PR's are merged:
* https://github.com/DigitalSlideArchive/large_image/pull/111
* https://github.com/girder/slicer_cli_web/pull/22


jbeezley:
@dgutman @cdeepakroy The last two commits implements the panel you asked about last week.  Currently, the region is not passed to the CLI's because there is no support on the server yet.

Note: this requires https://github.com/girder/large_image/pull/152

test

jbeezley:
@manthey @cdeepakroy I believe the end to end demo for nuclei detection is now working on this branch.  You just need to provide a file name that ends with the extension ".anot" and the annotation will be generated and displayed on the right panel.  You will need the branches [`roi-web-ui`](https://github.com/girder/slicer_cli_web/pull/32) of slicer_cli_web and [`render_annotations`](https://github.com/girder/large_image/pull/152) of large_image for it to work.

manthey:
The only issue I currently see is that when I start a new job, the list of previous annotations vanishes.  When the job completes, only my new annotation is displayed.  It is possibly I've failed to update something, though (but I think I'm on the latest of all branches).

jbeezley:
I found an issue related to calling `fetch` on the annotation collection incorrectly, but I'm not sure that was the cause of the issue you were seeing as I haven't been able to reproduce it.

manthey:
That appears to have fixed it.

manthey:
I'd like to get this branch merged to master, at which point only support Girder 2 going forward.  @cdeepakroy have you switched to Girder 2 yet?  Otherwise, we just need to fix the CI.

cdeepakroy:
Yes i have switched to girder-2. For algorithm dev it doesnt seem to cause any issues.

manthey:
:tada:

PR #236, opened 2017-01-18
Add a script to deploy via docker containers.

manthey:
Two docker containers are built, one for girder_worker and one for HistomicsTK.  Scripts optionally set the hosts file to refer to the docker host for mongo and rabbitmq, if desired.  The docker containers are based on the Girder-2 branches of all relevant repositories.

The girder/dev roles have been reapportioned to girder/provision roles with changes to the database all in the provision role.

The build tests won't work in the multiple docker-container deployment, as those tests expect direct control of mongo.

Ansible scripts have been modified to use the current recommended parameters (e.g., become rather than sudo).

More testing is needed: the scripts work when deployed via vagrant and via docker when locally built and using default options.

The docker images need to be published so that they do not need to be built locally.


manthey:
After experimentally trying things out, we have some added desired features and issues:

- [ ] Add external volume links (perhaps --external=(dir)[:(name)] which would mount internally to /opt/histomics/mounts/(name)
- [ ] memcached.  Docker should use the host's memcached if present
- [ ] Running docker jobs inside girder_worker doesn't work yet

kotfic:
@manthey, @danlamanna and I are starting to review this for merge,  would it be possible to rebase against or merge master?

manthey:
@kotfic master merged

Issue #239, opened 2017-02-07
git clone && vagrant up of master branch fails


cdeepakroy:
It produces the following error:

> TASK [girder : Install our external girder plugins] ****************************
[...]

PR #240, opened 2017-02-07
Fix versions used in ansible.

manthey:
The master branch is still on Girder 1.7.  Use the appropriate versions for ansible.

Fixes issue #239.

ping @cdeepakroy


cdeepakroy:
LGTM!

I just tried `git clone && vagrant up` in this branch and it succeeds. I confirm that this fix works.

PR #247, opened 2017-02-26
Implementation of Membrane Detection CLI

slee172:
Added membrane detection CLI.

- MembraneDetection.py : receives input image and saves membrane labeled image

- MembraneDetection.xml : sets cli environment.


slee172:
@cdeepakroy Could you check the "codecov/project" error above? It says " 1 untracked files changed. ".

slee172:
@cdeepakroy The problem above is solved.

cdeepakroy:
I would suggest wrapping all the logic for membrane segmentation into a function that can be placed in histomicstk.segmentation.membranes and just call that function on the membrane stain image in the CLI code.

cooperlab:
This is a good suggestion. We will also eventually need a method to assign membrane segments to nuclei.

slee172:
I would think so. Should we do wrapping now or after merging?

cdeepakroy:
sanghoon would be good to write the wrap function now.

histomicstk/segmentation/membranes/membrane_detection.py
```diff
@@ -0,0 +1,119 @@
...
+        Splitted image where positive values correspond to splitted points.
+    Branches : array_like
+        Branche image of where positive values correspond to membrane branches.
+    """
+
+    # membrane filtering
+    im_deviation, im_thetas = htk_shape_filters.membranefilter(
+        I, np.arange(min_sigma, max_sigma+1),
+        beta, c
+    )
```
cdeepakroy: Just use [skimage.filters.frangi](http://scikit-image.org/docs/dev/api/skimage.filters.html#frangi) instead of writing our own membranefilter.

cdeepakroy: It uses the same [logic](https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/_frangi.py#L74).

slee172:
This PR is outdated and may conflict with other updates. So I closed this PR.

PR #248, opened 2017-02-27
Implementation of Membrane segmentation ipython

slee172:
Added membrane segmentation ipython version.


server/MembraneDetection/MembraneDetection.py
```diff
@@ -0,0 +1,143 @@
...
+    im_membrane_stain = im_stains[:, :, 2].astype(np.float)
+
+    #
+    # Perform membrane enhancement
+    #
+    print('>> Performing membrane enhancement')
+
+    # membrane filtering
+    im_deviation, im_thetas = htk_shape_filters.membranefilter(
+        im_membrane_stain, [1, 2, 3], 4, 2
```
cdeepakroy: We need to expose sigma's, beta, and c as parameters to the CLI.

For sigmas, you can ask for sigma range (min_sigma, max_sigma) and the number of steps between min and max values. Then you can use numpy's linspace function to generate the sigma vector.

server/MembraneDetection/MembraneDetection.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    im_deviation = 255 * im_deviation / im_deviation.max()
+
+    # segmentation
+    im_mask = im_deviation > args.foreground_threshold * \
+        skimage.filters.threshold_otsu(im_deviation)
```
cdeepakroy: Why not just use the otsu threshold? Why are we multiplying it by a factor?

server/MembraneDetection/MembraneDetection.xml
```diff
@@ -0,0 +1,99 @@
...
+  </parameters>
+  <parameters>
+    <label>Membrane segmentation</label>
+    <description>Membrane segmentation parameters</description>
+    <double>
+      <name>foreground_threshold</name>
+      <label>Foreground Intensity Threshold</label>
+      <description>Intensity value to use as threshold to segment foreground in membrane stain image</description>
+      <longflag>foreground_threshold</longflag>
+      <default>0.4</default>
```
cdeepakroy: This parameter is being used as a multiple of the otsu threshold. Yet, here it says "intenisty value to use as threshold"

server/MembraneDetection/MembraneDetection.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # label detection
+    im_split = im_skeleton & ~sp.ndimage.binary_dilation(
+        im_branches,
+        structure=skimage.morphology.disk(args.branch_split_dilation)
+    )
+    im_split = sp.ndimage.binary_dilation(
+        im_split,
+        structure=skimage.morphology.disk(args.branch_merge_dilation)
+    )
```
cdeepakroy: Would a morphological closing operation be better here instead of dilation?

server/MembraneDetection/MembraneDetection.py
slee172: from my test on IHC image, the otsu threshold only does not work properly, so made a scale factor for the users.

server/MembraneDetection/MembraneDetection.xml
slee172: right, I need your suggestion to use the threshold properly on IHC image.

server/MembraneDetection/MembraneDetection.py
slee172: Good suggestion. Let me test it.

slee172: @cdeepakroy just wanted to make sure that the code the right above is not to make an erosion but to dilate only,

server/MembraneDetection/MembraneDetection.py
```diff
@@ -0,0 +1,143 @@
...
+def main(args):
+
+    #
+    # Read Input Image
+    #
+    print('>> Reading input image')
+
+    im_input = skimage.io.imread(args.inputImageFile)[:, :, :3]
+    size_min = min(im_input.shape[0], im_input.shape[1])
+    im_input = im_input[0:size_min, 0:size_min, :]
```
cdeepakroy: Why only restring to square shaped images? The rest of the code looks generic.

server/MembraneDetection/MembraneDetection.py
```diff
@@ -0,0 +1,143 @@
...
+        axis=2
+    )
+    im_membraned_color[px, py, :] = 255
+
+    #
+    # Save output filtered image
+    #
+    print('>> Outputting membrane labeled image')
+
+    skimage.io.imsave(args.outputMembraneLabelFile, im_membraned_color)
```
cdeepakroy: Should we have the output as one of the following:
* a labeled mask mask where 1 indicates membrane pixel, and 2 indicates junction/branch points
* an adjacency graph and the list of verticles that are junction/branch points

I think the choice depends on how we want to use this result (e.g. feature extraction within membrane area)

server/MembraneDetection/MembraneDetection.py
cdeepakroy: Are you doing this to close gaps? If so, morphological closing might be a better option.

server/MembraneDetection/MembraneDetection.py
slee172: Added parameters (sigma max, sigma min, beta, c) to cli.

server/slicer_cli_list.json
```diff
@@ -13,5 +13,9 @@
 
   "ComputeNucleiFeatures": {
     "type"    : "python"
+  },
+
+  "MembraneDetection": {
```
cdeepakroy: Rename to `DetectMembranes`

server/MembraneDetection/MembraneDetection.py
slee172: No. just wanted to test it on square image.

cdeepakroy:
The final [output](https://github.com/DigitalSlideArchive/HistomicsTK/blob/membranefilter-ipython/docs/examples/membrane-segmentation.ipynb) looks very noisy.

server/MembraneDetection/MembraneDetection.py
slee172: Let's talk about this tommrrow meeting.

slee172:
@cdeepakroy The result is exactly what I expected.. because it should be tested on different image. We need to find IHC image.

server/MembraneDetection/MembraneDetection.py
slee172: No, used it to make dilation only which will be used as input to be labeld.

cooperlab:
@cdeepakroy Eventually we should have metadata stored with images in Girder that capture stain information (localization, stain type, antibody etc.) Nothing prevents people from applying membrane detection to non-membrane stained images, so perhaps we could find a way to detect failure with a quality score.

cooperlab:
@slee172 We have a number of images stored on the server that have membrane staining. Let's get some screenshots from one of these.

server/slicer_cli_list.json
slee172: Renamed.

slee172:
Merged to membranefilter-cli

PR #249, opened 2017-03-07
Make NucleiDetection CLI work on WSI, ROIs and run in parallel using Dask

cdeepakroy:
This PR makes NucleiDetection CLI work on both small/simple png/jpg images and large whole-slide images. Furthermore, it can also be used to detect nuclei within an ROI (must be specified in base_pixels space) of the image.

The following additional parameters have been added to the CLI:

* **Analysis ROI** - specified the region-of-interest within which the analysis should be done. Must be a four element vector in the format "left, top, width, height" in the space of the base layer. Default behavior is to process the whole image.

*  **Analysis Tile Size** - The tile size to be used for block wise analysis.

* **Analysis Magnification** - The magnification (default: 20x) at which the analysis should be done.

* **Minimum foreground fraction** - minimum fraction of foreground that should be present in a tile for it to be analyzed.

* **LAB Mean and stddev of reference image** - for reinhard color normalization

* **Dask scheduler address** - Must be in the format "<ip-address>:<port>". By default it will use all but one core on the machine the CLI is running on.

The CLI runs error free from the command-line.


cdeepakroy:
When run from inside docker image using docker run it produces the following warning but still produces the correct result

```
test_data/L1.png: Not a TIFF or MDI file, bad magic number 20617 (0x5089).
ERROR:root:pylibmc exception
Traceback (most recent call last):
  File "/build/miniconda/lib/python2.7/site-packages/large_image/server/cache_util/memcache.py", line 106, in __getitem__
    return self._Cache__data[hexVal]
  File "/build/miniconda/lib/python2.7/site-packages/pylibmc/client.py", line 156, in __getitem__
    value = self.get(key)
ConnectionError: error 3 from memcached_get(d8ecf46a701fe737fcf05d6abaae8518): (0x128ead0) CONNECTION FAILURE(Connection refused),  host: 127.0.0.1:11211 -> libmemcached/connect.cc:156
ERROR:root:Memcached ServerDown exception
Traceback (most recent call last):
  File "/build/miniconda/lib/python2.7/site-packages/large_image/server/cache_util/memcache.py", line 125, in __setitem__
    self._Cache__data[hexVal] = value
  File "/build/miniconda/lib/python2.7/site-packages/pylibmc/client.py", line 163, in __setitem__
    if not self.set(key, value):
ServerDown: error 47 from memcached_set: (0x128ead0) SERVER HAS FAILED AND IS DISABLED UNTIL TIMED RETRY,  host: 127.0.0.1:11211 -> libmemcached/connect.cc:720
```

cdeepakroy:
@cooperlab Can you please review?

server/NucleiDetection/NucleiDetection.xml
```diff
@@ -110,4 +121,46 @@
...
+  <parameters>
+    <label>General</label>
+    <description>General parameters</description>
+    <region>
+      <name>analysis_roi</name>
+      <label>Analysis ROI</label>
+      <description>Region of interest within which the analysis should be done. Must be a four element vector in the format "left, top, width, height" in the space of the base layer. Default behavior is to process the whole image.</description>
+      <longflag>analysis_roi</longflag>
+    </region>
+    <double>
```
cdeepakroy: @jbeezley See above. This is how the CLI will accept the ROI -- as a comma-separated list of four values "left, top, width, height" in the base layer (highest magnification) image space.

manthey:
We can use python 2.7.13 if we are on the girder 1.x-maintenance branch instead of the v1.7.0 release.

cdeepakroy:
At this moment, since everything works with python 2.7.12 as well, I don't mind holding the upgrade until we have completely migrated ti girder-2.0.

server/NucleiDetection/NucleiDetection.xml
```diff
@@ -110,4 +121,46 @@
       <default>80</default>
     </double>
   </parameters>
+  <parameters>
```
jbeezley: The parameter list is getting fairly long.  I would make this (and perhaps others) `<parameters advanced="true">` to hide it by default.

server/NucleiDetection/NucleiDetection.xml
jbeezley: Because this is an optional parameter, it should specify a default value, even if it is `0,0,0,0` to indicate "process the whole image".

server/NucleiDetection/NucleiDetection.xml
cdeepakroy: Done!

server/NucleiDetection/NucleiDetection.xml
cdeepakroy: Made it -1,-1,-1,-1. Let me know if this is ok on the front-end side.

jbeezley: That should work thanks.

PR #260, opened 2017-03-22
Make complement_stain_matrix compute the cross-product.

neal-siekierski:
What `complement_stain_matrix` currently computes is not the cross-product.
This change uses NumPy's `cross` method to compute the desired cross-product.


cdeepakroy:
@cooperlab Can you please review this?

cooperlab:
LGTM.

The original I believe is from the paper (which has some mathematical errors). We rarely use the complement so I don't think this was causing any issues.

neal-siekierski:
In this repo itself, the only use of `complement_stain_matrix` is used in the (non-adaptive) `color_deconvolution` routine if only two stain vectors are passed (i.e., the third is all zeroes).

PR #262, opened 2017-03-24
Use the master large_image branch.

manthey:
Fix the version of Pillow used by conda, otherwise, JPEG2K images may not process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,11 +20,17 @@ RUN conda config --add channels https://conda.binstar.org/cdeepakroy && \
     --file requirements_c_conda.txt && \
     pip install -r requirements.txt -r requirements_c.txt && \
     # Install large_image
-    pip install 'git+https://github.com/girder/large_image@girder-1.7#egg=large_image' && \
+    pip install 'git+https://github.com/girder/large_image#egg=large_image' && \
     # Install HistomicsTK
     python setup.py install && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,13 +20,19 @@ RUN conda config --add channels https://conda.binstar.org/cdeepakroy && \
...
     # clean up
     conda clean -i -l -t -y && \
     rm -rf /root/.cache/pip/*
 
+# Ensure we have a locally built Pillow and libtiff in conda's environment
+RUN pip install --upgrade --no-cache-dir --force-reinstall --ignore-installed \
+    Pillow \
+    libtiff && \
+    rm -rf /root/.cache/pip/*
+
```
cdeepakroy: Should we do this in base_docker_image instead?

manthey: We have to do it after the `conda install` commands.

manthey: I'll see if I can get conda to not update libtiff and Pillow if they were installed via the base_docker_image.  It will take a while (base_docker_image builds ITK, which is time consuming).

cdeepakroy: Other that this rest of the stuff looks good to me. Feel free to merge this and do that in another PR, if you want.

manthey: Okay.  I'll merge this, then.

PR #266, opened 2017-04-03
Unpin versions.

manthey:
Use pip for installation of numpy and scipy, as the conda build conflicts with our home-built openjpeg and libtif.  Add tests in the docker build process that will fail when the libraries don't work together.


cdeepakroy:
Also, update HistomicsTK's [install documentation](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/docs/installation.rst#installing-histomicstk-as-a-python-toolkit).

cdeepakroy:
LGTM!

PR #279, opened 2017-04-06
Add large_image support to ColorDeconvolution

neal-siekierski:
As we work out how to nicely get the results of deconvolving large images, I'd like to get things going with these couple of commits.

These support specifying a region and extracting it from a large image, and writing out to the argument specified by `--returnparameterfile` (if specified) the relevant information: the paths to the output files and the region.


server/ColorDeconvolution/ColorDeconvolution.xml
```diff
@@ -60,5 +60,42 @@
...
+      <label>Large Image?</label>
+      <longflag>largeImage</longflag>
+      <description>True if the input image is a large image that should be handled with large_image</description>
+    </boolean>
+    <integer-vector>
+      <name>region</name>
+      <label>Region</label>
+      <longflag>region</longflag>
+      <description>left,top,width,height of the region of interest</description>
+    </integer-vector>
```
cdeepakroy: Use the region type in Slicer XML Spec. See example [here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/server/NucleiDetection/NucleiDetection.xml#L127-L133_

server/ColorDeconvolution/ColorDeconvolution.xml
```diff
@@ -60,5 +60,42 @@
...
+    <!-- And now the output parameters -->
+    <integer-vector>
+      <name>region</name>
+      <label>Region</label>
+      <channel>output</channel>
+      <description>left,top,width,height of the region of interest</description>
+    </integer-vector>
```
cdeepakroy: Use the region type here too. Need to test this end-to-end over the web to see if this gets passed along or not. I would suggest drop this until Dave adds writeImage support to large_image that abstracts passing in the location. We can then make a PR.

@manthey @jbeezley What are your thoughts on this?

server/ColorDeconvolution/ColorDeconvolution.xml
cdeepakroy: Also make this a default parameter with value `<default>-1,-1,-1,-1</default>` to mean the whole image.

server/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -16,7 +19,17 @@ def main(args):
...
-    im_input = skimage.io.imread(args.inputImageFile)
+    if args.largeImage:
+        if args.region is None:
+            raise ValueError('Region must be provided when using --largeImage')
+        ts = large_image.getTileSource(args.inputImageFile)
+        if len(args.region) != 4:
+            raise ValueError('Exactly four values required for --region')
+        im_input = ts.getRegion(format=large_image.tilesource.TILE_FORMAT_NUMPY,
+                                region=dict(zip(['left', 'top', 'width', 'height'],
+                                                args.region)))[0]
```
cdeepakroy: You can just use this for both normal and whole-slide images. No need for this if.

For the moment, you can use this for testing if the image is a whole-slide image.

server/ColorDeconvolution/ColorDeconvolution.xml
```diff
@@ -60,5 +60,42 @@
       <longflag>stainColor_3</longflag>
       <default>0,0,0</default>
     </double-vector>
+    <boolean>
+      <name>largeImage</name>
+      <label>Large Image?</label>
+      <longflag>largeImage</longflag>
+      <description>True if the input image is a large image that should be handled with large_image</description>
+    </boolean>
```
cdeepakroy: No need for this bool parameter. You can use [this](you can use [this](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/server/NucleiDetection/NucleiDetection.py#L225) for testing if the image is a whole-slide image) for testing if the image is a whole-slide image.

server/ColorDeconvolution/ColorDeconvolution.py
cdeepakroy: You may also want to raise an error if the region width and height is above a certain value to disallow people from running it on large regions or entire slide.

server/ColorDeconvolution/ColorDeconvolution.xml
```diff
@@ -60,5 +60,37 @@
...
+      <default>-1,-1,-1,-1</default>
+    </region>
+    <!-- And now the output parameters -->
+    <region>
+      <name>region</name>
+      <label>Region</label>
+      <channel>output</channel>
+      <description>left,top,width,height of the region of interest.  All -1 means the whole image was used.</description>
+    </region>
+    <string>
```
cdeepakroy: Let just use the `file` type for these outputs and hand it over to Dave as it is still not clear what is the most convenient way for the front-end. I am hoping the write support in large_image will provide a way to associate it will location information.

neal-siekierski: The output images are already file (actually, `image`) parameters as well. Should I drop the support for `--returnparameterfile` altogether, and not re-output the region and output file paths for now?

cdeepakroy: Yes, lets use image parameters and drop the returnparameterfile for now until we figure what would be convenient on the front-end side.

PR #281, opened 2017-04-07
Cythonize trace object boundaries fixing multi-label tracing issue

slee172:
Created cython version of trace object boundaries, and tested it on plugin_tests/segmentation_label_test. 
This version solves the multi-label tracing problem which occurs when a boundary tracing algorithm performs without any restrictions (e.g. without padding for each label box).


cdeepakroy:
@slee172 The membrane stuff is included in this. Can you please make a PR of just the fix to trace_object_boundaries?

We will do the membrane detection stuff from the other PR you have.

slee172:
This PR is closed because it includes the other branch commits.

PR #280, opened 2017-04-06
WIP: Add adaptive deconvolution methods (and some related fixes and enhancements)

neal-siekierski:
The main effect of this PR is to add adaptive deconvolution methods, represented currently by:
- The functions in `histomicstk.preprocessing.color_deconvolution` named `macenko_stain_matrix`, `snmf_stain_matrix`, and `rgb_macenko_stain_matrix`.
- The CLI `MacenkoStainMatrix`.

It also introduces the notion of "stain darkness" (SDA) as defined in Van Eycke et al. (Nature, 2017), represented by the following:
- Two new functions in `histomicstk.preprocessing.color_conversion`, `rgb_to_sda` and `sda_to_rgb`.
- A function to compute the involved background intensity measurement, `histomicstk.preprocessing.color_normalization.background_intensity`.
- Backwards-compatible changes to `color_convolution` and `color_deconvolution` in `histomicstk.preprocessing.color_deconvolution` to accept an additional parameter for background intensity, `I_0`.

In the process, `histomicstk.utils.sample_pixels` has been enhanced in the following ways:
- To allow sampling of the background as well as the foreground.
- To allow specifying a(n approximate) number of pixels to sample, instead of a percentage.

This PR also fixes some typos in related documentation and code, including one in `histomicstk.utils.sample_pixels` that prevented sampling at resolutions other than fully magnified.

The most noticeable absence is of the other recommended method in Van Eycke et al., "Li-init SNMF". The "Li-init" method is not yet implemented.

PR #283, opened 2017-04-07
Fix doc references

neal-siekierski:
I noticed that many of the references are syntactically comments, which means they don't show up in the documentation. Also, because of how Sphinx scopes target names, numbering these references per-file creates warnings (cf. how the Travis CI build of #280 was failing when my changes introduced additional references named [1] and [2] in the same module). reST fortunately supports auto-numbering, which handles that.

This PR should fix these issues.


cdeepakroy:
Great! LGTM! Merge Away!

PR #287, opened 2017-04-11
setup: use scikit-build for building nuclear and label C extensions

jcfr:
This commit introduces a new dependencies to "scikit-built" and "cmake".

scikit-build is a drop-in replacement to setuptools.setup function
allowing to easily compile and package extensions (C/C++/Cython) by
bridging CMake and setuptools. See http://scikit-build.org

CMake is is an open-source, cross-platform family of tools designed
to build, test and package software. See https://cmake.org

Currently, scikit-build and cmake have to be explicitly (see [1]) installed
on the system. This could be done by simply doing:

  pip install scikit-build cmake

How does it work ?
------------------

In addition to simplifying setup.py, the following files have been added:

```
<root>
  |
  |---- CMakeLists.txt                                   (1)
  .
  .
  .
  |--- ...histomicstk/segmentation/label
  |                                 |--- CMakeLists.txt  (2)
  .
  .
  |--- ...histomicstk/segmentation/nuclear
  |                                 |--- CMakeLists.txt  (3)
```

The first CMakeLists.txt specifies requirements for

* Python interpreter,

* the associated python libraries

* and a "PythonExtension" modules containing convenience CMake
  functions allowing to easily build extension by encapsulating
  system introspection and platform specific logic.

and include the subdirectory associated with the other CMakeLists.txt.

The second and third CMakeLists.txt are specific to "label" and "nuclear"
segmentation C extensions and they specify:

* requirement for NumPy

Then, they declare:

* libraries for each extension, associate python extension specific properties

* and finally add an install rule specifying where in the package
  hierarchy the extension should be installed.

Et voila,

[1] Note that improvement to python packaging will be available shortly and
it will be possible to include scikit-build and cmake directly in pyproject.toml
See here for more details: https://www.python.org/dev/peps/pep-0518/

Co-authored-by: Deepak Chittajallu <deepak.chittajallu@kitware.com>


.travis.yml
```diff
@@ -106,10 +106,10 @@ before_install:
 install:
     - cd $girder_path
     - python setup.py develop
-    - pip install --upgrade 'git+https://github.com/cdeepakroy/ctk-cli'
-    - pip install -r $main_path/requirements.txt
     - cd $main_path
-    - python setup.py build_ext --inplace
+    - pip install --upgrade 'git+https://github.com/cdeepakroy/ctk-cli'
```
jcfr: Adding 

`- pip install cmake`

should fix the build.

jcfr:
Since "pip install cmake" is now working, you could try the following:

```
pip install scikit-ci-addons
ci_addons travis/install_cmake.py 3.8.0
```

See http://scikit-ci-addons.readthedocs.io/en/latest/addons.html#travis

.travis.yml
```diff
@@ -106,19 +106,19 @@ before_install:
...
-    - pip install --upgrade 'git+https://github.com/cdeepakroy/ctk-cli'
-    - pip install -r $main_path/requirements.txt
+    - cd $large_image_path
+    - pip install -U -r requirements.txt
+    - python setup.py install
     - cd $main_path
-    - python setup.py build_ext --inplace
+    - pip install --upgrade 'git+https://github.com/cdeepakroy/ctk-cli'
+    - pip install -r requirements.txt
+    - python setup.py install
```
jcfr: If you do not generate and install a wheel, you have to install the package in development mode.

```
pip install -e .
```

or 

```
python setup.py develop
```

jcfr: This will ensure the "*.so" are also available in the source tree

.travis.yml
```diff
@@ -88,7 +88,7 @@ before_install:
     - mongod --dbpath=/tmp/db >/dev/null 2>/dev/null &
     - mongod --version
 
-    - CACHE=$HOME/.cache CMAKE_VERSION=3.1.0 CMAKE_SHORT_VERSION=3.1 source $girder_path/scripts/install_cmake.sh
+    - CACHE=$HOME/.cache CMAKE_VERSION=3.8.0 CMAKE_SHORT_VERSION=3.1 source $girder_path/scripts/install_cmake.sh
```
manthey: You need to change the short version, too.

.travis.yml
```diff
@@ -106,19 +106,19 @@ before_install:
 install:
     - cd $girder_path
     - python setup.py develop
-    - pip install --upgrade 'git+https://github.com/cdeepakroy/ctk-cli'
-    - pip install -r $main_path/requirements.txt
+    - cd $large_image_path
+    - pip install -U -r requirements.txt
+    - pip install -e .
```
manthey: Is there a reason you are installing large_image in development mode?  I've found that when in development mode, the HistomicsTK modules don't find the large_image modules unless you mess with PYTHONPATH.

Instead of these two lines, `python setup.py install` should be sufficient.

manthey: And, we probably want to do this after the installation of openslide, Pillow, and libtiff, since they rely on our built libraries and that way we don't have to uninstall and reinstall them.

jcfr:
👍

PR #288, opened 2017-04-11
Add cdog - a constrained difference-of-gaussian filter

cdeepakroy:
This PR implements a faster alternative to the constrained laplacian-of-gaussian filter (used for nuclei detection) based on a difference-of-guassians. 

The implementation is similar in concepts to the SIFT interest point detection (See [here](http://docs.opencv.org/trunk/da/df5/tutorial_py_sift_intro.html)). It divides the scale space into octaves and each octave is divided into a set of discrete levels such that the sigma/scale is doubled after each octave. The implementation also uses cascaded gaussian convolutions (`G(a) * G(b) = G(sqrt(a^2 + b^2))`) and downsamples the images by 2 at the end of each octave to keep the convolutional kernels small.


cdeepakroy:
@cooperlab Can you please review this PR?

Below is the comparison of the results of clog and cdog responses overlaid with estimated radius of seeds detected by max clustering.

image

image

histomicstk/filters/shape/clog.py
```diff
@@ -38,39 +42,52 @@ def clog(im_input, mask, sigma_min=30 * 1.414, sigma_max=50 * 1.414):
...
-    # translate filtered image
+            im_log_max[max_update_pixels] = im_log_cur[max_update_pixels]
+            im_sigma_max[max_update_pixels] = sigma
 
     # replace min floats
-    im_log_max[im_log_max == np.finfo(Distance.dtype).min] = 0
+    im_log_max[im_log_max == MIN_FLOAT] = 0
 
-    return im_log_max
+    return im_log_max, im_sigma_max
```
cdeepakroy: I am now returning the sigma values corresponding to the maximal response across scales for each pixel. 

We can use `sigma * sqrt(2)` to get an estimate of the radius.

histomicstk/filters/shape/cdog.py
```diff
@@ -0,0 +1,173 @@
...
+    n_octave = 0
+
+    for i in range(k + 1):
+
+        # calculate sigma at next level
+        sigma_next = sigma_cur * sigma_ratio
+
+        # Do cascaded convolution to keep convolutional kernel small
+        # G(a) * G(b) = G(sqrt(a^2 + b^2))
+        sigma_conv = np.sqrt(sigma_next ** 2 - sigma_cur ** 2)
```
cdeepakroy: Here is where i am using the cascaded gaussian approximation

PR #291, opened 2017-04-13
Introduce reference stain vectors and systematically use them in CLIs

neal-siekierski:
This PR contains changes to introduce pre-defined stain vectors into the `histomicstk` package and teach the `ColorDeconvolution`, `SparseColorDeconvolution`, and `NucleiDetection` CLIs to accept either them or custom stain vectors.

Changes:
- Create the dict `stain_color_map` in `histomicstk.preprocessing.color_deconvolution`
- Define utility argument parsing functions `get_stain_vector` and `get_stain_matrix` in `histomicstk.preprocessing.color_deconvolution.utils`
  - This is done to allow sharing of code between the CLIs. I'm not sure what would be a better location. `histomicstk.utils` as a location would create recursive import problems.
- Have CLIs take parameters of the form `stain_#` and `stain_#_vector`, of which the former indicates a name which, if `"custom"`, causes the latter to be used.
  - Parameters have been defaulted to H&E, with an empty "null" vector as a default if a third argument is provided.
- The documentation for `SparseColorDeconvolution`'s `beta` parameter is fixed.


histomicstk/preprocessing/color_deconvolution/__init__.py
```diff
@@ -4,19 +4,29 @@
...
+from . import stain_color_map as _stain_color_map
 from .color_convolution import color_convolution
 from .complement_stain_matrix import complement_stain_matrix
 from .sparse_color_deconvolution import sparse_color_deconvolution
 
 # must be imported after ComplementStainMatrix
 from .color_deconvolution import color_deconvolution
 
+stain_color_map = _stain_color_map.stain_color_map
+__doc__ += _stain_color_map.__doc__
```
cdeepakroy: Instead of lines 7, 15, 16 cant we just do

```
from .stain_color_map import stain_color_map
```

histomicstk/preprocessing/color_deconvolution/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    """Get the stain matrix corresponding to the args.stain_$index and
+    args.stain_$index_vector arguments for values of index 1 to count.
+    Return a numpy array of column vectors.
+
+    """
+    return numpy.array([get_stain_vector(args, i+1) for i in range(count)]).T
+
+__all__ = (
+    'get_stain_vector',
+)
```
cdeepakroy: What i don't like about this is the things only needed by the CLI are getting into the core python toolkit.

Now i see why you wanted a server.utils.

Can we put a `utils.py` file in the server directory with these functions and access them in the CLI's python file with a relative import?

```
from .. import utils as server_utils
```

neal-siekierski: Unfortunately it's more complicated. Relative imports only work in packages, something which the scattered executable files in `/server` are not. Cf. http://stackoverflow.com/a/73149 and comments.

I can think of a couple ways to try dealing with it, but they all require some reorganization.
- Create a secondary package alongside `histomicstk` -- `histomicstk_server` or some such. It would also have to be installed or otherwise on the `PYTHON_PATH` for `from histomicstk_server import utils` or the like to work.
- Organize the various CLIs into a package. This will require adding `__init__.py`s in several folders, and invoking the CLIs with `python -m path.to.CLI.CLI` instead of `python /path/to/CLI/CLI.py` (some variation of which is how I presume the CLIs are being immediately invoked currently). I'm not sure what all else would have to be changed for such a new launching scheme to work.

histomicstk/preprocessing/color_deconvolution/__init__.py
neal-siekierski: I'm doing it this way to get the docstring. Since variables can't have docstrings, my understanding is that module-level documentation is the way to go. In order to keep that documentation when we flatten the module hierarchy, I'm manually appending it to the docstring of the `color_deconvolution` module.

jbeezley: You can document variables in modules or classes for sphinx generated documentation using syntax like this:
```python
#: this is my description
foo
```
or 
```python
foo #: this is my description
```

neal-siekierski: @jbeezley, does Sphinx transfer that across module imports, which is the situation here? Or will I need to do something like
```
#: My description
stain_color_map = _stain_color_map.stain_color_map
```
in `__init__.py` to get that to work?

jbeezley: In that case, I think you would have to add the comment again.  If however, you were inheriting static variables from a class, it would inherit the documentation as well.  e.g.

```python
class A
    foo #: my variable

class B(A):
    pass
```
`B.foo` would be documented correctly

histomicstk/preprocessing/color_deconvolution/utils.py
cdeepakroy: It seems like overkill. Maybe we just do this in the CLI code

```
args_dict = vars(args)

stain_color_list = []

for i in range(3):
    
    if args_dict['stain_' + str(i+1)] == 'custom':
        stain_color_list.append(args_dict['stain_' + str(i+1) + '_vector']
    else:
        stain_color_list.append('stain_' + str(i+1))
                                
w = np.array(stain_color_list).T  
```

@manthey Do you have any ideas what would be a cleaner way to share some routines among different CLIs?

cdeepakroy: @neal-siekierski Apart from this one issue, everything else LGTM!

manthey: Another way to have common cli-only function would be to create a directory in `server/cli_common`, put an `__init__.py` script in it, plus whatever modules you want to have with cli-only routines (for the example below, suppose there is a file called `server/cli_common/common_cli_functions.py`).  Inside each primary CLI script that needs to access them (such as `server/ColorDeconvolution/ColorDeconvolution.py`), you can access these functions by flogging the python sys path:
```
sys.path.append(os.path.realpath(os.path.join(os.path.dirname(__file__), '../cli_common')))

import common_cli_functions
```

neal-siekierski:
@manthey The `sys.path` modification is working fine in principle, but for better or worse the relocated `server/cli_common/utils.py` is getting counted by codecov.io. Thoughts on how to manage this?

manthey:
@neal-siekierski Ideally, we would write tests so it would be properly covered.  Otherwise, I think we can add an exclusion rule (but I don't know how to off hand).  @jbeezley has fixed the codecov reporting for the rest of the of the repo as part of PR #296, so perhaps he knows.

cdeepakroy:
This [fix](https://github.com/DigitalSlideArchive/HistomicsTK/pull/296/commits/ae84e82082107cc0ae54c40cb265ad1292b1ae99) by @jbeezley seems to have expanded coverage to include the `server` and `web_client` directories but it still does not cover the core part of the toolkit located in the `histomicstk` directory.

jbeezley:
Girder creates a [coveragerc](https://github.com/girder/girder/blob/master/CMakeLists.txt#L28) file that it puts in the the build directory.  You could add this file to the list of omitted paths in the `[report]` section.  I'm not sure if there is a better way to do it other than running a `sed` replace after `cmake`.

jbeezley:
I guess we do have https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/plugin_tests/pycoverage.cfg in this repo.  I don't see anything obviously wrong.  I would first make sure the coverage is reported correctly when running it locally.

cdeepakroy:
@jbeezley Just spoke to zach. The `--sources` option is being set [here](https://github.com/girder/girder/blob/master/tests/PythonTests.cmake#L116-L117) in  `girder/tests/PythonTests.cmake`.

And the `${other_covg}` variable is being set [here](https://github.com/girder/girder/blob/master/tests/PythonTests.cmake#L99) to `",${PROJECT_SOURCE_DIR}/plugins/${fn_PLUGIN}/server"`.

We could either change it to `",${PROJECT_SOURCE_DIR}/plugins/${fn_PLUGIN}/` or create an option to `add_python_test` to set the `${other_covg}`.

Alternatively, maybe in [plugin_tests/pycoverage.cfg](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/plugin_tests/pycoverage.cfg) we set sources to include other directories apart from `server`. I am not sure if it accepts relative paths and if so relative to what.

jbeezley:
I tried adding a sources line to the pycoverage file, but it looks like the commandline flag overrides it.  I'm not sure there is any other option than to add an argument to the cmake function to configure the `--sources` option.

cdeepakroy:
Can we allow setting the ${other_covg} variable globally once instead of specifying it in each `add_python_test` command?

jbeezley:
Yes, I think we could make it a top level configuration option in girder.

zachmullen:
> Can we allow setting the ${other_covg} variable globally once instead of specifying it in each add_python_test command?

Alternatively, you could write a small function of your own that wraps `add_python_test` and passes in the appropriate option (once we add such an option to `add_python_test` in Girder).

cdeepakroy:
Writing a wrapper function would work too! @jbeezley Would you be able to take this up and make a PR on girder?

neal-siekierski:
@jbeezley and others: Thanks!

Since I suspect these changes will still cause a reported decrease in coverage because of the new `utils.py`, the question becomes what do about that. Perhaps it could be excluded, or I could get it tested, which might be done by either testing the CLIs or appending to `sys.path` again with a test in `plugin_tests` -- if it would get picked up the coverage tool, since it's not in the `histomicstk` directory. Thoughts?

jbeezley:
I think it's more important that the coverage report is accurate than to worry about coverage going down.  Ideally, the cli's should be tested--the low coverage is just a reminder that they are not.

neal-siekierski:
Changes since last time:
- Added a test in `cli_results_test` for `ColorDeconvolution`
- Create a new test, `cli_common_test`, since the subprocess invocation of `cli_results_test` doesn't run coverage.
  - This fixes the coverage issue.
- The front-end doesn't handle vector parameters with no default well; work around this by using -1 as a default.

girder/slicer_cli_web#38 has changes necessary for the REST endpoints of the changed CLIs to be usable (without it the endpoints generate but can't be used).

PR #294, opened 2017-04-14
Ensure locales are installed.

manthey:
Using today's Ubuntu 16.04 docker base, you have to install the locales package explicitly.  We need this to ensure that filenames are handles in UTF-8, since Girder uses the internal file names when sending files to girder worker.

A few days ago, locales was installed as a side-effect of another package.


manthey:
@kotfic, would you care to review this?

PR #296, opened 2017-04-17
Add a panel for drawing annotations

jbeezley:
This also includes a number of improvements to the existing panels and widgets.

1. Modal dialogs can now be dismissed with escape and submitted with return.
2. Added the ability to delete annotations.
3. Added tooltips to panel actions.
4. Added client side integration tests.
5. Changing the visible bounds now won't push to the browser history state.

This requires https://github.com/girder/large_image/pull/177 to be merged before the tests will pass.

test


web_client/dialogs/editAnnotation.js
```diff
@@ -0,0 +1,95 @@
...
+        var label = this.$('#h-element-label').val();
+        if (label) {
+            data.label = {
+                value: label
+            };
+        }
+
+        var lineWidth = this.$('#h-element-line-width').val();
+        if (lineWidth) {
+            data.lineWidth = parseFloat(lineWidth);
```
manthey: In general, do we need checks to make sure senseless values are rejected?  For instance, I can enter negative numbers.

manthey:
When we save the annotations, the new layer defaults to hidden.  It would be nice if it defaulted to shown.

I supposed I'd also like to see annotations automatically shown when they are generated as a result of an image processing job.

web_client/panels/ZoomWidget.js
```diff
@@ -88,7 +88,7 @@ var ZoomWidget = Panel.extend({
     setViewer(viewer) {
         var geo = window.geo;
         var range;
-        this.viewer = viewer;
+        this.viewer = viewer.viewer;
```
manthey: I find this mildly confusing -- in the ZoomWidget, `this.viewer` is the geojs map, whereas in the other widgets it is the viewer widget.  Perhaps for clarity it should always be the viewer widget and elsewhere in the ZoomWidget you could use `this.viewer.viewer`.  Alternately, document the result of `setViewer` in different widgets, since it differs.

web_client/panels/AnnotationSelector.js
```diff
@@ -33,21 +72,39 @@ var AnnotationSelector = Panel.extend({
...
+        var id = $(evt.currentTarget).parent('.h-annotation').data('id');
         var model = this.collection.get(id);
         model.set('displayed', !model.get('displayed'));
-        this.render();
+    },
+
+    /**
+     * Delete an annotation from the server.
+     */
+    deleteAnnotation(evt) {
```
manthey: When I delete a huge annotation (one with millions of elements), the delete takes a significant amount of time (more than a minute).  While that should be improved as a separate issue, there is no feedback that the delete has been triggered.  Perhaps the annotation entry have a class added to it that disables the associated buttons and greys it out (a reload or update might revert this look, but at least there would be some feedback).

jbeezley:
I made a simple change to show annotations after they are saved via the draw panel.  To automatically display annotations after a job is more difficult because it's hard to say if a new annotation on the item was added as a result of the job or not.  I think the best I can do is make the logic such that all annotations added after the initial render are shown by default.

manthey:
I see the problem.  This means that if another user manually adds an annotation, it will automatically appear, too.  I think this will be uncommon enough that having all new annotations appear would be preferred.

manthey:
I think some of the CSS is off for the save button (or is the intentional)?
image

manthey:
If I click one of the annotation buttons (point/rectangle/ploygon) exactly on the icon of the button, then the next click on the map doesn't generate the appropriate click event.  I think the icons need a pointer-events: none (this worked when I added such a property to `.h-draw span`, but that may not be the best css selector).

PR #298, opened 2017-04-18
Remove RTD files and consolidate all requirements into one file

cdeepakroy:
This PR includes the following changes
* Removes any  files specific to ReadTheDocs as we are not using it to host documentation anymore.
* Consolidates all the requirements (currently spread across multiple files) into one file and change setup.py to accomodate this.


docs/installation.rst
```diff
@@ -31,8 +31,7 @@ installed as follows::
     $ git clone https://github.com/DigitalSlideArchive/HistomicsTK.git
     $ cd HistomicsTK
     # pip install git+https://github.com/cdeepakroy/ctk-cli
-    $ pip install --no-cache-dir -r requirements_c_conda.txt
-    $ pip install --no-cache-dir -r requirements.txt -r requirements_c.txt
+    $ pip install --no-cache-dir -r requirements.txt
```
jbeezley: Any reason not to put the `ctk-cli` package in the requirements file as well?

cdeepakroy: That is part of my plan! 

I tried adding it, but parse_requirements was throwing an error. So i am first making sure TravisCI succeeds without adding it to the requirements.txt file to isolate the issue. Then i will add it back and check.

cdeepakroy: [This](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/setup.py#L31) where we use `pkg_resources.parse_requirements` to read in all dependencies from the requirements.txt file fails if we have editable requirements like `-e git+https://github.com/cdeepakroy/ctk-cli#egg=ctk_cli`.

Look at this discussion on a [pypa/pip github issue](https://github.com/pypa/pip/issues/2124) and this discussion on [stackoverflow](http://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py) that complicates doing this.

cdeepakroy: I will refrain from adding it to requirements.txt in this PR as i don't have a solution for this issue at the moment.

cdeepakroy: `ctk_cli` is anyway not needed when HistomicsTK is used as a pure python toolkit for analysis.

cdeepakroy:
@manthey Made a commit to fix references to requirements files in ansible roles. Can you please check if i got everything?

@kotfic This PR consolidates all the dependencies into a single `requirements.txt` file.

kotfic:
I believe you have all the ansible references sorted, once this goes in @danlamanna will need to merge/rebase his refactor branch,  that will propagate down to the vagrant fixes and we should be all set.

requirements.txt
```diff
@@ -1,9 +1,16 @@
-###############################################################
 #
-# Add all python packages that dont depend on C libraries here
+# scientific packages
+#
+Cython>=0.24
+nimfa==1.2.3
```
manthey: Can we change the nimfa requirement to `>=`?

requirements.txt
```diff
@@ -1,9 +1,16 @@
...
+scikit-image>=0.12.3
+scikit-learn>=0.17.1
+scipy>=0.16.0
+
+#
+# dask packages
 #
-###############################################################
-lxml>=3.4.4
-docker-py==1.9.0
```
manthey: We need docker-py for the `deploy-docker` script.  Perhaps it should be moved to requirements-dev.txt.

cdeepakroy: Added it!

cdeepakroy: Should i also change everything in requirements_dev.txt to  `>=`

manthey: docker-py needs to be >=1.9,<2

manthey: Otherwise, we probably just need to check to make sure they work with the current latest.

PR #299, opened 2017-04-19
Refactor Ansible to use official Girder role

danlamanna:
This is currently only implemented on the vagrant install, the docker yml needs to be altered to use the role.

The main benefit here is to lean on the upstream Girder role for the library (we were vendoring this directly) and installation of Girder.

This also adds some checksums and `creates` directives to avoid re-downloading and unpacking certain 3rd party libraries.


ansible/roles/openslide/tasks/main.yml
```diff
@@ -88,14 +90,19 @@
 
 # libtiff
 - name: Download libtiff
-  get_url:
-    url: http://download.osgeo.org/libtiff/tiff-4.0.6.tar.gz
-    dest: "{{ root_dir }}"
+
+  command: >-
+    wget --retry-connrefused --waitretry=1 --read-timeout=20 http://download.osgeo.org/libtiff/tiff-4.0.6.tar.gz
```
manthey: osgeo sometimes takes 100 seconds to respond.   Can we increase this to something even larger (perhaps 300 seconds)?

manthey: I'd also recommend explicitly specifying the output file name (`-O tiff-4.0.6.tar.gz`), since if this is run as a local provisioning step and the download is partial the first time, the second time a file called `tiff-4.0.6.tar.gz.1` will be created, and the next step will never pass.

manthey: When installing to docker containers, we don't have wget installed in Ubuntu 16.04 by default (we do have curl).  We either need to apt-get wget in the common task (or this task) or use curl (or go back to using ansible's get_url with a timeout).

manthey:
Aside from the wget/osgeo issue, this works for me for both vagrant and docker.  However, prior to this refactor, all of the girder web clients were built as part of the girder (histomicstk) role, which meant that the provisioning step didn't need to do this.  Now, the "Build all frontend assets" happens during the provisioning, and this substantially slows down docker deployment.

ansible/roles/girder-histomicstk/tasks/main.yml
```diff
@@ -1,61 +1,43 @@
...
   command: "apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 68576280"
-  become: true
+  become: yes
+  become_user: root
 
 - name: NodeJS | Add PPA
   apt_repository:
-    repo: "deb https://deb.nodesource.com/node_6.x {{ ansible_distribution_release }} main"
-  become: true
+    repo: "deb https://deb.nodesource.com/node_4.x {{ ansible_distribution_release }} main"
```
manthey: Is there a reason to use node 4 rather than node 6?  In the grider readthedocs we show using node 6 (http://girder.readthedocs.io/en/latest/prerequisites.html)

ansible/roles/girder-histomicstk/tasks/main.yml
```diff
@@ -1,61 +1,43 @@
...
+  become: yes
```
manthey: Is there a reason to use `become: yes` over `become: true`?  Should we be consistent?

Similarly, the default value of `become_user` is `root`, so is there a reason to specify it explicitly?

ansible/roles/girder-histomicstk/tasks/main.yml
```diff
@@ -1,61 +1,43 @@
...
+  become_user: root
 
 - name: NodeJS | Install package
   apt:
     name: nodejs
-    install_recommends: no
```
manthey: I suggest we should always use `install_recommends: no` when installing apt packages.  It (a) makes smaller docker images, and (b) makes our actual requirements much more explicit.  In the past, some of the suggested packages have changed resulting in not installing something we've actually needed.

manthey:
The `deploy_local.sh` script is now broken -- it checks for Ubuntu 14.04.  That should probably be changed to 16.04.  It needs to fetch the girder role via ansible galaxy.

kotfic:
@manthey I believe the last 4 commits resolves all issues you've mentioned. Due to upstream changes in the girder role i've moved the worker role before girder.girder.  This has allowed me to move the ```girder-install web``` call back into girder-histomics and prevent the long deployment delay for the docker containers.

ansible/roles/openslide/tasks/main.yml
```diff
@@ -57,10 +57,11 @@
 
 # openjpeg
 - name: Download openjpeg
-  get_url:
-    url: https://github.com/uclouvain/openjpeg/archive/v2.1.2.tar.gz
-    dest: "{{ root_dir }}/openjpeg-2.1.2.tar.gz"
-    checksum: "sha1:c8671e7f577fdc58abde1e1f32b10d372e6f9b07"
+  command: >-
+    wget --retry-connrefused --waitretry=1 --read-timeout=300 https://github.com/uclouvain/openjpeg/archive/v2.1.2.tar.gz -O openjpeg-v2.1.2.tar.gz
```
manthey: The output name has an extra v in it.

ansible/roles/openslide/tasks/main.yml
```diff
@@ -126,10 +126,11 @@
 
 # Openslide
 - name: Download openslide
-  get_url:
-    url: https://github.com/openslide/openslide/archive/v3.4.1.tar.gz
-    dest: "{{ root_dir }}/openslide-3.4.1.tar.gz"
-    checksum: "sha1:bd891d20434c9d99d71c3125f990f42755b0ebec"
+  command: >-
+    wget --retry-connrefused --waitretry=1 --read-timeout=300 https://github.com/openslide/openslide/archive/v3.4.1.tar.gz -O openslide-v3.4.1.tar.gz
```
manthey: Same here.

manthey:
We need `mq_private_ip: localhost` in the `deploy_local.yml` file.

PR #301, opened 2017-04-19
Use Dask in sample_pixels (and a fix)

neal-siekierski:
The main effect of this PR is to increase the use of Dask to the `sample_pixels` function in `histomicstk.utils` and all the CLIs that directly or indirectly make use of it, namely:
- `BackgroundIntensity`
- `SeparateStainsMacenkoPCA`

These CLIs now, like `NucleiDetection`, take an optional `--scheduler_address` parameter for an existing Dask distributed cluster. If not passed, a `LocalCluster` is created and used automatically.

Other changes are:
- `NucleiDetection` uses the default `None` when no other default is specified rather than `"null"`, simplifying argument handling
- The documentation for the magnification parameters of `BackgroundIntensity` and `SeparateStainsMacenkoPCA` has been updated to reflect the changes made to the defaults of `background_intensity` late in the life of #280. (This is the titular fix)

### DIFF
--- a/histomicstk/utils/sample_pixels.py
+++ b/histomicstk/utils/sample_pixels.py
@@ -8,7 +8,7 @@ from .simple_mask import simple_mask
 
 def sample_pixels(slide_path, sample_percent=None, magnification=None,
                   tissue_seg_mag=1.25, min_coverage=0.1, background=False,
-                  sample_approximate_total=None):
+                  sample_approximate_total=None, tile_grouping=256):
     """Generates a sampling of pixels from a whole-slide image.
 
     Useful for generating statistics or Reinhard color-normalization or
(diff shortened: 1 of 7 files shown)


histomicstk/utils/sample_pixels.py
```diff
@@ -75,15 +80,36 @@ def sample_pixels(slide_path, sample_percent=None, magnification=None,
     # generate sample pixels
     sample_pixels = []
 
-    scale_hres = {'magnification': magnification}
+    iter_args = dict(scale=dict(magnification=magnification),
+                     format=large_image.tilesource.TILE_FORMAT_NUMPY)
+
+    im_fgnd_mask_lres = dask.delayed(im_fgnd_mask_lres)
+
+    grouping = 256
```
cdeepakroy: May want to expose this as a parameter to the function

cdeepakroy: Rest LGTM!

PR #303, opened 2017-04-20
Add a test for cli results.

manthey:
Each CLI can be run with different command line parameters.  stdout can be checked to ensure it contains certain phrases but not other phrases.  Output files can be checked against sha256 hashes.

This also checks that all CLIs output help and xml, and that the xml can be used to construct REST endpoints using the slicer_cli_web functions.


manthey:
We can add tests for each CLI with some simple processing to ensure that we don't break core functionality.

Currently, I expect the build to fail due to some default-value issues in the SeparateStainsMacenkoPCA xml.

manthey:
@cdeepakroy This is ready for review.

plugin_tests/cli_results_test.py
```diff
@@ -0,0 +1,167 @@
...
+    base.startServer()
+
+
+def tearDownModule():
+    base.stopServer()
+
+
+class CliResultsTest(unittest.TestCase):
+    def _runTest(self, cli_args=(), cli_kwargs={}, outputs={}, contains=[],
+                 excludes=[]):
```
cdeepakroy: Really like this!

plugin_tests/cli_results_test.py
```diff
@@ -0,0 +1,167 @@
...
+            os.path.join(TEST_DATA_DIR, 'Easy1.png'),
+            'tmp_1.anot',
+        )
+        cli_kwargs = {
+            'analysis_roi': '-1.0, -1.0, -1.0, -1.0',
+        }
+        self._runTest(cli_args, cli_kwargs, outputs={
+            'tmp_1.anot': {
+                'contains': ['elements'],
+                # 'hash': '02b240586412c87ad5cbf349b7c22f80f1df31eef54ed8ee4ad1fd3624a89fa2',
```
cdeepakroy: Requiring the sha hash to match could be too rigid for comparing image analysis results.

But i guess we can extend this later.

PR #310, opened 2017-04-24
Update vagrant to 16.04 as a part of ansible refactor

kotfic:


manthey:
Trying to do `vagrant up` on a Windows machine, I get the error:
```
- downloading role from https://github.com/girder/ansible-role-girder/archive/master.tar.gz
- extracting girder.girder to /vagrant/ansible/roles/girder.girder
 [WARNING]: - girder.girder was NOT installed successfully: Could not update files in /vagrant/ansible/roles/girder.girder: [Errno 71] Protocol error
```

danlamanna:
This correctly provisioned and works for me.

ansible/roles/provision/tasks/main.yml
```diff
@@ -76,10 +76,8 @@
     plugins: "{{ girder_plugins }}"
     state: present
 
-# Girder web was already installed with all pluigins, so this shouldn't be
-# necessary.
-# - name: Install girder web if necessary
-#   command: girder-install web
+- name: Rebuild frontend assets
+  command: girder-install web
```
danlamanna: Is this necessary?

ansible/roles/worker/templates/girder_worker.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Girder Worker Service
+After=network.target
+
+[Service]
+Type=simple
+User={{ worker_exec_user }}
+Group={{ worker_exec_group }}
+WorkingDirectory=/opt/girder_worker
+ExecStart=/bin/sh -c "LANG=en_US.UTF-8 python -m girder_worker --concurrency={{worker_concurrency}} --pidfile=.worker_pid --logfile=/opt/girder_worker/worker.log -l info"
```
danlamanna: Might be a better question for @manthey, but why does `LANG` have to be set here? Is this not using the system locale for some reason?

manthey: We have to have both Girder and Girder worker in a UTF8 locale.  We really don't care which one (I picked US English because of my own cultural bias).  The system locale defaults to `C`, which is *not* UTF8.  Without this, files that have non-ASCII characters (or maybe its non-LATIN1 characters), cannot be sent from Girder to girder_worker or back.

Vagrantfile
```diff
@@ -1,6 +1,7 @@
 Vagrant.configure("2") do |config|
   config.vm.box_url = "https://cloud-images.ubuntu.com/vagrant/trusty/current/trusty-server-cloudimg-amd64-vagrant-disk1.box"
```
manthey: I think we can get rid of this `box_url` line that refers to trusty.

ansible/roles/worker/templates/girder_worker.service
danlamanna: Is this something you think should be the default in the upstream Girder Worker role?

manthey: I think it probably should be in both Girder and Girder Worker roles, at least when they are run in docker containers.

manthey:
Using vagrant up (or vagrant provision) on Windows results in an `[Errno 71] Protocol error`.  This can be bypassed by running vagrant as `admin`.  Although that works, it is the equivalent of requiring root access on linux -- a bad idea if we can avoid it.  This happens when it creates the ansible/roles/girder.girder role, apparently because of how ansible creates symbolic links through a vagrant-mounted directory.  Perhaps if the girder.girder role didn't have a symbolic link from roles/girder.girder/library/test/library/girder.py to ../../girder.py, then the admin access wouldn't be required.

danlamanna:
@manthey Would you mind trying https://github.com/girder/girder/pull/1947 to see if that resolves the issue? I think Ansible redownloads the dependencies from Galaxy each time, so you would have to manually put the subdirectory from that PR in your roles path, and remove the girder requirements from `requirements.yml`.

manthey:
I no longer get the issue when running as an ordinary user.

ansible/roles/provision/tasks/main.yml
kotfic: Yes

manthey:
`vagrant up` now works as expected for me (confirming that the upstream change was effective without manual intervention).

PR #314, opened 2017-04-25
Improve CLI XML

neal-siekierski:
This improves the XML specifications for `BackgroundIntensity` and `SeparateStainsMacenkoPCA`, namely by:
- Specifying the input images with `<image>` rather than `<file>`
- Removing the `minimum` constraint on `--samplePath`, allowing the special default value of `-1` to be valid.

### DIFF
--- a/server/SeparateStainsMacenkoPCA/SeparateStainsMacenkoPCA.xml
+++ b/server/SeparateStainsMacenkoPCA/SeparateStainsMacenkoPCA.xml
@@ -11,13 +11,13 @@
   <parameters>
     <label>IO</label>
     <description>Input/output parameters.</description>
-    <file>
+    <image>
       <name>sample_slide_path</name>
       <label>Input Slide</label>
       <channel>input</channel>
       <index>0</index>
       <description>Path to input slide image to be deconvolved</description>
-    </file>
+    </image>
     <double-vector>
       <name>macenko_I_0</name>
       <label>Background Intensity</label>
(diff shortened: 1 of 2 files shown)


server/SeparateStainsMacenkoPCA/SeparateStainsMacenkoPCA.xml
```diff
@@ -30,7 +30,6 @@
       <longflag>samplePercent</longflag>
       <description>Percentage of pixels to sample.  Specify either this or --sampleApproximateTotal</description>
       <constraints>
-	<minimum>0</minimum>
 	<maximum>1</maximum>
```
manthey: Looking at this, if this is a percent, shouldn't the maximum be 100, not 1?  Otherwise, should it be called something other than percent?

neal-siekierski: It is used as a fraction in `histomicstk.utils.sample_pixels`, despite the name (which is `sample_percent` there as well). I'm happy to change all related names to the appropriate variety of `sample_fraction`.

manthey: Yes, I think it would be better to change the `_percent` to `_fraction` where appropriate.

cdeepakroy: I agree change it to `sample_fraction`

cdeepakroy: Rest LGTM!

neal-siekierski: I've changed this along with other things as part of #317.

PR #315, opened 2017-04-25
Fix the Dockerfile build

manthey:
Setup tools doesn't upgrade cleanly unless --ignore-installed is added.

### DIFF
--- a/web_client/stylesheets/body.styl
+++ b/web_client/stylesheets/body.styl
@@ -1,4 +1,2 @@
-body
-    user-select none
 #h-body-container
     height 100%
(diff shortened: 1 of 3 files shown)


web_client/stylesheets/body.styl
```diff
@@ -1,2 +1,4 @@
+body
+    user-select none
 #h-body-container
     height 100%
```
cdeepakroy: Should this be in this PR?

manthey: No.  .. reverting that.

PR #319, opened 2017-04-26
Guard against setting properties on tests that don't exist

### LINKED ISSUE #316: Local failure building tests -- test "web_client_HistomicsTK_annotations" does not exist
@jbeezley: @cdeepakroy and I get the error below when trying to build the tests on master. Clearly Travis has no problem with it, though. I get this error with a new build directory for Girder, and the particular revisions 6a7d593 and girder/girder@7b4022b. With `git bisect` I was able to locate bab64d8 as the first commit that causes the test build to fail.

```
-- Including plugin.cmake from "HistomicsTK"
CMake Error at plugins/HistomicsTK/plugin.cmake:100 (set_property):
  set_property given TEST names that do not exist:

    web_client_HistomicsTK_annotations

Call Stack (most recent call first):
  tests/CMakeLists.txt:111 (include)
  tests/CMakeLists.txt:123 (_add_plugin)


-- Configuring incomplete, errors occurred!
```

jbeezley:
Fixes #316

### DIFF
--- a/plugin.cmake
+++ b/plugin.cmake
@@ -14,6 +14,11 @@
 #  limitations under the License.
 ###############################################################################
 
+# support the "if(TEST ...)" command
+if(POLICY CMP0064)
+  cmake_policy(SET CMP0064 NEW)
+endif()
+
 function(add_histomicstk_python_test case)
   add_python_test("${case}" PLUGIN HistomicsTK
     COVERAGE_PATHS "${PROJECT_SOURCE_DIR}/plugins/HistomicsTK/histomicstk"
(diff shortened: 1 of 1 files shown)


cdeepakroy:
I confirm that this resolves the error!

plugin.cmake
```diff
@@ -14,6 +14,11 @@
 #  limitations under the License.
 ###############################################################################
 
+# support the "if(TEST ...)" command
+if(POLICY CMP0064)
```
neal-siekierski: Will this work properly with older versions of CMake that don't have this policy?

jbeezley: For cmake < 3.4, the `if(TEST)` statement will be ignored.  This means the #316 will still occur, but using the defaults will still work.

neal-siekierski: All right. Would it be reasonable to make the error more transparent in the event that someone uses an old version of CMake and disables the tests, or are we not too concerned about that?

jbeezley: I'm not personally concerned about it, but I could address it.  One issue is that I can't think of a blanket way to improve the error message.  For this specific case, we could check a combination of cmake version + js testing option, but there are other possible reasons why a test would not be created by `add_web_client_test` or `add_python_test`.

Perhaps the best option is to add a new argument to our cmake functions to add custom environment variables.

neal-siekierski: If adding custom environment variables is something we'll want to keep doing, that seems like a good approach. Since these changes make for a quick fix for those of us on cmake>=3.4, if that's liable to be more complicated it might be worth doing as a separate PR.

PR #320, opened 2017-04-26
Show tooltips when hovering the mouse over annotations

jbeezley:
Depends on https://github.com/girder/large_image/pull/183.

For now, the tooltip contains the annotation name, description, creator, and creation date.  It will be easy to add any additional information returned by the `GET /annotation/:id` endpoint.  Additionally, there is a hook for responding to click events on the annotations.  We could use this to (for example) edit properties of annotations after they are created.

popover


web_client/views/popover/AnnotationPopover.js
```diff
@@ -0,0 +1,196 @@
...
+                this._users[id] = user;
+                this.render();
+            });
+        }
+    },
+
+    /**
+     * Format a Date object as a localized string.
+     */
+    _formatDate(s) {
```
manthey: Is there a reason `_formatDate` is a view method whereas `point`, `length`, and `rotation` are module functions?

manthey:
We added some pug linting, which now reports some style errors on this branch.

web_client/views/popover/AnnotationPopover.js
jbeezley: Nope.  Fixed.

manthey:
If I have an image opened and show the tooltip, then open a different image, sometimes the old tooltip information stays and will show on the new image.

PR #324, opened 2017-04-28
Add a link to download annotations

### LINKED ISSUE #323: Add download annotation button

jbeezley:
This functional, however, there is a problem when attempting to download a large annotation.  It takes the server several seconds to respond while the user has no indication that anything is happening.  The only way I can think of to deal with this is:
1. Add a `click` handler to the link which does the download and adds a spinner.  This will not work if the user tries to download via "Save As" in the context menu.
2. Add an endpoint `GET annotation/:id/download` that initiates a multipart download via paged responses.
If 2. is possible, I think it would be best because it doesn't override the browser's default behavior.

Closes #323

### DIFF
--- a/web_client/panels/AnnotationSelector.js
+++ b/web_client/panels/AnnotationSelector.js
@@ -13,7 +13,8 @@ import '../stylesheets/panels/annotationSelector.styl';
 var AnnotationSelector = Panel.extend({
     events: _.extend(Panel.prototype.events, {
         'click .h-toggle-annotation': 'toggleAnnotation',
-        'click .h-delete-annotation': 'deleteAnnotation'
+        'click .h-delete-annotation': 'deleteAnnotation',
+        'click .h-download-annotation': 'downloadAnnotation'
     }),
 
     /**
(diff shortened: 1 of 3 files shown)


web_client/panels/AnnotationSelector.js
```diff
@@ -13,7 +13,8 @@ import '../stylesheets/panels/annotationSelector.styl';
 var AnnotationSelector = Panel.extend({
     events: _.extend(Panel.prototype.events, {
         'click .h-toggle-annotation': 'toggleAnnotation',
-        'click .h-delete-annotation': 'deleteAnnotation'
+        'click .h-delete-annotation': 'deleteAnnotation',
+        'click .h-download-annotation': 'downloadAnnotation'
```
manthey: I'm failing to understand how this works -- where is `downloadAnnotation` defined?

manthey:
I think we'll want a custom download endpoint.

Should we just be returning the `annotation` entry from this, as that is what gets uploaded?  The other top-level fields include potentially useful metadata (`creatorId`, `created`), plus some we may not want to return (`_elemenQuery`, `_version`) .

web_client/panels/AnnotationSelector.js
jbeezley: That's leftover code.  I'll remove it.

jbeezley:
I'm not sure.  My first instinct is to say just the annotation object, but I don't think it hurts to include the toplevel metadata as well.

manthey:
Do either @cooperlab and @dgutman have an opinion on what should be downloaded when you download an annotation?  If including the metadata is desired, then this works now.

cooperlab:
If we are restricting to human annotations then I would prefer including metadata (and any other info). Is small and can be easily discarded if not wanted.

Issue #284, opened 2017-04-11
Implement a CLI for Positive Pixel Counting


cdeepakroy:
The Positive Pixel Count algorithm is used to quantify the amount of a specific stain present in a slide. The algorithm parameters include a color (range of hues and saturation) of interest and three intensity ranges (weak, positive, and strong). For pixels which satisfy the color specification, the algorithm counts the number and intensity‐sum in each intensity range, along with three additional quantities: average intensity, ratio of strong/total number, and average intensity of weak+positive pixels. 

[Here](http://tmalab.jhmi.edu/aperiou/userguides/Positive_Pixel.pdf) is a reference describing this method.

cdeepakroy:
Fixed in PR #325

PR #325, opened 2017-04-28
WIP: Positive pixel counting

neal-siekierski:
This introduces a CLI `PositivePixelCount` to do positive pixel counting. It calculates the summary statistics specified in #284, as well as an optional image output for small region sizes.

Notes:
- There is no progress reporting at the moment
- No parallelization is done when creating an output image
- I've picked the colors for the output image from the coolwarm color map (cf. [\[1\]][1])

[1]: https://www.ncl.ucar.edu/Document/Graphics/ColorTables/MPL_coolwarm.shtml


server/PositivePixelCount/PositivePixelCount.py
```diff
@@ -0,0 +1,103 @@
...
+
+def positive_pixel_count_single_tile(args, tile, makeLabelImage):
+    tile = tile[..., :3]
+    total = np.empty(6)
+    tile_hsi = rgb_to_hsi(tile / 255.)
+    mask_all_positive = (
+        (np.abs((tile_hsi[..., 0] - args.hueValue + 0.5 % 1) - 0.5) <= args.hueWidth / 2.)
+        & (tile_hsi[..., 1] >= args.saturationMinimum)
+        & (tile_hsi[..., 2] < args.intensityUpperLimit)
+        & (tile_hsi[..., 2] >= args.intensityLowerLimit)
```
neal-siekierski: Also, this (with the `&`s leading the lines instead of ending them) runs afoul of Girder's flake8 configuration.

server/PositivePixelCount/PositivePixelCount.py
```diff
@@ -0,0 +1,103 @@
...
+        labelImage = np.full_like(tile, 255)
+        # Colors from the "coolwarm" color map
+        labelImage[mask_all_positive] = (
+            mask_weak[..., np.newaxis] * [60, 78, 194]
+            + mask_pos[..., np.newaxis] * [221, 220, 220]
+            + mask_strong[..., np.newaxis] * [180, 4, 38]
+        )
+        return total, labelImage
+    else:
+        return total
```
cdeepakroy: I would suggest using a dict for the results (whatever you are putting into the `total` array here). 

With a flat array it is hard to keep track of what is being stored where and might result in misplacement errors when someone else modifies this later.

cdeepakroy: Rest LGTM!

neal-siekierski: Done.

neal-siekierski:
Is progress reporting in a state where we'd like to integrate it into `PositivePixelCount` as part of the PR?

Also, there's still the matter of [`flake8` complaining](#discussion_r113964857). @manthey or @jbeezley, what would be the most appropriate way to handle this? Girder's configuration could be amended, or perhaps HistomicsTK could include its own configuration for `flake8`.

jbeezley:
It doesn't look like there is currently a way for a plugin to override the flake8 config.  Given the discussion in https://github.com/PyCQA/pycodestyle/issues/498, it seems reasonable to disable W503 in girder.

PR #326, opened 2017-04-28
When building docker files, only build the girder front-end once.

manthey:
It is built in the girder-histomicstk role, so there is no need to have the girder.girder role build it.


manthey:
@kotfic Could you review this.  It speeds up building the docker images.

PR #327, opened 2017-04-28
Support running ctest from vagrant machine

kotfic:
To test, from a fresh directory:

```
git clone git@github.com:DigitalSlideArchive/HistomicsTK.git
cd HistomicsTK
vagrant up
# Wait for provision to finish
vagrant ssh
cd /opt/histomicstk/build
ctest -VV -R "HistomicsTK"
```


manthey:
I think we want ctest to default to just testing HistomicsTK, so the cmake command should change to
```
cmake -DRUN_CORE_TESTS:BOOL="OFF" -DTEST_PLUGINS:STRING="HistomicsTK" {{root_dir }}/girder
```
and then the example ctest command becomes
```
ctest -VV
```

Ideally, we'd also add some documentation (but that can be done in another PR).

manthey:
One more request: we are only covering a small amount of the python code, so please add `-DCOVERAGE_MINIMUM_PASS=30` to the cmake command.

PR #331, opened 2017-05-05
Make ComputeNucleiFeatures CLI work on WSI images and refactor other things

cdeepakroy:
This PR:
* Makes ComputeNucleiFeatures CLI to work on WSI images using Dask for parallelization
* Generalizes and moves a few functions from NucleiDetection CLI to cli_common.utils so they can be shared by other CLIs
* Renames feature extraction functions to follow PEP8 conventions
* Bumps versions of all scientific python package dependencies to latest


server/cli_common/utils.py
```diff
@@ -30,7 +39,162 @@ def get_stain_matrix(args, count=3):
...
+
+    im_lres = im_lres[:, :, :3]
+
+    # compute foreground mask at low-res
+    im_fgnd_mask_lres = htk_utils.simple_mask(im_lres)
+
+    return im_fgnd_mask_lres, fgnd_seg_scale
+
+
+def detect_nuclei_kofahi(im_nuclei_stain, args):
```
neal-siekierski: Would `detect_nuclei_kofahi` be better in the core package? Except for taking `args` instead of individual parameters, it doesn't seem to do anything CLI-specific.

plugin.cmake
```diff
@@ -113,7 +113,12 @@ add_histomicstk_python_test(blob_detection_filters
 
 add_histomicstk_python_test(cli_common
   # Work around CMake bug when using the same image multiple times
-  # EXTERNAL_DATA "plugins/HistomicsTK/Easy1.png"
+  EXTERNAL_DATA
+  # "plugins/HistomicsTK/Easy1.png"
```
neal-siekierski: The `# Work around CMake bug` comment ought to be right before the commented-out item

server/ComputeNucleiFeatures/ComputeNucleiFeatures.py
```diff
@@ -1,113 +1,259 @@
...
+        # append result to list
+        tile_result_list.append(cur_result)
+
+    tile_result_list = dask.delayed(tile_result_list).compute()
+
+    nuclei_annot_list = [annot
+                         for annot_list, fdata in tile_result_list
+                         for annot in annot_list]
+
+    nuclei_fdata = pd.concat([fdata for annot_list, fdata in tile_result_list],
```
neal-siekierski: As far as I know there's indeed not a significantly nicer way to do this.

server/ComputeNucleiFeatures/README.md
```diff
@@ -1,4 +1,4 @@
-HistomicsTK ComputeNucleiFeatures Application
+HistomicsTK compute_nuclei_features Application
```
neal-siekierski: Find-and-replace gone awry, it seems.

server/cli_common/utils.py
cdeepakroy: Yes that is ultimate goal! 

Will do that once i write similar wrapper functions for other nuclei segmentation algorithms - min model, gvf tracking etc.

neal-siekierski: No worries, then.

PR #332, opened 2017-05-08
Deploy pinned versions unless asked to use latest.

manthey:
Also, deploy a CLI docker image by default.  The program that does this can also run tests on the CLI to ensure that the server, girder worker, and libraries are deployed properly.  Add --cli-test to the start or provision commands to run the CLI tests.  Add --no-cli to not add a CLI.  The CLI docker image version is either a pinned version or latest, using the same logic as the main docker images.

This also updates the installation docs.


ansible/README.rst
```diff
@@ -0,0 +1,131 @@
...
+    sudo pip install -U ansible
+    git clone https://github.com/DigitalSlideArchive/HistomicsTK
+
+Deploy
+######
+
+::
+
+    cd HistomicsTK/ansible
+    ./deploy_local.sh
```
cdeepakroy: This documentation looks really great!

I would suggest moving this stuff to [docs/installation.rst](https://github.com/DigitalSlideArchive/HistomicsTK/blob/master/docs/installation.rst) as that is where a lay user will go to first?

manthey: It is linked to docs/installation.rst, so it appears in both locations.

cdeepakroy: Never mind, I see that you include this rst file in installation.rst. Great!

manthey:
@cdeepakroy Thanks.

PR #335, opened 2017-05-11
Simplify the docker tests.

manthey:
The docker endpoints are used from the slicer_cli_web plugin, and are tested there.  These tests were duplicates, and didn't actually cover any local code.  The tests have been simplified to just check that the expected endpoints are present.

This also removes docker-py from the development requirements, as it was only used in the tests.


manthey:
A change was necessary now that slicer_cli_web is using the 2.x docker client.  Since Girder moved to the 2.x docker client, vagrant up would probably fail; this should fix it.

PR #337, opened 2017-05-19
Implement a CLI for Nuclei classification and fix coord swap bug of nuclear boundary annotations

cdeepakroy:
This PR implements a CLI for nuclei classification that takes as input

* an input image
* a pickled file of the scikit-learn classification model trained to classify nuclei
* a csv or HDF5 feature file containing nuclear features computed using the `histomicstk.features.compute_nuclei_features` function
* a nuclear annotation (*.anot) file containing the annotations of the nuclei to classify

and outputs an annotation file with a list of annotations grouped by their predicted class.

It also fixes the XY coordinate swap bug in `cli_common.utils.create_tile_nuclei_boundary_annotations`.

It also uses a smaller svs file file to `cli_common_test`.


server/NucleiClassification/NucleiClassification.py
```diff
@@ -0,0 +1,181 @@
...
+
+    fname, feature_file_format = os.path.splitext(args.inputFeatureFile)
+
+    if feature_file_format == '.csv':
+
+        ddf = dd.read_csv(args.inputFeatureFile)
+
+    elif feature_file_format == '.h5':
+
+        ddf = dd.read_hdf(args.inputModelFile, 'Features')
```
neal-siekierski: I suspect this should be `args.inputFeatureFile` again.

server/NucleiClassification/NucleiClassification.py
```diff
@@ -0,0 +1,181 @@
...
+import logging
+logging.basicConfig(level=logging.CRITICAL)
+
+sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__), '..')))
+from cli_common import utils as cli_utils # noqa
+
+
+def gen_distinct_rgb_colors(n, seed=None):
+    """
+    Generates N visually distinct RGB colors
```
neal-siekierski: For what it's worth, the hue of HSV isn't perceptually uniform, which means that generated colors will "clump" visually. Since doing it right is a pain, hopefully this isn't an issue.

server/NucleiClassification/NucleiClassification.py
```diff
@@ -0,0 +1,181 @@
...
+
+    h = np.random.random(n)
+    golden_ratio_conjugate = 0.618033988749895
+
+    for i in range(n):
+
+        h += golden_ratio_conjugate
+        h %= 1
+
+        color_list.append(colorsys.hsv_to_rgb(h, 1.0, 1.0))
```
neal-siekierski: This could be replaced with
```python
h = np.random.random()  # (Passing a parameter would return an array)
golden_ratio_conjugate = (np.sqrt(5) - 1) / 2  # (This is an analytic expression for the constant)
# (The hue parameter is automatically taken mod 1)
color_list = [colorsys.hsv_to_rgb(h + i * golden_ratio_conjugate, 1, 1) for i in range(n)]
```

server/NucleiClassification/NucleiClassification.py
```diff
@@ -0,0 +1,181 @@
...
+
+    class_color_map = dict(zip(clf_model.classes_,
+                               gen_distinct_rgb_colors(num_classes, seed=1)))
+
+    for i in range(len(nuclei_annot_list)):
+
+        cur_class = pred_class.iloc[i]
+
+        cur_anot = nuclei_annot_list[i]
+        cur_anot['line_color'] = class_color_map[cur_class]
```
neal-siekierski: According to https://github.com/girder/large_image/blob/master/docs/annotations.md, the attribute name is `"lineColor"` (camel-case).

According to the same document, colors must a string in one of several different formats. `rgb(123, 123, 123)` might be the easiest.

server/NucleiClassification/NucleiClassification.py
```diff
@@ -0,0 +1,181 @@
...
+
+        raise ValueError('The number of nuclei in the feature file and the '
+                         'annotation file do not match')
+
+    #
+    # Perform nuclei classification
+    #
+    def predict_nuclei_class_prob(df, clf_model):
+
+        return pd.Dataframe(data=clf_model.predict_proba(df.as_matrix()),
```
neal-siekierski: We're not using Dask for this at the moment, then.

neal-siekierski:
Re Travis: `web_client_HistomicsTK_annotations` is failing, and in one of the builds `test_segment_wsi_foreground_at_low_res` is failing due to the output arrays not matching by 0.455%.

server/NucleiClassification/NucleiClassification.py
neal-siekierski: Oh, I see. Never mind.

server/NucleiClassification/NucleiClassification.py
```diff
@@ -155,7 +153,8 @@ def predict_nuclei_class_prob(df, clf_model):
         cur_class = pred_class.iloc[i]
 
         cur_anot = nuclei_annot_list[i]
-        cur_anot['line_color'] = class_color_map[cur_class]
+        cur_anot['lineColor'] = 'rgb({})'.format(
+            ', '.join(str(int(round(255 * class_color_map[cur_class])))))
```
neal-siekierski: You'll need a generator expression or such, I believe, something like
```python
', '.join(str(int(round(255 * c))) for c in class_color_map[cur_class])
```

server/NucleiClassification/NucleiClassification.py
```diff
@@ -0,0 +1,179 @@
...
+        `numpy.random.seed` for more details
+
+    Returns
+    -------
+    colors_list : list
+        A list of n RGB colors
+
+    References
+    ----------
+    .. [#] http://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/  # noqa
```
neal-siekierski: I don't think `# noqa` is needed here. `python_static_analysis_HistomicsTK_plugins` doesn't fail locally for me if it's removed.

PR #340, opened 2017-05-24
Fix minor issues with ComputeNucleiFeatures and NucleiClassification CLIs

cdeepakroy:
* Make outputNucleiAnnotationFile a indexed output parameter
* Add NucleiClassification CLI to slicer_cli_list.json
* Remove pandas index column while writing features files
* Mentions file extensions in the descriptions of output files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,6 +109,9 @@ install:
     - cd $girder_path
     - python setup.py develop
     - cd $large_image_path
+    - pip install --no-cache-dir numpy>=1.10.2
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -109,6 +109,9 @@ install:
     - cd $girder_path
     - python setup.py develop
     - cd $large_image_path
+    - pip install --no-cache-dir numpy>=1.10.2
```
neal-siekierski: `requirements.txt` has `numpy>=1.12.1`. Would it be better for these to match?

PR #345, opened 2017-06-27
Update files based on new eslint rules.

manthey:
Also, add --test-local option to the cli_test script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,10 @@ addons:
       - libgif-dev
 
 before_install:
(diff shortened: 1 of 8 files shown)


jbeezley:
Everything looks fine to me, but I'm having trouble running the test changes.  We seem to require `docker<2`, but I don't see anything on pypi matching that.  Am I missing something?

manthey:
The older version of the python docker module is `docker-py`.  The docker-deploy file uses it; our other repos have been updated to use docker>=2, which is the `docker` module.

PR #348, opened 2017-07-05
Lists of annotations

manthey:
Support ingesting annotations that are lists of annotations.

Use the name of the parameter to detect annotations.   Prior to this, output annotations had to be named .anot by the user, which was brittle.  Now, output annotations must be in parameters that are internally named `.*AnnotationFile`, and the entered name no longer matters.

Update cmake file to use standard tests.   This lints pug which caught some errors.

Removed the example test.

Created a test for handling annotations.


neal-siekierski:
Re the annotation file reference linking, that will require modifications to the CLIs, presumably as part of this PR so as not to break the front-end, unless that's done in another PR first, which I can do.

Also, this approach is less brittle but still magic; in the long term, could the SEM XML spec be extended?

As a minor point, regarding the example test, I've used it previously as a template for new tests, which seems cleaner than copying and deleting from some other test. Would it be worth keeping for that use alone?

manthey:
The reference linking doesn't require changes to the CLIs -- it adds the reference on the Girder side of things and then gets the value passed back by girder worker.  It does require remove and readding the CLIs from the endpoints.

Let's make sure there is a less magic solution when we switch to item tasks.  If there isn't initially, we can make a PR there to ensure it is so.

I'll re-add the example test.

PR #349, opened 2017-07-05
Get coverage for CLIs

neal-siekierski:
This PR implements one possible way to get coverage information for the CLIs, namely by using the built-in `runpy` module &ndash; `_runTest` in `cli_results_test.py` is extended to support invoking a CLI via `runpy.run_path`.

It supporting the old subprocess behavior via a flag, which is necessary for supporting `testListCLI`, which uses a separate Python script and needs its stdout to be captured. Capturing, and therefore testing the contents of, stdout (and stderr) is not supported with `runpy`, but I don't expect this to be an important limitation as all important CLI outputs are stored in files.

This also fixes a typo in `_runTest`'s file-contains check, which was using the stdout-contains-check parameter instead.


plugin_tests/cli_results_test.py
```diff
@@ -64,28 +65,56 @@ def _runTest(self, cli_args=(), cli_kwargs={}, outputs={}, contains=[],
...
+            cmd = [arg if not arg.startswith('tmp_') else
+                   os.path.join(tmppath, arg) for arg in cli_args]
             cmd += ['--%s=%s' % (
                 k, v if not v.startswith('tmp_') else os.path.join(tmppath, v))
                 for k, v in six.iteritems(cli_kwargs)]
-            process = subprocess.Popen(
-                cmd, shell=False, stdout=subprocess.PIPE, cwd=cwd)
-            stdout, stderr = process.communicate()
-            self.assertEqual(process.returncode, 0)
+            if in_process:
```
manthey: You don't need two separate code paths here -- you can use `runpy.run_path` *and* capture stdout and stderr.  Much like with `old_sys_argv`:
```
                stdout = None
                old_sys_stdout = sys.stdout
                old_sys_stderr = sys.stderr
                sys.stdout = stdout = six.BytesIO()
                sys.stderr = stderr = six.BytesIO()
                try:
...
                finally:
                    sys.argv[:] = old_sys_argv
                    sys.stdout = old_sys_stdout
                    sys.stderr = old_sys_stderr
                stdout = stdout.getvalue()
                stderr = stderr.getvalue()
```

plugin_tests/cli_results_test.py
```diff
@@ -44,15 +45,15 @@ def tearDownModule():
...
-        """
-        Test a cli by calling subprocess.  Ensure that output files match a
-        sha256 and stdout contains certain phrases and excludes other phrases.
-        A value in the cli_args or cli_kwargs of 'tmp_<value>' will create a
-        temporary file and use that instead.  The same value in the output
-        list can be used to test the existence and contents of that file at the
-        end of the process.
+    def _runTest(self, cli_args=(), cli_kwargs={}, outputs={},
+                 in_process=True, contains=[], excludes=[]):
+        """Test a cli by calling runpy or subprocess.  Ensure that output
```
manthey: To match elsewhere, place `"""` on a line by itself.

plugin_tests/cli_results_test.py
neal-siekierski: My only concern is that, while this method does capture the output containing the desired strings, it doesn't work in the general case that something doesn't use the current value of `sys.stdout` or `sys.stderr` to determine where to write its output. Concretely, this doesn't capture the warning/error output from `large_image` when it tries to load a PNG as a TIFF.

plugin_tests/cli_results_test.py
neal-siekierski: I can do that to match within the file. More generally, we seem to be pretty split -- by my count, 68 of the other 112 doc strings in the entire project start on the same line, and the others don't.

manthey: I'm trying to encourage us to use Girder's convention, which is to have the doc string start on the next line.  Girder isn't wholly consistent either, but it has 1590 on the next line versus 65 on the same line...

neal-siekierski:
Is there anything else? Coverage is down, but that seems to only be because all CLIs are now being tested, and most of them aren't run beyond the `parse_args` call.

PR #353, opened 2017-07-10
Optionally merge in a girder configuration file.

manthey:
The girder.local.cfg file in the histomicstk container can be modified by the provisioning process to merge the current file with a file specified by a command line parameter.  For instance, `deploy_docker.py
provision --cfg=sample.cfg` will add the local file `sample.cfg` into the existing configuration.


manthey:
This is based on the `girder-worker-tmp-root` branch.  Those changes should do no harm, even if more changes are needed to support certain uses of Dask, so this PR and that branch can be merged sooner rather than later.

manthey:
@dgutman As a usage example, create a local file called `candygram.cfg` with the appropriate memcached settings:
```
[large_image]
cache_memcached_url: ["dockerhost"]
```
Then add `--cfg=candygram.cfg` to your `deploy_docker.py` parameters.  Make sure that the host machine's memcached configuration (`/etc/memcached`) listens to docker containers (`-l 0.0.0.0`) and has ample memory (`-m <size in Mb>`).

manthey:
`-l 127.0.0.1,172.18.0.1` appears to be sufficient, and much more secure than `0.0.0.0`, but I don't know if the second IP address will always be okay.

PR #355, opened 2017-07-14
Support newer versions of docker-py/docker.

### LINKED ISSUE #354: Update the docker module version used in deploy_docker
deploy_docker.py uses version <2 of the python `docker-py` module.  Internally, girder and girder_worker use version >=2 of the `docker` module (it was renamed).  When running on a local system, this is annoying.

manthey:
This now works with any version of the docker or docker-py module >= 1.9, including version 2.x.  The documentation has been changed to prefer the newer version.

This resolves #354 (and allows either version to be used).

### DIFF
--- a/ansible/deploy_docker.py
+++ b/ansible/deploy_docker.py
@@ -74,7 +74,7 @@ def containers_provision(**kwargs):
     """
     Provision or reprovision the containers.
     """
-    client = docker.from_env()
+    client = docker.from_env(version='auto')
     client = client if not hasattr(client, 'api') else client.api
     ctn = get_docker_image_and_container(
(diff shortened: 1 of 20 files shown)


ansible/deploy_docker.py
```diff
@@ -74,7 +74,8 @@ def containers_provision(**kwargs):
     """
     Provision or reprovision the containers.
     """
-    client = docker.from_env()
+    client = docker.from_env(version='auto')
+    client = client if not hasattr(client, 'api') else client.api
```
neal-siekierski: These two lines, since they occur six times, might be better as a function.

PR #356, opened 2017-07-17
Add a new dialog to open a recently annotated image

jbeezley:
This uses the new large_image endpoint that lists recently annotated images.  Currently, it will only show the most recent 10 images, we can add paging to it in a later PR if necessary.

test


manthey:
I have the same image multiple times in my system, which makes it hard to tell which image is which in this list.  If it isn't too much work, can we add a tooltip or sub-text for the image that shows its resource path?  That way I'd see that one is "/user/manthey/Public/sample_images/TCGA-02-0010-01Z-00-DX4.07de2e55-a8fe-40ee-9e98-bcb78050b9f7.svs" and the other is "/collection/HistomicsTK Tests/Images/TCGA-02-0010-01Z-00-DX4.07de2e55-a8fe-40ee-9e98-bcb78050b9f7.svs".

dgutman:
I've actualyl had the same issue-- I think this will be a useful thing to be able to access as we do file management/cleanup...

manthey:
@jbeezley There is another commit to the `fix-travis-build` branch that should finish fixing this branch's build.  We should probably merge that branch in first, since this branch has inherited some of those changes.

Issue #352, opened 2017-07-10
UI to download a ROI as a TIFF


manthey:
This would allow the user to select an ROI and then pick a magnification.  An estimate of the file size should be shown along with a download button.  If the output would require a BigTIFF (or, simplistically, has more than a gigapixel of data), don't allow a download.

The region endpoint can do the bulk of the work, but current is probably limited to png and jpeg.

manthey:
@Pierre-Assemat made a branch called `download-button-ROI` which has an initial implementation which downloads the current window at the current window's resolution.

Pierre-Assemat:
To select a ROI could we use the same tool than it used to draw annotation : like draw a rectangle then choose a magnification, show the size of the file ?

manthey:
Yes.  It might be nice to have both the current option and a second button that would let you select the area.  The current button could be relabeled "Download View", and then add a second button which would be "Download Area"..  After drawing the rectangle, we could show a dialog with a choice of magnifications, the output width, height, total number of pixels, and estimated file size, and a Download and Cancel button.

Pierre-Assemat:
I have few questions to clarify expectations :
- What is the best for the client: select magnification or zoom level ?
- What are the alterable parameters (Magnification/Zoom and Format are the only ones for now) ?
- How can I get corrects value of compression ratio for these different format (JPEG, PNG, and TIFF ) and the constant in the size file calculation ?
- Which is the default format ? 
- Do we need some verification test ?

manthey:
- Magnification is best.
- Magnification and format for now.
- We'll have to make a few files and see if what sort of compression ratio we get.  We'll try a few different typical files.
- @dgutman originally asked for TIFF.  He can say if he'd prefer the default to be PNG or JPEG instead.
- We need to add tests once we get the the feature working.  Tests can be complicated, so perhaps ask @jbeezley while he is still physically around to give you an introduction to them.

Pierre-Assemat:
I noticed different bug on this PR, without be detected by Travis CI tests, but I don't know the origin of this bugs : 

- I can't open an image 
![screenshot from 2017-10-05 12-20-43](https://user-images.githubusercontent.com/29961430/31238330-a5f85af2-a9c7-11e7-87eb-eb6ce52c9fce.png)
It's weird because the test : `web_client_HistomicsTK.annotations` try to open images and is passing on Travis but not on my local environment.

[...]

PR #357, opened 2017-07-17
WIP: Add download button in the ZoomWidget and select the ROI

Pierre-Assemat:
- [x] download view button
    - [x] Fix bug : Only right size of the image is download when zoom < 5
- [ ] ROI support
    - [x] download area button
    - [x] selection of an region of interest
    - [x] edit ROI dialog box
        - [x] edit magnification
        - [x] show Width and Height
        - [x] show Number of pixels 
        - [x] show Size of the file
        - [x] select Format (JPEG, PNG, TIF...)
    - [x] download edited ROI
    - [ ] debug 
        - [x] magnification impact the image output
        - [ ] ratio compression 
            - [ ] JPEG
            - [ ] PNG
            - [ ] TIFF
        - [x] crash of browser when submit with wrong value
        - [x] good value on Width, Height, and Size
    - [x] Re-organize the code following backbone structure
- [x] Give the same name for this two way of downloading the image
- [x] Add all the tests to assure the code coverage

Fixes : #352

### DIFF
--- a/plugin_tests/client/annotationSpec.js
+++ b/plugin_tests/client/annotationSpec.js
@@ -169,91 +169,89 @@ $(function () {
                 });
             });
         });
-    });
 
-    describe('Download view and region of interest', function () {
-        it('check href attribute of \'Download View\' link', function () {
-            runs(function () {
-                $('#download-view-link').bind('click', function (event) {
-                    event.preventDefault();
+        describe('Download view and region of interest', function () {
+            it('check href attribute of \'Download View\' link', function () {
+                runs(function () {
+                    $('#download-view-link').bind('click', function (event) {
+                        event.preventDefault();
+                    });
+                    $('.h-download-button-view').click();
                 });
-                $('.h-download-button-view').click();
+
+                waitsFor(function () {
+                    return $('#download-view-link').attr('href') !== undefined;
+                }, 'to be the url');
+
+                runs(function () {
+                    expect($('#download-view-link').attr('href')).toMatch(/\/item\/[0-9a-f]{24}\/tiles\/region\?width=[0-9-]+&height=[0-9-]+&left=[0-9-]+&top=[0-9-]+&right=[0-9-]+&bottom=[0-9-]+&contentDisposition=attachment/);
+                });
             });
 
-            waitsFor(function () {
-                return $('#download-view-link').attr('href') !== undefined;
-            }, 'to be the url');
+            it('open the download dialog', function () {
+                var interactor = geojsMap.interactor();
+                $('.h-download-button-area').click();
 
-            runs(function () {
-                expect($('#download-view-link').attr('href')).toMatch(/\/item\/[0-9a-f]{24}\/tiles\/region\?width=[0-9-]+&height=[0-9-]+&left=[0-9-]+&top=[0-9-]+&right=[0-9-]+&bottom=[0-9-]+&contentDisposition=attachment/);
+                interactor.simulateEvent('mousedown', {
+                    map: {x: 100, y: 100},
+                    button: 'left'
+                });
+                interactor.simulateEvent('mousemove', {
+                    map: {x: 200, y: 200},
+                    button: 'left'
+                });
+                interactor.simulateEvent('mouseup', {
+                    map: {x: 200, y: 200},
+                    button: 'left'
+                });
+
+                girderTest.waitForDialog();
+                runs(function () {
+                    expect($('.modal-title').text()).toBe('Edit Area');
+                });
+            });
+
+            it('test modifying form elements', function () {
+                const oldSettings = [];
+                const elements = [];
+                elements.push($('#h-element-width'), $('#h-element-height'),
+                    $('#h-nb-pixel'), $('#h-size-file'));
+                oldSettings.push($('#h-element-width').val(), $('#h-element-height').val(),
+                    $('#h-nb-pixel').val(), $('#h-size-file').val());
+                runs(function () {
+                    $('#h-element-mag').val(10).trigger('change');
+                    var i = 0;
+                    // Check all the setting labels change
+                    for (var value in oldSettings) {
+                        expect(elements[i].val()).not.toEqual(value);
+                        i++;
+                    }
+                });
+                runs(function () {
+                    $('#h-download-image-format').val('TIFF').trigger('change');
+                    // Check the size label change
+                    expect($('#h-size-file').val()).not.toEqual(oldSettings[3]);
+                });
+            });
+
+            it('ensure the download link is correct', function () {
+                waitsFor(function () {
+                    return $('#h-download-area-link').attr('href') !== undefined;
+                }, 'to be the url');
+
+                runs(function () {
+                    expect($('#h-download-area-link').attr('href')).toMatch(/\/item\/[0-9a-f]{24}\/tiles\/region\?regionWidth=[0-9-]+&regionHeight=[0-9-]+&left=[0-9-]+&top=[0-9-]+&right=[0-9-]+&bottom=[0-9-]+&encoding=[EFGIJNPT]{3,4}&contentDisposition=attachment&magnification=[0-9-]+/);
+                });
+            });
+
+            it('close the dialog', function () {
+                $('#g-dialog-container').girderModal('close');
+                waitsFor(function () {
+                    return $('body.modal-open').length === 0;
+                });
             });
         });
 
-        it('open the download dialog', function () {
-            var interactor = geojsMap.interactor();
-            $('.h-download-button-area').click();
-
-            interactor.simulateEvent('mousedown', {
-                map: {x: 100, y: 100},
-                button: 'left'
-            });
-            interactor.simulateEvent('mousemove', {
-                map: {x: 200, y: 200},
-                button: 'left'
-            });
-            interactor.simulateEvent('mouseup', {
-                map: {x: 200, y: 200},
-                button: 'left'
-            });
-
-            girderTest.waitForDialog();
-            runs(function () {
-                expect($('.modal-title').text()).toBe('Edit Area');
-            });
-        });
-
-        it('test modifying form elements', function () {
-            const oldSettings = [];
-            const elements = [];
-            elements.push($('#h-element-width'), $('#h-element-height'),
-                $('#h-nb-pixel'), $('#h-size-file'));
-            oldSettings.push($('#h-element-width').val(), $('#h-element-height').val(),
-                $('#h-nb-pixel').val(), $('#h-size-file').val());
-            runs(function () {
-                $('#h-element-mag').val(10).trigger('change');
-                var i = 0;
-                // Check all the setting labels change
-                for (var value in oldSettings) {
-                    expect(elements[i].val()).not.toEqual(value);
-                    i++;
-                }
-            });
-            runs(function () {
-                $('#h-download-image-format').val('TIFF').trigger('change');
-                // Check the size label change
-                expect($('#h-size-file').val()).not.toEqual(oldSettings[3]);
-            });
-        });
-
-        it('ensure the download link is correct', function () {
-            waitsFor(function () {
-                return $('#h-download-area-link').attr('href') !== undefined;
-            }, 'to be the url');
-
-            runs(function () {
-                expect($('#h-download-area-link').attr('href')).toMatch(/\/item\/[0-9a-f]{24}\/tiles\/region\?regionWidth=[0-9-]+&regionHeight=[0-9-]+&left=[0-9-]+&top=[0-9-]+&right=[0-9-]+&bottom=[0-9-]+&encoding=[EFGIJNPT]{3,4}&contentDisposition=attachment&magnification=[0-9-]+/);
-            });
-        });
-
-        it('close the dialog', function () {
-            $('#g-dialog-container').girderModal('close');
-            waitsFor(function () {
-                return $('body.modal-open').length === 0;
-            });
-        });
-    });
-
-    describe('Annotation tests', function () {
         describe('Draw panel', function () {
             var annotationInfo = {};
 
--- a/plugin_tests/client/annotationSpec.js
+++ b/plugin_tests/client/annotationSpec.js
@@ -147,30 +147,72 @@ $(function () {
             it('login', function () {
                 girderTest.waitForLoad();
 
-                runs(function () {
-                    $('.g-login').click();
-                });
-
-                girderTest.waitForDialog();
-                runs(function () {
-                    $('#g-login').val('user');
-                    $('#g-password').val('password');
-                    $('#g-login-button').click();
-                });
-
-                waitsFor(function () {
-                    return $('.h-user-dropdown-link').length > 0;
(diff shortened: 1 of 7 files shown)


jbeezley:
@Pierre-Assemat I added a commit that should get you started with the client side testing.  I left a couple of places for you to add some assertions.  Once that is done, we can take a look at where the coverage numbers are and think about where we can add more test cases.

Pierre-Assemat:
I write futher test, but I raise an error : 

screenshot from 2017-08-11 14-54-43

I don't really know why, do you have an idea ?

jbeezley:
I'm not sure.  What code did you add?

Pierre-Assemat:
How can I assure this PR to pass the tests ? 
`Test #29: web_client_HistomicsTK.annotations` is failed on travis but pass on local...

Do I need to update the branch with Master ?

jbeezley:
Sorry @Pierre-Assemat, I've been out.  I'm going to look into this error today.

Pierre-Assemat:
No it's ok, I asked David, and he explain me that it was caused by a PR #2273 and will be fix by the PR #2297 !

Pierre-Assemat:
I think there is still some work to do here, first the value of the size (bytes) isn't really close from the truth, then there are some visualisation bug on my instance, I don't know the origin, and if it's only in this PR, I prefere wait other review before merge it into master
What do you think @jbeezley @manthey ?

jcfr:
@Pierre-Assemat Well done, CI is green. Before merging, I suggest you squash together all commits into one. I can help you tomorrow with this.

And to capture the fact you worked with @jbeezley on this, you could add the following text at the end of the message:

```
This is the title of the commit

And some explaination

Co-authored-by: Jonathan Beezley <jonathan.beezley@kitware.com>
```

web_client/dialogs/editRegionOfInterest.js
```diff
@@ -0,0 +1,207 @@
...
+     * Get the size of the file in the appropriate unity (Bytes, MB, GB...)
+     */
+    getConvertFileSize() {
+        var Nbytes = this.getFileSize();
+        var convertedSize = formatSize(Nbytes);
+        if (Nbytes >= Math.pow(2, 30)) {
+            this.downloadDisable(true);
+        } else {
+            this.downloadDisable(false);
+        }
```
jbeezley: This can just be
```javascript
this.downloadDisable(Nbytes >= Math.pow(2, 30));
```

web_client/dialogs/editRegionOfInterest.js
```diff
@@ -0,0 +1,207 @@
...
+    getFileSize() {
+        var fileSize = (this.getNumberPixels() * 3 + this._sizeCte) * this._compressionRatio;
+        return fileSize;
+    },
+
+    /**
+     * Get the size of the file in the appropriate unity (Bytes, MB, GB...)
+     */
+    getConvertFileSize() {
+        var Nbytes = this.getFileSize();
```
jbeezley: Here an elsewhere, avoid making variables names starting with capital letters.  These are usually reserved for classes.

web_client/dialogs/editRegionOfInterest.js
```diff
@@ -0,0 +1,207 @@
...
+     * Disable the Download button if SizeFile > 1GB
+     */
+    downloadDisable(bool) {
+        var element = $('#msgDisable').attr('id');
+        if (bool === true) {
+            $('#download-submit').attr('disabled', 'disabled');
+            if (typeof element === typeof undefined) {
+                var msgDisable = $('<span></span>').text('Size > 1GB : Impossible Download ');
+                msgDisable.attr('id', 'msgDisable');
+                msgDisable.css({'color': 'red', 'margin-right': '120px'});
```
jbeezley: Move custom css to a stylesheet.  For the color, you can just add the `g-validation-failed-message` class to the element to keep the style consistent.

web_client/dialogs/editRegionOfInterest.js
```diff
@@ -0,0 +1,207 @@
...
+                $('#download-area-link').before(msgDisable);
+            }
+        } else if (bool === false) {
+            $('#download-submit').removeAttr('disabled');
+            if (typeof element !== typeof undefined) {
```
jbeezley: Normally you would want to use `_.isUndefined(element)` to check if a variable is undefined.  However in this case, I think the check you are doing is unnecessary.  It's usually better to allow a bug to throw an error, unless there is a specific reason to suppress it.

web_client/dialogs/editRegionOfInterest.js
```diff
@@ -0,0 +1,207 @@
...
+                break;
+            case 'TIFF':     // TIFF
+                this._format = 'TIFF';
+                this._compressionRatio = 0.8;
+                break;
+            default:     // JPEG is the default format
+                this._compressionRatio = 0.35;
+        }
+        this._magnification = parseFloat($('#h-element-mag').val());
+        var bounds = this.scaleBounds();
```
jbeezley: `var` definitions are always [hoisted](https://www.w3schools.com/js/js_hoisting.asp) to the top of the function.  It would be better to define the variables at the top or change all declarations to `let`/`const`.

web_client/panels/ZoomWidget.js
```diff
@@ -166,6 +178,77 @@ var ZoomWidget = Panel.extend({
...
+            this.viewer.annotationLayer.mode(null);
+            this._cancelSelection = false;
+            $('.h-download-button-area').css('background-color', '');
+        }
+        var hrefAttrView = this.$('a.h-download-link#download-view-link').attr('href');
+        if (typeof hrefAttrView === typeof undefined || hrefAttrView !== urlView) {
+            this.$('a.h-download-link#download-view-link').attr({
+                href: urlView
+            });
+        }
```
jbeezley: Setting an attribute is cheap.  I would just set it without trying to detect if it needs to be updated.

web_client/panels/ZoomWidget.js
```diff
@@ -166,6 +178,77 @@ var ZoomWidget = Panel.extend({
...
+            width: window.innerWidth,
+            height: window.innerHeight,
+            left: boundsTab[0],
+            top: boundsTab[1],
+            right: boundsTab[2],
+            bottom: boundsTab[3],
+            contentDisposition: 'attachment'
+        });
+        var urlView = `/${apiRoot}/item/${imageId}/tiles/region?${params}`;
+        if (this._cancelSelection === true) {
```
jbeezley: `if (this._cancelSelection) {`

web_client/templates/dialogs/editRegionOfInterest.pug
```diff
@@ -0,0 +1,39 @@
...
+        .form-group
+          label(for='nb-pixel') Number of Pixels
+          input#nb-pixel.update-form.form-control(
+            type='text', value=numberOfPixel, disabled)
+        .form-group
+          label(for='size-file') Approximate size of the file
+          input#size-file.update-form.form-control(type='text', value=fileSize, disabled)
+        .g-validation-failed-message.hidden
+      .modal-footer
+        a#download-area-link.h-download-link.btn.btn-small.btn-primary
```
jbeezley: In Girder's ecosystem, we have a convention of prefixing class and id names with a letter.  In girder core, we use `g-`.  In HistomicsTK, it is `h-`.  Could you add `h-` to all of the custom names you defined?

web_client/panels/ZoomWidget.js
```diff
@@ -166,6 +178,77 @@ var ZoomWidget = Panel.extend({
...
+     *
+     */
+    _downloadArea(evt) {
+        var mag = Math.round(this._getSliderValue() * 10) / 10;
+        var maxZoom = this._maxZoom;
+        var maxMag = this._maxMag;
+        if (this._cancelSelection === true) {
+            this.viewer.annotationLayer.mode(null);
+            this._cancelSelection = false;
+            $('.h-download-button-area').css('background-color', '');
```
jbeezley: Rather than set css on the element, just add or remove the `active` class on the button.

Also, here and elsewhere in this PR, you are referencing the global `$` object.  In most cases, this should be replaced with `this.$` to scope your selectors to just the current view.

web_client/templates/dialogs/editRegionOfInterest.pug
Pierre-Assemat: Thanks I didn't know !!

web_client/dialogs/editRegionOfInterest.js
Pierre-Assemat: I think I need this test, if the element don't exist and I don't want to disable the download
I enter in the 'else' and I check if the element : `this.$('#h-msgDisable')` exist before remove it
If I don't check the existance of the element, the remove will raise and error right ?

jbeezley: No, `$('invalid selector').remove()` will not throw an error.

manthey:
A few issues I've noticed:

When merged with the latest master, there is still a linting error and a test error.

I note that when you change the format control, it doesn't have any immediate affect -- the download still uses the previous format.  However, if you change the magnification, the format that becomes what was selected.

It behaves oddly on images that have no scale information (no magnification data).

Pierre-Assemat:
I found why the format wasn't update thank you @manthey !

When you say a linting error and test error, what do you mean ? A failure of a test ?

manthey:
The tests are failing on travis.  If you follow the latest travis PR log, you'll see
```
9: Test command: /home/travis/build/DigitalSlideArchive/HistomicsTK/build/girder/node_modules/.bin/eslint "--ignore-path" "/home/travis/build/DigitalSlideArchive/HistomicsTK/build/girder/.eslintignore" "--config" "/home/travis/build/DigitalSlideArchive/HistomicsTK/build/girder/.eslintrc.json" "/home/travis/build/DigitalSlideArchive/HistomicsTK/build/girder/plugins/HistomicsTK/web_client"
9: Test timeout computed to be: 1500
9: 
9: /home/travis/build/DigitalSlideArchive/HistomicsTK/build/girder/plugins/HistomicsTK/web_client/panels/ZoomWidget.js
9:   11:1  error  More than 1 blank line not allowed  no-multiple-empty-lines
9: 
9:  1 problem (1 error, 0 warnings)
9: 
26/31 Test  #9: eslint_HistomicsTK ...................................................***Failed    2.71 sec
```
You can also run these tests locally before you push a commit.

web_client/dialogs/editRegionOfInterest.js
```diff
@@ -0,0 +1,194 @@
...
+                break;
+            case 'TIFF':
+                this._format = 'TIFF';
+                this._compressionRatio = 0.8;
+                break;
+            default:
+                // JPEG is the default format
+                this._compressionRatio = 0.35;
+        }
+        this._magnification = parseFloat($('#h-element-mag').val());
```
manthey: The `const bounds = this.scaleBounds();` line above needs to come after the `_magnification` is set, otherwise the bounds  aren't correct.

web_client/panels/ZoomWidget.js
```diff
@@ -166,6 +180,74 @@ var ZoomWidget = Panel.extend({
         this._zoomSliderChange();
     },
 
+     /**
+     * A handler called when the download view button is clicked.
+     */
+    _downloadView(evt) {
+        var imageId = router.getQuery('image');
+        var bounds = router.getQuery('bounds');
+        var boundsTab = bounds.split(',');
```
manthey: There may not be any bounds in the route if an image was just loaded.

Pierre-Assemat: I'm not sure to see what you're talking about, I tried to reproduce this in downloading an image that was just upload, this didn't cause any issue, and the bounds were set. What kind of changes could be more appreciate here ?

manthey: I only saw it when I tried to merge the branch into master and get the tests to pass.  There was a merge conflict with some of the tests, and I may have not resolved them appropriately.

manthey: We probably should be using `this.viewer.bounds()` rather then reparsing the router information.

Pierre-Assemat: I'm resolving the conflict and fix the `annotations test`, I will see how I can use the `viewer.bounds` rather than using the router information. Thank you !

Pierre-Assemat:
Do you have any idea of what issue can cause this kind of display ?

screenshot from 2017-10-16 13-32-09

manthey:
I'd guess that there is a javascript exception being thrown which stops some aspect of the renderer from running.

manthey:
Currently, we try to download a large region, the original image may need to access more than 0.5 gigapixels.  If so, PIL can throw an out-of-memory error, which results in a failure.  The replaces the HistomcsTK interface with the error message.  If we add a `target='_blank'` to the download link area, this at least appears in another window/tab (e.g., change the download link in pug to
```
a#h-download-area-link.h-download-link.btn.btn-small.btn-primary(target='_blank')
```
).  This isn't quite as nice looking as not opening a new tab, but guards against this error.

Pierre-Assemat:
This should happens only when we use the download button, but I have this kind of bad rendering each time I change the zoom value.

manthey:
@Pierre-Assemat Do you still have the problem with the zoom messing up the display?  Are you on the latest large_image and girder?  I haven't been able to reproduce the issue locally.

@jbeezley Would you care to take another look at this?

Pierre-Assemat:
@manthey Yes I have some weird behavior, I try to find the origin but I don't know.

First when I open the page the style is really weird but I don't see any error on the console log: 
screenshot from 2017-11-07 15-18-34

Then, each time the zoom is changed the render don't work well: 
screenshot from 2017-11-07 15-21-39

I rebased on a master just after fetch it, I also updated girder, these errors has to come from one of the changes in the PR

jbeezley:
I'll give this branch a try.  If I can't reproduce it, we may need to do a hangout to debug what's going wrong in your development environment.

jbeezley:
This branch works for me as is.  I'll schedule a hangout with you to debug your development environment.

Pierre-Assemat:
Ok sounds great, thank you

plugin_tests/client/annotationSpec.js
```diff
@@ -170,7 +169,91 @@ $(function () {
                 });
             });
         });
+    });
+
+    describe('Download view and region of interest', function () {
```
manthey: Can you indent this entire describe block?  This will keep the outer describe block as a single block rather than divided into multiple describe blocks with the same description.

manthey:
@jbeezley I've asked @Pierre-Assemat for one minor change in the layout of the tests, but otherwise I'm content to get this merged in if you are okay with it.  If we want further changes they can be in a different PR.

plugin_tests/client/annotationSpec.js
Pierre-Assemat: Do you mean put this entire block into the discribe(Annotation tests) and remove the second discribe(Annotation tests) in order to keep only one block with this description ?

manthey: Yes.  I'll just push the change.

PR #361, opened 2017-07-27
Internalize positive pixel count routines

neal-siekierski:
This PR takes the primary code for performing positive pixel count and moves it from the `PositivePixelCount` CLI into a new `positive_pixel_count` module, which I've chosen to put under `histomicstk.preprocessing`.

The motivation for this change is to allow creating the same sort of Jupyter notebook documentation that we have for color deconvolution, which should be more informative with an actual Python API.


histomicstk/preprocessing/positive_pixel_count.py
```diff
@@ -0,0 +1,232 @@
...
+    'IntensitySumWeakPositive',
+    'IntensitySumPositive',
+    'IntensitySumStrongPositive',
+])
+
+PPCOutput = namedtuple('PPCOutput', PPCOutputTotals._fields + (
+    'IntensityAverage',
+    'RatioStrongToTotal',
+    'IntensityAverageWeakAndPositive',
+))
```
cdeepakroy: I feel we can combine PPCOutputTotals and PPCOutput into one. What's the reason to keep them separate?

neal-siekierski: (We discussed this in person. For others' reference, the reason is that this allows returning only a `PPCOutputTotals` from `_positive_pixel_count_simple`, thereby allowing, in the case that `_positive_pixel_count_simple` is called multiple times (namely when processing multiple tiles with Dask), only computing the extra fields in `PPCOutput` at the end, which is the only time the computation results are actually needed.)

histomicstk/preprocessing/positive_pixel_count.py
```diff
@@ -0,0 +1,232 @@
...
+        results = delayed(_combine)(results).compute()
+    return _totals_to_stats(results),
+
+
+def _combine(results):
+    return PPCOutputTotals._make(sum(r[i] for r in results)
+                                 for i in range(len(PPCOutputTotals._fields)))
+
+
+def positive_pixel_count_tiles(slide_path, ppc_params, kwargs, position, count):
```
cdeepakroy: Prefix this function with `_` as this is an internal function.

histomicstk/preprocessing/positive_pixel_count.py
```diff
@@ -0,0 +1,232 @@
...
+])
+
+PPCOutput = namedtuple('PPCOutput', PPCOutputTotals._fields + (
+    'IntensityAverage',
+    'RatioStrongToTotal',
+    'IntensityAverageWeakAndPositive',
+))
+
+
+def positive_pixel_count(slide_path, ppc_params, region=None,
```
cdeepakroy: Rename to `postive_pixel_count_wsi`

docs/histomicstk.preprocessing.rst
```diff
@@ -12,4 +12,5 @@ histomicstk.preprocessing
     color space conversion <histomicstk.preprocessing.color_conversion>
     color deconvolution <histomicstk.preprocessing.color_deconvolution>
     color normalization <histomicstk.preprocessing.color_normalization>
+    positive pixel counting <histomicstk.preprocessing.positive_pixel_count>
```
cdeepakroy: Move it to the segmentation package instead. It fits better there.

histomicstk/preprocessing/positive_pixel_count.py
```diff
@@ -0,0 +1,232 @@
...
+    total = [0] * lpotf
+    for pos in range(position, position + count):
+        tile = ts.getSingleTile(tile_position=pos, **kwargs)['tile']
+        subtotal = _positive_pixel_count_simple(tile, ppc_params)[0]
+        for k in range(lpotf):
+            total[k] += subtotal[k]
+    return PPCOutputTotals._make(total)
+
+
+def positive_pixel_count_simple(image, parameters):
```
cdeepakroy: Rename `parameters` as `ppc_params` to be consistent with the slide version

neal-siekierski:
I've made the requested changes. I've used `count_slide` instead of `count_wsi` because there's no requirement that the provided image have magnification metadata or such, and because HistomicsTK already refers to both "slide"s and "wsi"s. More subjectively, I also don't find that the use of the acronym adds significant clarity.

histomicstk/segmentation/positive_pixel_count.py
```diff
@@ -0,0 +1,232 @@
...
+    total = [0] * lpotf
+    for pos in range(position, position + count):
+        tile = ts.getSingleTile(tile_position=pos, **kwargs)['tile']
+        subtotal = _count_simple(tile, params)[0]
+        for k in range(lpotf):
+            total[k] += subtotal[k]
+    return OutputTotals._make(total)
+
+
+def count_simple(image, params):
```
cdeepakroy: Rename this to `count_image`.

Rest LGTM!

neal-siekierski: Done. Will merge once the build passes.

PR #363, opened 2017-07-31
Work around user and group ID issues in Docker.

manthey:
This merges the two set_hosts.py scripts into a new set_environment.py script that works for both the worker and girder roles.  This script sets up some /etc/hosts values (as before), can set the worker tmp_root if a command line flag is specified, can adjust the specified user's UID and GID, and adds the specified user to the host's docker group.  If the UID or GID of the user changes, all files are changed appropriately and the ~/.ansible directory is removed (somehow it remained stuck to the old UID).

The docker scripts have been modified to run the appropriate commands with the new groups, so the docker containers no longer need to be restarted to pick up the group changes.

This has been tested with a system with different UID, GID, and docker GID.  It would probably be good to test it with a system with a conflicting UID or GID.


jcfr:
FWIW, using `gosu` could help here. 

Here is what we have been doing in dockcross:

* https://github.com/dockcross/dockcross/blob/da5aebb14f7b3df5093604ecc45db34f42a8dac4/imagefiles/entrypoint.sh#L19-L51
* https://github.com/dockcross/dockcross/blob/master/imagefiles/sudo.sh

cc: @thewtex

thewtex:
gosu: https://github.com/tianon/gosu

manthey:
I've considered refactoring this to use `chroot` (of `gosu`), but it is working as is for our purposes, and I'd prefer to push that to a later PR.

manthey:
This has been confirmed working in the field.

manthey:
I'd love to get this tested and merged in (ping @jbeezley, @zachmullen, and @cdeepakroy).

As a way to show it is working, you can create a Vagrantbox using a stock Ubuntu 16.04 image.  Log in and create a new user.  Install docker and git clone this branch.  Make that user a member of the docker group.  Log in as the new user.  Run `python docker_deploy.py start --latest --cli-test` in the ansible directory.  This should result in success and have files in ~/.histomicstk that are owned by the new user (not UID 1001).

PR #364, opened 2017-08-01
Some architectural features

neal-siekierski:
This PR adds the function `compute_global_graph_features` to HistomicsTK, namely in the `histomicstk.features` package. It computes global features in the form of summary statistics over a collection of centroids, namely ones relating to the Voronoi tessellation, Delaunay triangulation, minimum spanning tree, and density of the centroids.

A test based on a small example (6 points) is included.

### DIFF
--- a/histomicstk/features/compute_global_graph_features.py
+++ b/histomicstk/features/compute_global_graph_features.py
@@ -33,18 +33,28 @@ def compute_global_graph_features(centroids, scale=10.):
     -------
     props : collections.namedtuple
         Nested namedtuples with the structure:
+
         - .voronoi: Voronoi diagram features
+
           - .area: Polygon area features
           - .peri: Polygon perimeter features
+
         - .delaunay: Delaunay triangulation features
+
           - .sides: Triangle side length features
           - .area: Triangle area features
+
         - .mst_branches: Minimum spanning tree branch features
         - .density: Density features
+
           - .neighbors_in_distance
+
             - [radius]: Neighbor count within given radius features
+
           - .distance_for_neighbors
+
             - [count]: Minimum distance to enclose count neighbors features
+
         Each leaf node is itself a namedtuple with fields 'mean',
         'stddev', 'minmaxr', and 'disorder'.  'minmaxr' is the
         minimum-to-maximum ratio, and disorder is stddev / (mean +
--- a/histomicstk/features/compute_global_graph_features.py
+++ b/histomicstk/features/compute_global_graph_features.py
@@ -17,7 +17,7 @@ DensityProps = namedtuple('DensityProps', ['neighbors_in_distance',
 Props = namedtuple('Props', ['voronoi', 'delaunay', 'mst_branches', 'density'])
 
 
-def compute_global_graph_features(centroids, scale=10.):
+def compute_global_graph_features(centroids, scale=10., neighbor_counts=(3, 5, 7)):
     """Compute global (i.e., not per-pixel) graph-based features of the
     nuclei with the given centroids.
 
@@ -28,6 +28,10 @@ def compute_global_graph_features(centroids, scale=10.):
     scale : float
         Multiplier for distances.  Used for determining the radii to
         count neighbors in.
+    neighbor_counts : sequence
+        Sequence of numbers of neighbors, each of which is used to
+        compute statistics relating to the distance required to reach
+        that many neighbors.
 
     Returns
     -------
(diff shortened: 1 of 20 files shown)


histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    vor = Voronoi(centroids)
+    centroids = vor.points
+    vertices = vor.vertices
+
+    regions = [r for r in vor.regions if r and -1 not in r]
+    areas = np.stack(poly_area(vertices[r]) for r in regions)
+    peris = np.stack(poly_peri(vertices[r]) for r in regions)
+    # TODO Consider using more numpy
+    max_dists = np.stack(max(dist(x, y) for x, y in combinations(vertices[r], 2)) for r in regions)
```
cdeepakroy: You can use `scipy.spatial.distance.pdist` here.

histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+from collections import namedtuple
+from itertools import combinations
+
+import numpy as np
+from numpy import linalg
+from scipy.spatial import cKDTree as KDTree, Voronoi
+from scipy import sparse
+from scipy.sparse.csgraph import minimum_spanning_tree
+
+PopStats = namedtuple('PopStats', ['mean', 'stddev', 'minmaxr', 'disorder'])
```
cdeepakroy: rename `minmaxr` to `minmaxratio`

histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+    distance_for_neighbors = dict(zip(
+        neighbor_counts,
+        map(pop_stats, tree.query(centroids, [c + 1 for c in neighbor_counts])[0].T),
+    ))
+    density_props = DensityProps(neigbors_in_distance, distance_for_neighbors)
+
+    return Props(poly_props, tri_props, mst_branches, density_props)
+
+
+def poly_area(vertices):
```
cdeepakroy: Prefix all functions not part of public API with `_`

histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+def pop_stats(pop):
+    # Filter out outliers (here defined as points more than three
+    # standard deviations away from the mean)
+    while True:
+        mean = pop.mean()
+        stddev = pop.std()
+        mask = abs(pop - mean) <= 3 * stddev
+        if mask.all():
+            break
+        pop = pop[mask]
```
cdeepakroy: `mean` and `stddev` are sensitive to outliers. 

Use [this](https://www.datacrucis.com/research/find-outliers-in-an-array.html) IQR based approach.

Also compute all population stats after outlier rejection.

histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+    """Compute global (i.e., not per-pixel) graph-based features of the
+    nuclei with the given centroids.
+
+    Parameters
+    ----------
+    centroids : array_like
+        Nx2 numpy array of nuclear centroids
+    scale : float
+        Multiplier for distances.  Used for determining the radii to
+        count neighbors in.
```
cdeepakroy: Use an array of radii instead rather internally generating it with `scale * np.arange(6)`. You can use the settings in the paper as the default value.

histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+            - [radius]: Neighbor count within given radius features
+
+          - .distance_for_neighbors
+
+            - [count]: Minimum distance to enclose count neighbors features
+
+        Each leaf node is itself a namedtuple with fields 'mean',
+        'stddev', 'minmaxr', and 'disorder'.  'minmaxr' is the
+        minimum-to-maximum ratio, and disorder is stddev / (mean +
+        stddev).
```
cdeepakroy: Return a pandas dataframe instead of a nest named tuple similar to other feature computation functions. 

Ultimately we will concat different types of features which will be easy if all of them return in the same format. Otherwise, client code of this function will have to write code to convert this one to a pandas dataframe

histomicstk/features/compute_global_graph_features.py
```diff
@@ -0,0 +1,168 @@
...
+
+PopStats = namedtuple('PopStats', ['mean', 'stddev', 'minmaxr', 'disorder'])
+PolyProps = namedtuple('PolyProps', ['area', 'peri', 'max_dist'])
+TriProps = namedtuple('TriProps', ['sides', 'area'])
+DensityProps = namedtuple('DensityProps', ['neighbors_in_distance',
+                                           'distance_for_neighbors'])
+Props = namedtuple('Props', ['voronoi', 'delaunay', 'mst_branches', 'density'])
+
+
+def compute_global_graph_features(centroids, scale=10., neighbor_counts=(3, 5, 7)):
```
neal-siekierski: If we are considering adding "local" versions of these -- ones that give per-nucleus values -- as well as "global" ones that give values describing an entire region, how about `compute_global_cell_graph_features`?

cdeepakroy: `compute_global_cell_graph_features` sounds good to me!

PR #366, opened 2017-08-03
Add positive pixel count example

neal-siekierski:
This PR adds examples of the use of the new `histomicstk.segmentation.positive_pixel_count` module as well as the CLI. Feedback is welcome for the writing, the particular examples given (I didn't spend that much time trying to find good parameter values), or anything else you may think of.


cdeepakroy:
@cooperlab Can you please take a look at this when you can? Maybe we should use a DAB image instead.

cooperlab:
@cdeepakroy @neal-siekierski I think we should use DAB - do you have an example or do you need one?

cdeepakroy:
Yes, can you please send Neal a DAB example to use?

neal-siekierski:
@cooperlab Any other thoughts?

PR #370, opened 2017-08-31
Item tasks migration

itcthienkhiem:


jbeezley:
@itcthienkhiem Do you have an interest in the branch or was this a misclick?  Unfortunately, it isn't quite ready yet.  I still need to add back a few features on the client and migrate some CLI related code from the slicer_cli_web repository.

PR #372, opened 2017-09-07
Rename an ansible variable to avoid an upstream name conflict.

manthey:


manthey:
@zachmullen or @jbeezley This should be a quick PR to review, if either of you gets a chance.

PR #375, opened 2017-09-20
Make drawing tools stay active.

manthey:
After drawing an annotation, stay in the drawing mode until the drawing button is clicked again.


web_client/panels/DrawWidget.js
```diff
@@ -98,18 +107,34 @@ var DrawWidget = Panel.extend({
...
+     *      mode.
      */
-    drawElement(evt) {
-        var $el = this.$(evt.currentTarget);
-        var type = $el.data('type');
-        if ($el.hasClass('active')) {
+    drawElement(evt, type) {
+        var $el;
+        if (evt) {
+            $el = this.$(evt.currentTarget);
```
jbeezley: For some reason, the tooltips stay open after clicking a button with this change.  Adding `$el.tooltip('hide')` here gets rid of it.

Issue #379, opened 2017-09-21
'odict_keys' object has no attribute 'reverse'


hsinewu:
Ubuntu 16.04, reading [doc](https://digitalslidearchive.github.io/HistomicsTK/installation.html#prerequisites)

I try to install with docker but when it starts, localhost:8080 shows `connection was reset`

So I try to pull updates, wonder would it fix the problem, but then I got the following error when I try to run  deploy_docker stop or rm.
```
$ python deploy_docker.py stop # or rm
Traceback (most recent call last):
  File "deploy_docker.py", line 869, in <module>
    containers_stop(remove=args.command in ('rm', 'remove'))
  File "deploy_docker.py", line 418, in containers_stop
    keys.reverse()
AttributeError: 'odict_keys' object has no attribute 'reverse'
```

It seems like the container is still occupying 8080, and since `stop` doesn't work I can't shut it down.

btw not sure if it matters, but seems like I didn't stop the container before pulling updates.

manthey:
It sounds like a Python 3 issue.  I don't think we've tested the `deploy_docker.py` script on Python 3 (only on Python 2).  I'll see what needs to be done to make it Python 3 compatible.

In the mean time, you can stop and remove the containers manually.  Type `docker ps -a` to list the containers.  `docker rm -f <container id> <container id ...>`.  You can probably also remove them by name: `docker rm -f histomicstk_histomicstk histomicstk_girder_worker histomicstk_mongodb histomicstk_rmq`.

hsinewu:
Thanks, I can remove the containers with that command, now I have another problem.

Running `start` shows following message.
```
$ python deploy_docker.py start
Creating rabbitmq:management - histomicstk_rmq
Starting rabbitmq:management - histomicstk_rmq
Creating mongo:latest - histomicstk_mongodb
Starting mongo:latest - histomicstk_mongodb
Creating dsarchive/girder_worker:v0.1.3 - histomicstk_girder_worker
Starting dsarchive/girder_worker:v0.1.3 - histomicstk_girder_worker
Creating dsarchive/histomicstk_main:v0.1.3 - histomicstk_histomicstk
Starting dsarchive/histomicstk_main:v0.1.3 - histomicstk_histomicstk
Waiting for Girder to report version: ..Traceback (most recent call last):
  File "/home/hsinewu/anaconda3/lib/python3.6/site-packages/docker/api/client.py", line 220, in _raise_for_status
    response.raise_for_status()
[...]

PR #380, opened 2017-09-21
Fix python3 issues with deploy_docker.

manthey:
This has been tested with start, provision, stop, and remove.

This fixes the issue reported in #379.

### DIFF
--- a/ansible/utils/cli_test.py
+++ b/ansible/utils/cli_test.py
@@ -123,7 +123,7 @@ def test_cli(client, folder, opts):
         if item['name'].startswith('TCGA-02'):
             testItem = item
             break
-    localFile = client.listFile(testItem['_id']).next()
+    localFile = next(client.listFile(testItem['_id']))
     path = 'HistomicsTK/%s/NucleiDetection/run' % (
         opts['cli'].replace('/', '_').replace(':', '_'), )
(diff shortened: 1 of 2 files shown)


ansible/utils/cli_test.py
```diff
@@ -123,7 +123,7 @@ def test_cli(client, folder, opts):
         if item['name'].startswith('TCGA-02'):
             testItem = item
             break
-    localFile = client.listFile(testItem['_id']).next()
+    localFile = next(client.listFile(testItem['_id']))
```
zachmullen: The py2/3 safe way to call this would be `six.next(...)`

manthey: I thought `next` worked the same in Python 2 and Python 3 provided Python >= 2.6.

zachmullen: I have encountered situations in the past where `next` did not work but replacing with `six.next` did, but those may have been weird edge cases of iterators or something. If this works, we can go with it.

manthey:
We are assuming that the config file is written in utf8, which is the same assumption made in Girder.  For Girder, in general, I think we *want* to assume this.  Otherwise, we need to expose some method to know what the config file is encoded in (which would go in a config file...).

PR #381, opened 2017-09-22
Update the pinned versions of the docker images.

manthey:
The older versions don't play well with the changes made to handle docker ID issues.


manthey:
@cdeepakroy Thanks.  Since this is based on the `deploy-docker-python3`, would you mind reviewing that as well?

PR #382, opened 2017-09-26
A CLI implementation of autoencoder

slee172:
This PR implements two CLIs for extracting encoded features from Superpixels.

1. TrainSuperpixelFeatures

Input: 
- slide path and a compiled autoencoder model.

Output:
- a trained autoencoder model

2. ComputeSuperpixelFeatures

Input: 
- slide path and a trained autoencoder model.

Output:
- slide information(slide list, slide index, first slide index, mean, std, encoded features
        x, y centroids, and boundary file)


server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.xml
```diff
@@ -0,0 +1,190 @@
...
+      <description>Tile size for blockwise analysis</description>
+      <longflag>analysis_tile_size</longflag>
+      <default>4096</default>
+    </double>
+    <double>
+      <name>analysis_mag</name>
+      <label>Analysis Magnification</label>
+      <description>The magnification at which the analysis should be performed.</description>
+      <longflag>analysis_mag</longflag>
+      <default>20</default>
```
cdeepakroy: Are you using 5x for your analysis? If so i would suggest setting that as default.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+    im_hematoxylin_stain = im_stains[:, :, 0].astype(np.float)/255.0
+    im_eosin_stain = im_stains[:, :, 1].astype(np.float)/255.0
+
+    # get the number of super-pixels
+    im_width, im_height = im_tile.shape[:2]
+    n_superpixels = (im_width/args.patchSize) * (im_height/args.patchSize)
+
+    # perform SLIC to get labels with max compactness
+    # slic has a value 0 as a first label, so added 1
+    im_label = slic(im_nmzd, n_segments=n_superpixels, compactness=100) + 1
```
cdeepakroy: Expose compactness as a parameter to the CLI with a reasonable default value that works on multiple slides.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+    cdata = np.zeros((n_labels, 2))
+
+    for i in range(n_labels):
+        cen_x, cen_y = region_props[i].centroid
+
+        # get bounds of region
+        min_row, max_row, min_col, max_col = \
+            get_patch_bounds(cen_x, cen_y, args.patchSize, im_width, im_height)
+
+        # get image centers at low-res
```
cdeepakroy: This comment is misleading. The code below is computing the centers in space of highest-res layer or basepixel units.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+
+        # append result to list
+        tile_result_list.append(cur_result)
+
+    tile_result_list = dask.delayed(tile_result_list).compute()
+
+    print 'Number of tiles processed = %d' % n_tiles
+    print('\n>> tile_result_list to centers and features ...\n')
+
+    slide_centers = np.asarray([centers
```
cdeepakroy: Rename to `superpixel_centers` to be clear.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+                                for cdata, fdata in tile_result_list
+                                for centers in cdata])
+
+    slide_pixel_array = np.asarray([features
```
cdeepakroy: rename to `superpixel_data` to be clear

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+    return cdata, fdata
+
+
+def get_encoded_features(input_tensors, input_model, args):
+
+    # load input model
+    autoencoder = load_model(input_model)
+
+    # input: the first input of the model, output: the last output of the model
+    encoder = Model(inputs=autoencoder.layers[0].input, outputs=autoencoder.layers[6].output)
```
cdeepakroy: Layer indexes are hard-coded

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+from ctk_cli import CLIArgumentParser
+
+import logging
+logging.basicConfig(level=logging.CRITICAL)
+
+sys.path.append(os.path.normpath(os.path.join(os.path.dirname(__file__), '..')))
+from cli_common import utils as cli_utils  # noqa
+
+
+def compute_input_tensors(slide_path, tile_position, args, **it_kwargs):
```
cdeepakroy: Maybe rename to `compute_super_pixel_data`

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+
+    # load input model
+    autoencoder = load_model(input_model)
+
+    # input: the first input of the model, output: the last output of the model
+    encoder = Model(inputs=autoencoder.layers[0].input, outputs=autoencoder.layers[6].output)
+
+    # fit to autoencoder with the number of epochs and the batch size
+    autoencoder.fit(input_tensors, input_tensors, epochs=args.epochs, batch_size=args.batchSize,
+                    shuffle=True, validation_data=(input_tensors, input_tensors))
```
cdeepakroy: This is something to be discussed i think. 

Why are we training the auto-encoder from scratch on super-pixel data of this slide? 

Would it be better to train the autoencoder on super-pixels of a set of training slides. After that, for any test slide, we would use the pretrained autoencoder to reduce the dimensionality of the super-pixel data.

@cooperlab Can you pitch in here?

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+        output = h5py.File(args.outputSuperpixelFeatureFile, 'w')
+        output.create_dataset('slides', data=slide_names)
+        output.create_dataset('slideIdx', data=slide_index)
+        output.create_dataset('dataIdx', data=slide_patch_index)
+        output.create_dataset('mean', data=slide_feature_mean)
+        output.create_dataset('std_dev', data=slide_feature_stddev)
+        output.create_dataset('features', data=encoded_features)
+        output.create_dataset('x_centroid', data=slide_center_x_list)
+        output.create_dataset('y_centroid', data=slide_center_y_list)
+        output.close()
```
cdeepakroy: In addition to the center, may be we should also store the patch-size. People may run this CLI with different patch sizes. If that is not stored in the output file format then, it will get lost.

Also, we need make any changes necessary to minimize the effort in loading this file into HistomicsML

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
cooperlab: This parameter is a bit abstract - at the least we need to describe what the effect of increasing or decreasing is on the superpixels

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
cooperlab: Not cheating since we are dealing with unsupervised learning. It's like performing PCA on your entire dataset which is an accepted practice. Your autoencoders might not perform as well on prospective data, but we aren't using any data labels during training.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+
+        fgnd_frac_comp_time = time.time() - start_time
+
+        print 'Number of foreground tiles = %d (%.2f%%)' % (
+            num_fgnd_tiles, percent_fgnd_tiles)
+
+        print 'Time taken = %s' % cli_utils.disp_time_hms(fgnd_frac_comp_time)
+
+    #
+    # Detect and compute nuclei features in parallel using Dask
```
cooperlab: What nuclei features?

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,387 @@
...
+    im_nmzd = htk_cnorm.reinhard(im_tile,
+                                 args.reference_mu_lab,
+                                 args.reference_std_lab)
+
+    # perform color decovolution
+    w = cli_utils.get_stain_matrix(args)
+
+    im_stains = htk_cdeconv.color_deconvolution(im_nmzd, w).Stains
+
+    im_hematoxylin_stain = im_stains[:, :, 0].astype(np.float)/255.0
```
cooperlab: This implies that we are analyzing hematoxylin and eosin images. I think there is a way to detect the number of real stains using the 'w' matrix. Names should be neutral and number dependent on inputs.

cooperlab: You could also get the number of input channels from the model too.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: @cdeepakroy @cooperlab For this, just wanted to mention that this implementation was for testing only one slide not multiple slides. For multiple slides, we should be able to extract image patches for each slide, and then feed the image patches into the autoencoder model so that we could get the entire training results for the multiple slides. Again, this implementation would not produce any good result for other slides. I think we should discuss about this in the meeting.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Renamed to compute_superpixel_data

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Added a dictionary to identify stains

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Modified to the highest-res.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Added some lines to detect encoding layers

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Fixed.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Renamed to slide_superpixel_data.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
slee172: Will be renamed to slide_x_centroids and slide_y_centroids

server/ComputeSuperpixelFeatures/TrainSuperpixelFeatures.xml
```diff
@@ -0,0 +1,218 @@
...
+    </region>
+    <integer>
+      <longflag>maxRegionSize</longflag>
+      <label>Maximum region size</label>
+      <description>Maximum width and height allowed when processing an image, in order to prevent accidentally running on too large a region.  Use -1 for no limit</description>
+      <default>5000</default>
+    </integer>
+  </parameters>
+  <parameters advanced="true">
+    <label>Autoencoder Training</label>
```
slee172: Added Autoencoder Training label for some args related to autoencoder training and superpixel

server/ComputeSuperpixelFeatures/TrainSuperpixelFeatures.xml
```diff
@@ -0,0 +1,218 @@
...
+      <label>Red and green ratio for superpixel</label>
+      <description>Red and green ratio for superpixel</description>
+      <default>1.3</default>
+    </double>
+    <double>
+      <longflag>min_var_superpixel</longflag>
+      <label>Minumum variance of superpixel</label>
+      <description>Minumum variance of superpixel</description>
+      <default>0.002</default>
+    </double>
```
slee172: Added some args to remove background region for superpixel. Simply get the foreground region using red and green ratio and variance of superpixel region.

server/ComputeSuperpixelFeatures/TrainSuperpixelFeatures.py
```diff
@@ -0,0 +1,378 @@
...
+              cli_utils.disp_time_hms(superpixel_detection_time)
+
+    #
+    # Train input model
+    #
+    print('\n>> Train input model ...\n')
+
+    start_time = time.time()
+
+    model = create_train_model(slide_superpixel_data, args.inputModelFile, args)
```
slee172: Here, the trained model is created.

slee172:
@cdeepakroy @manthey I got an error [here](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/286806379#L6943), so added # noqa: C 901 next to the main function. However, I still didn't get it passed. Could you take a look at [this](https://travis-ci.org/DigitalSlideArchive/HistomicsTK/builds/286820153)?

manthey:
This is a transient error of some sort -- I'll hunt it down later, but I just hit the rebuild button on travis and it succeeded.

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.xml
```diff
@@ -0,0 +1,245 @@
...
+    <integer>
+      <name>maxRegionSize</name>
+      <longflag>maxRegionSize</longflag>
+      <label>Maximum region size</label>
+      <description>Maximum width and height allowed when processing an image, in order to prevent accidentally running on too large a region.  Use -1 for no limit</description>
+      <default>5000</default>
+    </integer>
+  </parameters>
+<parameters advanced="true">
+    <label>Autoencoder Testing</label>
```
cdeepakroy: Rename to Autoencoder

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.xml
```diff
@@ -0,0 +1,245 @@
...
+    <description>Autoencoder test parameters</description>
```
cdeepakroy: Change to Autoencoder parameters

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.xml
```diff
@@ -0,0 +1,245 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<executable>
+  <category>HistomicsTK</category>
+  <title>Computes Encoded Features</title>
```
cdeepakroy: Computes Superpixel Features

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.xml
```diff
@@ -0,0 +1,245 @@
...
+  <description>Computes encoded features for Superpixels</description>
```
cdeepakroy: Computes superpixel features using an autoencoder

server/ComputeSuperpixelFeatures/ComputeSuperpixelFeatures.py
```diff
@@ -0,0 +1,558 @@
...
+            for y_c in y_cent:
+                y_centroids.append(y_c)
+
+            for x_b in x_brs:
+                x_boundaries.append(x_b)
+
+            for y_b in y_brs:
+                y_boundaries.append(y_b)
+
+        superpixel_data = np.asarray(superpixel_data, dtype=np.float32)
```
cdeepakroy: This size of `superpixel_data` could be huge. It would essentially be equal to all foreground part of the image that can be quite large. The scheduler machine may not have enough RAM to store it.

I would suggest moving the autoencoder feature extraction part into the function `compute_super_pixel_data` that you parallelize with dask.

slee172:
This PR is outdated and should be updated with the most current work. I will update soon for the new ones.

PR #385, opened 2017-09-28
Use python libtiff 0.4.1.

manthey:
Also, use libtiff 4.0.8.


manthey:
Yes -- large_image now uses pylibtiff 0.4.1, and this removes the overrides from what is specified there.

PR #388, opened 2017-10-02
Use the fileId, not originalId for large images.

manthey:
Before, the CLI was passed the original image, not the large image.  This was a hold over from before CLIs used large image.

### DIFF
--- a/web_client/views/body/ImageView.js
+++ b/web_client/views/body/ImageView.js
@@ -172,9 +172,7 @@ var ImageView = View.extend({
      *    to make another rest call to get the files contained
      *    in the item.
      *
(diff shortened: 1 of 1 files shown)


web_client/views/body/ImageView.js
```diff
@@ -225,7 +223,7 @@ var ImageView = View.extend({
             // the original file if available.
             promise = $.when(
                 getTilesDef(this.model.id),
-                getFileModel(largeImage.originalId || largeImage.fileId)
+                getFileModel(largeImage.fileId || largeImage.originalId)
```
jbeezley: Could you update or delete the comment here as well?

PR #391, opened 2017-10-03
[WIP] Item tasks migration

jbeezley:
This branch is now functional, so it would be a good idea to start the review/migration process.  In particular, I have only tested this with the NucleiDetection, so other analyses may not work.  It is likely there are changes that went in to slicer_cli_web that I missed and need to be ported over item_tasks.  I also haven't yet ported over the deployment scripts, but that shouldn't be too hard.

There are a few places where the UI is different from before.  Some of these could be seen as a regression, specifically
* The analysis dropdown is now a dialog containing the task list widget.  There isn't currently a way to filter the tasks that are shown.  Completing https://github.com/girder/girder/pull/2004 would allow us to attach a specific tag to Histomics related analyses.
* The grouping of parameters is now merged into two panels (inputs and outputs) and it doesn't respect the original ordering.  Some simple changes to item_tasks will allow us to at least show the important parameters (those without defaults) toward the top.  We will have to make additions to the item_tasks schema to restore the multiple panels.
* My unmerged PR at https://github.com/girder/slicer_cli_web/pull/43 isn't implemented, nor is handling of any output types other than girder uploads.  We need to somehow encode the parameters file side-effect into item_tasks.  I'm not yet certain how to do that in a task agnostic way.

The good news is that everything related to draw, editing, and displaying annotations and images is untouched.


manthey:
Closing this, as the Girder 3 version of slicer_cli_web replaces it.

PR #397, opened 2017-10-04
Add autosaving of annotations to the DrawWidget

### LINKED ISSUE #390: Autosave annotations
Annotations can be lost with a browser crash and should auto-save on creation.

jbeezley:
Fixes #390.

This pushes the drawn annotations to the server at most every 1 second.  If you reload the page without saving, the annotations will appear in the annotation panel with an autogenerated name containing your user name and a time stamp.  I suspect we will also want to allow one to edit saved annotations in a future PR.  This will at least prevent drawn annotations from being lost entirely.

### DIFF
--- a/plugin_tests/client/annotationSpec.js
+++ b/plugin_tests/client/annotationSpec.js
@@ -147,6 +147,11 @@ $(function () {
                 var annotation;
 
                 girderTest.waitForLoad();
+
+                // If the next rest request happens too quickly after saving the
+                // annotation, the database might not be synced.  Ref:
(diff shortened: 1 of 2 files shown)


manthey:
As a separate PR, we should pre-populate the name in the Save Annotation dialog with the name we are using for the auto-save.

jbeezley:
It looks like this created a race condition in the test.  I'll look into it next week.

jbeezley:
I'll go ahead and merge this.  I couldn't reproduce it locally.  The fail log indicated that the rest request getting the annotation occurred (correctly) after the auto save completed request completed, but before it was committed to the database.  I went ahead and added an extra 100ms delay just to help alleviate the problem.

PR #399, opened 2017-10-05
Use relative urls.

manthey:
We had a few absolute URLs, which break when in a proxy environment (such as on demo.kitware.com/histomicstk).


manthey:
There might be a better way to do this.  For instance, the thumbnail links in the recent annotation list include `api/v1`, so they should probably use `rest.getApiRoot()`.

PR #402, opened 2017-10-10
Switch to 2.x-maintenance branch of girder for testing and deployment

jbeezley:
I replaced all instances of `girder_version=master` that I found.  I didn't test it fully because I don't think I'm aware of all the deployment scenarios in use.  No real rush on this anyway.


manthey:
As part of PR #401, I removed one spot where we were misusing the girder version (passing it to the ctk-cli version).  That should probably be merged first or that changed moved to here.

PR #406, opened 2017-10-18
Use annotation model save/delete rather than custom restRequests

jbeezley:
This simplifies much of the auto saving logic by relying on more easily tested methods on the annotation model itself.  These changes are intended to make it easier to pass annotation models between the Annotation and Draw widgets for editing previously saved annotations.


manthey:
If I draw a point, save the annotation under a name, and then draw another point, the second annotation is auto-saved with the first annotation's name, not the default save name.

PR #408, opened 2017-10-20
Reduce the size of the generated docker files.

manthey:
This deletes a variety of caches as the docker files are built.  With current master, it reduces histomicstk_main from 3.72 GB to 2.25 GB and girder_worker from 1.61 GB to 1.27 GB.

Further gains are possible: shallow cloning git repos, deleting doc, man, and locale information, etc.  Some of these will have repercussions in using the docker container for development, however.


manthey:
@cdeepakroy These are the changes that resulted in the more efficient docker files.

PR #409, opened 2017-10-26
Install keras, tensorflow and check if accessible in Dockerfile

cdeepakroy:
This PR:

* Adds keras and tensorflow to requirements.txt. Note that only CPU version of tensorflow is in requirements.txt. If the user wants to take advantage of the GPU, then the GPU version (`tensorflow-gpu`), the user is responsible for uninstalling the CPU version and installing the GPU version explicitly after installing HistomicsTK. This is clarified in the documentation for installing HistomicsTK as a python toolkit.

* Creates a separate virtual environment called `venv-gpu` and installs `tensorflow-gpu` in the Dockerfile with HistomicsTK CLIs.

* Adds a new shell script called `server/docker-entrypoint.sh` that activates the `venv-gpu` environment only if the docker container is run with `nvidia-docker` and if `nvidia-smi` works. This is to ensure that the CLIs run on all machines (with or without GPU) irrespective of whether they are run with plain `docker run` or `nvidia-docker run`.

Note that the CLIs in the docker container will take advantage of the GPU only if run with `nvidia-docker` that is expected to be pre-installed on the machine.

### DIFF
--- a/server/docker-entrypoint.sh
+++ b/server/docker-entrypoint.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-if type nvidia-smi > /dev/null;
+if type nvidia-smi >/dev/null 2>/dev/null;
 then
     source /venv-gpu/bin/activate
 fi
 
-python /build/slicer_cli_web/server/cli_list_entrypoint.py
+python /build/slicer_cli_web/server/cli_list_entrypoint.py "$@"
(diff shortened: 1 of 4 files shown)


server/docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if type nvidia-smi > /dev/null;
```
manthey: Can you change this to
```
if type nvidia-smi >/dev/null 2>/dev/null;
```
to also suppress the `type: nvidia-smi: not found` error message?

Issue #411, opened 2017-11-01
 Group members can delete each others' annotations 


kheffah:
It seems that individual members of a group that is part of the same project can delete each other's annotations. If a user clicks on the 'x' symbol on an annotation document (for example, by error due to a lagging internet connection), that could potentially delete all 50 annotations in that document created by another user!

I think the ideal solution would be to be able to turn on and off the "God mode". There's probably an easy way to do this: only allowing the "owners" of a group to be able to delete other people's annotations, the rest being able to view/edit their annotations only.

jbeezley:
For the most part, this is resolved by #415.  It
* Adds a confirmation dialog when deleting annotations
* Removes the "delete annotation" button from regular (non-admin) users

It does not implement toggling god mode.  If that is still desired, feel free to add it to a new issue.

PR #415, opened 2017-11-02
Protect annotations from accidental deletion

jbeezley:
This does two things that we previously discussed:

1. Adds a confirmation dialog that appears when trying to delete an annotation.
2. Removes the "delete annotation" button from the panel for annotations not created by the current user.  For admins, the icon will always appear.


jbeezley:
Fixes #411

PR #416, opened 2017-11-03
Use extras requirements when installing large_image.

manthey:
large_image will soon require this to be installed in the expected way (see https://github.com/girder/large_image/pull/213).  This change allows installation through ansible (via vagrant, docker, etc.) before and after that PR gets merged.


manthey:
@brianhelba I'd like to get this merged either before or very soon after https://github.com/girder/large_image/pull/213, if you could please review it.

PR #417, opened 2017-11-06
Install large_image dependencies from setup.py

jbeezley:
The requirements file was removed in https://github.com/girder/large_image/pull/213, so just get the dependencies from the `setup.py`.


.travis.yml
```diff
@@ -117,7 +117,7 @@ install:
     - pip install --no-cache-dir numpy>=1.12.1
     - pip install --upgrade --no-cache-dir --force-reinstall --ignore-installed openslide-python Pillow
 
-    - pip install -U -r requirements.txt
+    - pip install -U -e .
```
manthey: We probably want this to be `pip install -U .[memcached,openslide]`.

PR #424, opened 2017-11-24
Refactor annotation panel for editable elements

jbeezley:
This commit allows the user to click on an entry in the annotation panel
and open it in the draw panel.  Going with this are several changes in
behavior that simplify the logic of annotation creation.

1. All server communiction for annotation objects is now handled by the
   annotation panel.  This includes the auto-save logic which was moved
   from the DrawPanel.  The annotation panel now simply listens for any
   change events on the annotation models and posts those changes to the
   server automatically.

2. The draw panel no longer has an special concept of a "new
   annotation".  Instead, all annotations in the draw panel are assumed
   to exist both on the server and in the draw panel.  Global annotation
   properties (i.e. those not related to individual elements) are now
   handled exclusively by the annotation panel.

3. There is special logic ensuring that annotations with large numbers
   of elements are not rendered by the draw panel.  This limit is
   currently set to 1000 elements.  In addition, any annotation that
   must be paged to the client cannot be edited.  In this case, a
   warning popup notifies the user.

4. Annotations are refetched every time they are toggled.  This could
   lead to performance issues.  It should be straightforward to
   implement some client side caching to handle this if necessary.


jbeezley:
Fixes #424.

jbeezley:
I broke my tests on the last commit, but I should have fixed the form submission in the create annotation modal.

manthey:
It would be nice if there was some indicator of which annotation is being edited in the Draw panel.  Perhaps we could have the name of the annotation in the draw panel title bar:
image
where the title elements now look like:
`<span class="s-panel-title"><span class="icon-pencil"></span> Draw<span class="s-panel-name"> Sample Annotation goes here with a really long name</span></span>`.  In my crude mockup, I needed to add `white-space: nowrap;text-overflow: ellipsis;` to the `.s-panel-group .s-panel-title-container .s-panel-title` selector and also
```
.s-panel-name::before {
    content: " - ";
}
s-panel-name {
    display: inline;
    overflow: hidden;
    white-space: nowrap;
    text-overflow: ellipsis;
    font-weight: normal;
}
```

manthey:
Sometimes when clicking on the annotations to change which annotation is active, the data in the draw panel is updated by the highlight on the annotations panel doesn't change.

Clicking on the active annotation reloads it, which seems wasteful (but it does correct it not being highlighted).

manthey:
You can hide the active annotation.  But then, when you draw another element, that is also invisible.  Perhaps drawing an element should automatically unhide the active annotation.

jbeezley:
PTAL I've made a number of improvements.  I think I fixed the issues with the occasional UI inconsistency in selecting annotations, but there may be other related problems that I missed.

manthey:
Showing the annotation is quicker then trying to load it for the draw panel, so it seems like we should be faster on refusing to edit large annotations, but we can improve that later if it becomes an issue.

PR #428, opened 2017-11-27
Add basic unit tests for all object-level feature extraction functions 

cdeepakroy:
This PR adds basic unit tests to check the sanity of the outputs returned by all object-level feature extraction functions in `histomicstk.features`


plugin_tests/feature_extraction_test.py
```diff
@@ -0,0 +1,182 @@
...
+
+        from histomicstk.features.compute_intensity_features import \
+            feature_list
+
+        fdata = htk_features.compute_intensity_features(
+            self.im_nuclei_seg_mask, self.im_nuclei_stain)
+
+        self.check_fdata_sanity(fdata, feature_list)
+
+        self.check_fdata_sanity(self.fdata_nuclei, feature_list,
```
slee172: Just wanted to make sure that this line seems to compare the columns of the entire features that starts with 'Nucleus.' with the columns of the intensity features.

cdeepakroy: Yes, this will check if each of the expected features are in the columns of `fdata_nuclei` that start with 'Nucleus.'.

PR #432, opened 2017-11-29
Fix anonymous annotation viewing.

manthey:
If the user isn't defined, a console error would be thrown and annotations were not visible.


manthey:
@law12019 This should fix the issue you had.  If it does, please review and approve.

PR #434, opened 2017-12-01
Add brief documentation on installing analyses.

manthey:


docs/installation.rst
```diff
@@ -51,6 +51,23 @@ There are several methods that can be used to install HistomicsTK.  Each of thes
 .. include:: ../ansible/README.rst
    :start-after: __methods
 
+Installing Analysis Images
```
cdeepakroy: Maybe rename to something along the lines of `Installing Docker Images containerizing Slicer CLIs for Analysis` to be clear.

docs/installation.rst
```diff
@@ -51,6 +51,23 @@ There are several methods that can be used to install HistomicsTK.  Each of thes
...
+page).
+
+Selecting the Web API will show a list of API endpoints for Girder.  In the
+``PUT`` ``/HistomicsTK/HistomicsTK/docker_image`` endpoint, you can add a list
+of Docker images, such as ``["dsarchive/histomicstk:latest"]``, and then click
+"Try it out".  This will pull the specified Docker image and make it available
+to the HistomicsTK interface.  You could also use a specific tag (version) from
+Docker instead of the ``latest`` tag.  Multiple versions and multiple images
+can be installed so that they can be compared against each other.
+
```
cdeepakroy: [Here](https://github.com/cdeepakroy/slicer_cli_web_plugin) is an example of how to package Analytics Slicer CLIs into a Docker image. We need more detailed documentation for this but pointing this could be a good start. I am guessing Item-tasks will provide that documentation in the future, and we can just reference it then.

PR #439, opened 2017-12-20
Add a setting to set and reset the default styles in the draw panel.

manthey:
This exposes the setting in JSON in the HistomicsTK plugin configuration page, plus adds buttons to the styles panel to reset to defaults or to set the defaults to the current values.

This also adds a user-readable settings endpoint (since otherwise settings are admin-only).

Fix coverage reporting based on recent girder changes.

Fix a bug when logging out and logging in had off behavior.   Specifically, after logging out a warning could be shown because annotations could not be loaded.  If, after logging in again, the same image was selected that had been used before log out, it would not be shown.

### DIFF
--- a/plugin_tests/client/histomicstkSpec.js
+++ b/plugin_tests/client/histomicstkSpec.js
@@ -12,6 +12,9 @@ girderTest.startApp();
 
 describe('Test the HistomicsTK plugin', function () {
     it('change the HistomicsTK settings', function () {
+        var styles = [{'lineWidth': 8, 'id': 'Sample Group'}];
+        var styleJSON = JSON.stringify(styles);
+
         girderTest.login('admin', 'Admin', 'Admin', 'password')();
         waitsFor(function () {
             return $('a.g-nav-link[g-target="admin"]').length > 0;
@@ -37,8 +40,6 @@ describe('Test the HistomicsTK plugin', function () {
             return $('#g-histomicstk-form input').length > 0;
         }, 'settings to be shown');
         girderTest.waitForLoad();
-        var styles = [{'lineWidth': 8, 'id': 'Sample Group'}];
-        var styleJSON = JSON.stringify(styles);
         runs(function () {
             $('#g-histomicstk-default-draw-styles').val(styleJSON);
             $('.g-histomicstk-buttons .btn-primary').click();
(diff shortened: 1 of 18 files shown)


web_client/views/body/ConfigView.js
```diff
@@ -0,0 +1,78 @@
...
+                value: this.$('#g-histomicstk-default-draw-styles').val()
+            }]);
+        },
+        'click #g-histomicstk-cancel': function (event) {
+            router.navigate('plugins', {trigger: true});
+        }
+    },
+    initialize: function () {
+        this.breadcrumb = new PluginConfigBreadcrumbWidget({
+            pluginName: 'Database assetstore',
```
jbeezley: HistomicsTK

plugin_tests/client/histomicstkSpec.js
```diff
@@ -0,0 +1,78 @@
...
+        runs(function () {
+            expect($('.g-plugin-config-link[g-route="plugins/HistomicsTK/config"]').length > 0);
+            $('.g-plugin-config-link[g-route="plugins/HistomicsTK/config"]').click();
+        });
+        waitsFor(function () {
+            return $('#g-histomicstk-form input').length > 0;
+        }, 'settings to be shown');
+        girderTest.waitForLoad();
+        var styles = [{'lineWidth': 8, 'id': 'Sample Group'}];
+        var styleJSON = JSON.stringify(styles);
```
jbeezley: It's probably best to keep definitions at the top of functions given how the individual steps run async.

web_client/views/body/ImageView.js
```diff
@@ -56,6 +56,9 @@ var ImageView = View.extend({
         this.listenTo(this.annotationSelector, 'h:deleteAnnotation', this._deleteAnnotation);
 
         this.listenTo(events, 's:widgetChanged:region', this.widgetRegion);
+        this.listenTo(events, 'g:login g:logout.success g:logout.error', () => {
+            this._openId = null;
```
jbeezley: This is an improvement over the current (broken) behavior.  There is still an issue with the state of the panels remaining after logout/login.  I think the panels should be reset.  We can save that for a future PR though.

PR #440, opened 2017-12-20
Highlight hovered annotations

jbeezley:
This adds two visual indicators for the user:

1. When an annotation element is hovered on the image, the corresponding
   annotation and element are highlighted in the panels.
2. When an annotation is hovered in the annotation panel or an element
   is hovered in the draw panel, the corresponding annotation or element
   is highlighted in the image.  This uses the `highlightAnnotation`
   method on the image viewer, which by default will make all other
   annotation elements more transparent.

Depends on https://github.com/girder/large_image/pull/236

### DIFF
--- a/web_client/panels/AnnotationSelector.js
+++ b/web_client/panels/AnnotationSelector.js
@@ -116,9 +116,7 @@ var AnnotationSelector = Panel.extend({
         var id = $(evt.currentTarget).parents('.h-annotation').data('id');
         var model = this.collection.get(id);
         model.set('displayed', !model.get('displayed'));
-        if (model.get('displayed')) {
-            model.set('highlight', true);
-        } else {
+        if (!model.get('displayed')) {
             model.unset('highlight');
         }
     },
--- a/web_client/panels/AnnotationSelector.js
+++ b/web_client/panels/AnnotationSelector.js
@@ -183,10 +183,15 @@ var AnnotationSelector = Panel.extend({
     editAnnotation(evt) {
         var id = $(evt.currentTarget).parents('.h-annotation').data('id');
         var model = this.collection.get(id);
+
+        // deselect the annotation if it is already selected
         if (this._activeAnnotation && model && this._activeAnnotation.id === model.id) {
-            model.set('displayed', true);
+            this._activeAnnotation = null;
+            this.trigger('h:editAnnotation', null);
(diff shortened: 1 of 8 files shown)


web_client/panels/DrawWidget.js
```diff
@@ -50,6 +53,16 @@ var DrawWidget = Panel.extend({
                 this._groups.get(this._style.id).save();
             }
         });
+        this.on('h:mouseover', (model) => {
```
manthey: The `mouseover` and `randomevent` handlers are too slow.  Specifically, the `this.render()` call rerenders the entire draw panel widget when we really just want to toggle a class on a single element.  Instead of `this._skipRedraw = true; this.render();`, `this.$('.h-element[data-id="' + model.id + '"]').addClass('h-highlight-element');` would functionally do the same thing (although duplicating some logic that is in the pug template), and would be vastly faster.  Similarly, to turn off the highlighting, the code could be the same except removing the class.

web_client/panels/AnnotationSelector.js
```diff
@@ -241,7 +255,10 @@ var AnnotationSelector = Panel.extend({
         );
     },
 
-    _saveAnnotation(annotation) {
+    _saveAnnotation(annotation, attribute, options) {
+        if (options && options.noSave) {
```
manthey: Are you using this option anywhere?

web_client/panels/DrawWidget.js
```diff
@@ -50,6 +53,14 @@ var DrawWidget = Panel.extend({
...
+            this._highlighted[model.id] = true;
+            this.$(`.h-element[data-id="${model.id}"]`).addClass('h-highlight-element');
+        });
+        this.on('h:randomevent', (model) => {
```
manthey: Why `h:randomevent` and not `h:mouseout`?  And, to that end, `h:mouseover` is fired on `mouseon` rather than `mouseover`.  Should its name reflect that?

jbeezley: I honestly don't know where that event name came from.

manthey:
Occasionally I get an odd highlight effect.  I click to view (not edit) a large annotation, then mouseout before it finishes loading.  The name of the annotation in the Annotations panel remains highlighted.

Issue #443, opened 2018-01-06
Disappearing annotations


kheffah:
After the incorporation of new features (editable individual annotations and highlighting annotations when mouse hovered over boundary), at least 4 users reported disappearing annotations (i.e. not saved). It did not happen with me, so I didn't get the chance to get the curl variant or other more useful metrics, but I will try.

kheffah:
If it's almost a necessity that the interface would be slower when interactive highlighting of annotations occurs, I suggest adding a checkbox "Interactive Mode", which toggles this on- and off-, if possible.

manthey:
Is there any more data on when they disappeared?  For instance, after initially drawing the annotation versus editing an existing annotation?

kheffah:
Here is the account from one of the users, I will ask others for more description:

_"while annotating the slide disappeared totally and then refreshed the page many times then it loaded and annotation was not saved that happened with one slide.. other one annotation disappeared completely exactly half of it.."_

kheffah:
Account from another user:

_"After editing more than 10 annotations in a group, some disappear and others remain."_

kheffah:
Another user:

_"I opened the website, then did annotations in three new folders, then I left the slide page open for some time without annotating. When I came back to do some more, I found that all the annotations in these three folders are gone. Although the old annotations remained the same. This was one time.
The other 2 times I don't remember exactly but it was something like that. I left the page open for some time then I found them gone. Also when I click refresh they don't appear.

N.B.: I'm using Mozilla Firefox web browser"_

jbeezley:
I can try to do some more testing here.  Do you have a database dump somewhere?  It would be interesting to see what the histories are on these... I.e. are the annotations appearing and then being overwritten with old data?

Just to make sure, people aren't editing the same annotations at the same time are they?

manthey:
I think I've just reproduced it.  It has to do with switching to another tab and then switching back after a while, I think.

kheffah:
@jbeezley I don't have a dump, unfortunately, but I don't think it would be very helpful anyways because people are working on new annotations, not editing old ones. Also, it's very unlikely that people are working on the same annotation because:
1- They are assigned different slides.
2- Even if they did try to edit the same annotation, they cannot because the permissions only allow the annotation owner to see the annotation document components and to edit it.

kheffah:
@manthey This makes sense, and seems to be consistent with some of the user reports.

manthey:
[...]

PR #445, opened 2018-01-08
Fix the active annotation when the event stream reconnects.

manthey:
When the event stream reconnects, we check to see if there are any changes to annotations.  If there is an active annotation, reattach to it by the reloaded annotation model.  Prior to this, the active annotation was no longer bound to the appropriate events, resulting in not auto-saving.

Also, guard against undefined models in mouse highlighting that can occur in this instance.

Remove the setup.cfg file and adjust the plugin.cmake file to track changes in Girder.  Flake8 has an issue that it uses the wrong setup.cfg.  I don't think we should be specifying a universal with with the cython extensions, and we weren't actually using bumpversion.


manthey:
Fixes #443.

PR #446, opened 2018-01-12
Change how the command yaml is structured.

manthey:
When a command in a yaml >- block, it is quoted differently than when it
is on the same line as the command: keyword.  It works with proper
quoting on one line, but fails for reasons I don't fully get in multiple
lines.  Switch to the one-line version to get the expected quoting
behavior.

Also, do not log ansible commands where a password is used.

### DIFF
--- a/ansible/roles/provision/tasks/main.yml
+++ b/ansible/roles/provision/tasks/main.yml
@@ -235,18 +235,18 @@
   tags: data_import
 
 - name: Install CLI
-  command: >-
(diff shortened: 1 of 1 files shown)


manthey:
I'm not sure what the best way to test this for review is -- I've generated a local docker image with the change and can show that a password like `pass" 'word` gets passed through properly (which it didn't before).

manthey:
This fix was confirmed in the field and verbally approved.

PR #450, opened 2018-01-30
Set a default location for output files

jbeezley:
This sets a default value for all `new-file` widget types.  The file
name is derived from the anaylsis name and input label.  The file is
placed in the user's `Private` directory. This relies on
https://github.com/girder/slicer_cli_web/pull/55 to get the correct extension.

This branch makes it possible to run the `Nuclei Detection` analysis on
the full slide image without changing any defaults on the control panel.

I didn't add tests for this because there isn't currently any client
side infrastructure for testing the analysis code paths.  Most of the
logic is contained in slicer_cli_web, which is tested.  To get coverage
for this and the rest of the control panel code, we should add a new
test case which would probably be more appropriate in a separate PR.

### DIFF
--- a/web_client/views/body/ImageView.js
+++ b/web_client/views/body/ImageView.js
@@ -2,8 +2,10 @@
 import _ from 'underscore';
 
 import { restRequest } from 'girder/rest';
+import { getCurrentUser } from 'girder/auth';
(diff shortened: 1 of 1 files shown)


manthey:
We definitely need to add tests in another PR.

PR #453, opened 2018-01-31
Add tests for the autofilling behavior of the analysis control panel

jbeezley:
This adds tests for #450, plus some infrastructure to extend the testing of the control panel.


plugin_tests/client/common.js
```diff
@@ -0,0 +1,133 @@
...
+            done = true;
+        }).catch(function (err) {
+            if (description) {
+                console.error(description);
+            }
+            throw err;
+        });
+        waitsFor(function () {
+            return done;
+        }, description || 'Promise to resovle');
```
manthey: `resovle` -> `resolve`

plugin_tests/web_client_test.py
```diff
@@ -12,9 +15,77 @@
...
+            }
+        }
+
+    @access.public
+    @describeRoute(
+        Description('Mock an analysis description route.')
+    )
+    def testAnalysisXml(self, params):
+        """Return the nuclei detection XML spec as a test case."""
+        xml_file = os.path.join(
```
manthey: There is no need for this outer `os.path.join`

plugin_tests/web_client_test.py
```diff
@@ -12,9 +15,77 @@
...
+            os.path.abspath(
+                os.path.join(
+                    os.path.dirname(__file__), 'test_files', 'NucleiDetection.xml'
```
manthey: Can we change this to `os.path.dirname(__file__), '..', 'server', 'NucleiDetection', 'NucleiDetection.xml'` and not duplicate the NucleiDetection.xml file?

PR #460, opened 2018-02-13
Show the image name and path on the header bar.

### LINKED ISSUE #457: Show current image name
Show the image name in the HistomicsTK interface.

Even better would be to show the image breadcrumbs.

manthey:
Adjust various styles to make sure that a narrow window or long image path still looks okay.

This doesn't reuse the Girder item breadcrumbs widget because too much needed to change.

Closes #457.


manthey:
For example:
image

PR #466, opened 2018-02-20
Allow filtering by annotation creator and name

jbeezley:
This adds two form elements to the "recently annotated images dialog" allowing the user to filter by annotation creator and/or image name.  This relies on https://github.com/girder/large_image/pull/259 to perform the query.  I could (but haven't yet) add some sort of paging to this dialog so you can go back through more than the last ten images.  The endpoint does the filtering in python so it will get slow if you try load too many pages.  We'll have to see if the performance is *good enough* in the real world.

test

### DIFF
--- a/web_client/dialogs/openAnnotatedImage.js
+++ b/web_client/dialogs/openAnnotatedImage.js
@@ -87,15 +87,30 @@ const OpenAnnotatedImage = View.extend({
             limit: 10
         };
         let items;
+        let changed = false;
 
-        this._creator = this.$('#h-annotation-creator').val();
-        if (this._creator) {
-            data.creatorId = this._creator;
+        const creator = this.$('#h-annotation-creator').val();
+        if (this._creator !== creator) {
+            this._creator = creator;
+            changed = true;
+
+            if (creator) {
+                data.creatorId = creator;
+            }
         }
 
-        this._imageName = (this.$('#h-image-name').val() || '').trim();
-        if (this._imageName) {
-            data.imageName = this._imageName;
+        const imageName = (this.$('#h-image-name').val() || '').trim();
+        if (this._imageName !== imageName) {
+            this._imageName = imageName;
+            changed = true;
+
+            if (imageName) {
+                data.imageName = this._imageName;
+            }
+        }
+
+        if (!changed) {
+            return $.Deferred().resolve(this.collection).promise();
         }
 
         return restRequest({
@@ -109,7 +124,7 @@ const OpenAnnotatedImage = View.extend({
             return $.when(...promises);
         }).then(() => {
             this.collection.reset(items);
-            return items;
+            return this.collection;
         });
     },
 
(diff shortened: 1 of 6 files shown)


manthey:
I probably should have caught this on https://github.com/girder/large_image/pull/259.  Currently, this doesn't work if you type in a string with capital letters.  "TCGA" won't match, but "tcga" will.  It would be better to fix this in large_image, but it could be changed here, too.

jbeezley:
I think I'll put in a PR in large_image to fix that.

web_client/dialogs/openAnnotatedImage.js
```diff
@@ -3,46 +3,104 @@ import _ from 'underscore';
...
+            limit: 10
+        };
+        let items;
+
+        this._creator = this.$('#h-annotation-creator').val();
+        if (this._creator) {
+            data.creatorId = this._creator;
+        }
+
+        this._imageName = this.$('#h-image-name').val();
```
manthey: We should probably `trim` the value to remove whitespace from the search.

web_client/dialogs/openAnnotatedImage.js
```diff
@@ -3,46 +3,104 @@ import _ from 'underscore';
...
+        if (this._imageName) {
+            data.imageName = this._imageName;
+        }
+
         return restRequest({
```
manthey: If `_imageName` and `_creator` are the same as the last search, it would be nice not to redo the search.

web_client/dialogs/openAnnotatedImage.js
```diff
@@ -3,46 +3,104 @@ import _ from 'underscore';
...
+        this._users.sortField = 'login';
+        this._users.pageLimit = 500;
+        this._usersIsFetched = false;
+        this._users.fetch().done(() => {
+            this._usersIsFetched = true;
+            this.render();
+        });
+
+        if (createDialog.debounceTimeout > 0) {
+            this._debouncedFetch = _.debounce(this.fetch, createDialog.debounceTimeout);
```
manthey: Debouncing by time isn't really the right way to debounce this call.  Rather, if we are in a pending fetch, we shouldn't issue another fetch until that one returns.  If the fetch is slower than the `debounceTimeout`, this could still saturate the server (and if it quicker, it won't be as responsive as it could be).

web_client/dialogs/openAnnotatedImage.js
```diff
@@ -1,57 +1,134 @@
...
+        this._nextQuery = {};
+
+        if (imageName) {
+            this._nextQuery.imageName = imageName;
+        }
+        if (creator) {
+            this._nextQuery.creatorId = creator;
+        }
+
+        this._fetchImages();
```
manthey: This will always call `_fetchImages` when there is an input change.  To prevent having multiple calls at once, you probably need to add a value like `this._inFetch = true;` just before the rest request that is cleared when the `done` function is called, and then `_fetchImages` would not get called here if `this._inFetch` was set.

jbeezley: That's what I was attempting with the `this._nextQuery`, but I see now my logic was flawed.  I'll make another attempt.

PR #467, opened 2018-02-22
Speedup nuclei detection on small ROIs

cdeepakroy:
This PR significantly speeds up the Nuclei Detection and Nuclei Feature Extraction CLIs on small ROIs and adds some profiling code.

When i profiled the old code on an ROI of size 1000 x 1000 in a slide, i got the following stats

```
histomicsk import time = 0.991516113281 seconds
large_image import time = 1.90734863281e-06 seconds
Total packages import time = 1.27036499977 seconds
Dask setup time = 0.432317018509 seconds
low-res foreground mask computation time = 41.9267101288 seconds
time to find tiles with significant foreground = 0.321593999863 seconds 
nuclei detection time = 4.45551609993 seconds
Total analysis time = 47.8499279022 seconds
```
The `low-res foreground mask computation time`  step that takes the most of the time, reads the slide at a low resolution (approx 2048 x 2048) and computes a binary mask representing the foreground in slide. This is then used to throw away tiles falling in the large amount of white space in the borders and speedup the analysis of the whole-slide.

However, this step is an overkill when analyzing small ROIs. Hence, i modified the code to skip this step, and now the code is significantly faster on small ROIs (< 10 sec on 1k x 1k ROI).

cc: @dgutman @cooperlab @manthey


server/NucleiDetection/NucleiDetection.py
```diff
@@ -143,12 +156,18 @@ def main(args):
...
+        if process_whole_image:
+
+            tile_fgnd_frac_list = htk_utils.compute_tile_foreground_fraction(
+                args.inputImageFile, im_fgnd_mask_lres, fgnd_seg_scale,
+                **it_kwargs
+            )
+
+        else:
+
+            tile_fgnd_frac_list = [1.0] * num_tiles
```
manthey: This should be `tile_fgnd_frac_list = np.full(num_tiles, 1.0)`, because we expect it to be a numpy array when we compute the nonzero count.

PR #473, opened 2018-03-02
Reset the ROI annotation when switching analyses

### LINKED ISSUE #469: Switching analysis doesn't reset the ROI
If you set an ROI and then switch analysis, the ROI is still shown, but the new analysis defaults to the whole image.  It should either use the set ROI (preferred) or clear the ROI so it is obvious it does not apply.

jbeezley:
Fixes #469


manthey:
If it isn't too much work, it might be nicer to keep the ROI if the new analysis has the appropriate input.

jbeezley:
I managed to make it work without too much trouble.

PR #474, opened 2018-03-05
Install bokeh to enable dask diagnostics web-interface

cdeepakroy:
This PR
* Adds `bokeh` to `requirements.txt` to enable [Dask diagnostics web interface](http://distributed.readthedocs.io/en/latest/web.html) that is very useful in evaluating the parallelization throughput of our algorithms.
* Bumps the versions of dask packages: `dask, distributed`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN pip install --upgrade --ignore-installed pip setuptools && \
     # Try twice; conda sometimes causes pip to fail the first time, but if it
     # fails twice then there is a real issue.
     pip install -r requirements.txt && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -23,6 +23,7 @@ RUN pip install --upgrade --ignore-installed pip setuptools && \
     # Try twice; conda sometimes causes pip to fail the first time, but if it
     # fails twice then there is a real issue.
     pip install -r requirements.txt && \
+    pip install -r bokeh>=0.12.14 && \
```
manthey: You don't want the `-r` on this line.

PR #478, opened 2018-03-06
Add a slider to set displayed opacity in the annotation panel

### LINKED ISSUE #458: Add slidebar for filled annotation transparency
My group is using HTK to review mask overlays, so we are using the "fillColor" of annotations. However, the filled annotations are too opaque to show underlying structures, so we need some sort of a sliding bar to control displayed "alpha" (transparency) of filled annotations. The temporary "hacky" workaround we are using is by hovering the mouse over the annotation name on the right, which decreases the opacity of all annotations except the one the mouse is overlayed on. (see attached screenshots for pre- and post- mouse hovering)
screenshot from 2018-02-11 19-02-25
screenshot from 2018-02-11 19-02-42

jbeezley:
This adds a simple slider in the annotations panel to change the global opacity on the annotation layer itself.  The UI doesn't look very good to me (something about the color/position/spacing seems off), so I would welcome any suggestions to improve it.

Depends on https://github.com/girder/large_image/pull/264
Fixes #458
test

### DIFF
--- a/web_client/panels/AnnotationSelector.js
+++ b/web_client/panels/AnnotationSelector.js
@@ -316,7 +316,7 @@ var AnnotationSelector = Panel.extend({
     _changeGlobalOpacity() {
         this._opacity = this.$('#h-annotation-opacity').val();
         this.$('.h-annotation-opacity-container')
-            .attr('title', `Annotation opacity ${this._opacity * 100}%`);
+            .attr('title', `Annotation opacity ${(this._opacity * 100).toFixed()}%`);
         this.trigger('h:annotationOpacity', this._opacity);
     }
 });
--- a/web_client/panels/AnnotationSelector.js
+++ b/web_client/panels/AnnotationSelector.js
@@ -315,6 +315,8 @@ var AnnotationSelector = Panel.extend({
 
     _changeGlobalOpacity() {
         this._opacity = this.$('#h-annotation-opacity').val();
+        this.$('.h-annotation-opacity-container')
+            .attr('title', `Annotation opacity ${this._opacity * 100}%`);
         this.trigger('h:annotationOpacity', this._opacity);
     }
 });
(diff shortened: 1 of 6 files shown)


cooperlab:
Looks great.

web_client/panels/AnnotationSelector.js
```diff
@@ -31,7 +31,8 @@ var AnnotationSelector = Panel.extend({
         'click .h-hide-all-annotations': 'hideAllAnnotations',
         'mouseenter .h-annotation': '_highlightAnnotation',
         'mouseleave .h-annotation': '_unhighlightAnnotation',
-        'change #h-toggle-labels': 'toggleLabels'
+        'change #h-toggle-labels': 'toggleLabels',
+        'change #h-annotation-opacity': '_changeGlobalOpacity'
```
manthey: Updating the layer opacity is fast enough, let's do this as the slider is dragged (add `'input #h-annotation-opacity': '_changeGlobalOpacity'`).

web_client/panels/AnnotationSelector.js
```diff
@@ -307,6 +311,11 @@ var AnnotationSelector = Panel.extend({
 
     _changeAnnotationHighlight(model) {
         this.$(`.h-annotation[data-id="${model.id}"]`).toggleClass('h-highlight-annotation', model.get('highlighted'));
+    },
+
+    _changeGlobalOpacity() {
+        this._opacity = this.$('#h-annotation-opacity').val();
```
manthey: I'd like to see the opacity value somewhere.  A simple way would be to add a title attribute: `this.$('#h-annotation-opacity').attr('title', (100.0 * this._opacity).toFixed(0) + '%');`.  This would only show when hovering, not dragging, so there might be a better way to do this.

web_client/templates/panels/annotationSelector.pug
```diff
@@ -11,7 +11,7 @@ block content
     .btn-group.btn-group-sm(role='group')
       button.btn.btn-default.h-hide-all-annotations(type='button', title='Hide all annotations')
         | #[span.icon-eye-off]
-    span(title='Annotation opacity')
+    span.h-annotation-opacity-container(title=`Annotation opacity ${opacity * 100}%`)
```
manthey: Can you use `toFixed()`?  I see values like `Annotation opacity 7.000000000000001%`

jbeezley: That's odd, the slider uses `step="0.01"`.  I wasn't seeing that behavior, but I'll add it to be sure.

kheffah:
Superb!! Thank you!

PR #480, opened 2018-03-07
Dask distributed doesn't work with tornado 5.0.

manthey:
Pin the tornado version for now.


manthey:
See https://github.com/dask/distributed/issues/1725.

PR #483, opened 2018-03-12
Add a checkbox to toggle annotation interactivity

### LINKED ISSUE #459: Add checkbox to toggle interactive mode
Since the new changes were pulled, specifically the highlighting of individual annotations when mouse is hovered over boundary and vice versa, panning a heavily-annotated slide became slower. This is especially evident with extensive annotations, where any slight mouse movement inevitably crosses some boundary, even if the original intention is just to pan the slide. I suggest adding a checkbox for interactive_mode, where default is True and can be turned off by user.

jbeezley:
This add a new checkbox similar to the "enable labels" control to turn on or off the annotation highlighting behavior on mouse pointer hover.  There is also a small fix to the opacity slider, which I found would sometimes wrap below the show/hide buttons.

Fixes #459

### DIFF
--- a/web_client/stylesheets/panels/annotationSelector.styl
+++ b/web_client/stylesheets/panels/annotationSelector.styl
@@ -9,7 +9,7 @@
     background none
 
   input
-    width 210px
+    width 195px
     float right
     margin-left 10px
     margin-top 2px
--- a/web_client/stylesheets/panels/annotationSelector.styl
+++ b/web_client/stylesheets/panels/annotationSelector.styl
@@ -40,8 +40,18 @@
 .h-annotation.h-highlight-annotation
   background-color #ccc
 
-.h-create-annotation
-  float right
+.h-annotation-toggle
+  margin 0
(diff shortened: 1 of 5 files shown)


web_client/stylesheets/panels/annotationSelector.styl
```diff
@@ -9,7 +9,7 @@
     background none
 
   input
-    width 210px
+    width 200px
```
manthey: In the current version of Chrome I'm using, 200px still wraps (but 199px doesn't).

manthey:
If I have interactive mode on, hover over an annotation that overlaps the control panel, and the turn off interactions, the annotation in the list never de-highlights.

web_client/stylesheets/panels/annotationSelector.styl
jbeezley: I changed it to 195.  I'm sure there is a better way to do this in css...

PR #486, opened 2018-03-22
Allow setting number of dask workers and threads per worker and resolve memory overload issue

cdeepakroy:
This PR:
* Adds two parameters `num_workers` and `num_threads_per_worker` to all CLIs using Dask that will allow the user to set the Number of dask workers and number of threads per each dask worker, respectively, while setting a dask cluster internally (i.e. when no scheduler address is passed).
* When `num_workers` is negative (e.g. -1) then number of dask workers is set to number of physical cores (obtained using `psutils.cpu_count(logical=False)`) minus the number of workers specified. For example if `num_workers = -1`  then number of dask workers is set to `psutils.cpu_count(logical=False) - 1`. 
* Resolves memory overload issue by broadcasting `im_fgnd_mask_lres` to all workers before starting the dask computation in `compute_tile_foreground_fraction`.  If we don't do this, dask will make a copy of it and pickle it for each call to the dask.delayed function it is being passed to. 
* Lowers the `analysis_tile_size` parameters in CLIs using Dask to `1024`. The earlier setting of `4096` freezes my machine. With `1024` it runs smoothly and was able finish the NucleiDetection CLI on a `32k x 38k` pixels slide in ~7 minutes without freezing the machine. My machine has Dual Intel Xeon E5-2609 Processor with 8 cores each, 32 GB of RAM, and Ubuntu 16.04 LTS.
* Also lowered the `min_radius` to 6 for more accurate results. I will make another PR with better defaults after rigorous testing on some benchmark datasets.


histomicstk/utils/sample_pixels.py
```diff
@@ -86,14 +87,19 @@ def sample_pixels(slide_path, sample_fraction=None, magnification=None,
     iter_args = dict(scale=dict(magnification=magnification),
                      format=large_image.tilesource.TILE_FORMAT_NUMPY)
 
-    im_fgnd_mask_lres = dask.delayed(im_fgnd_mask_lres)
+    c = dask.distributed.get_client()
```
cdeepakroy: @manthey This throws an exception if no client has been setup!

Should i guard it with a try catch block so it scatters only if client has been setup?

Or would we want it to throw an exception if no client has been setup

server/cli_common/utils.py
```diff
@@ -215,14 +213,58 @@ def create_dask_client(args):
       empty string to start one locally
```
cdeepakroy: @manthey Can you update this doc string with a brief description of the semantics of the values that can be passed for scheduler address?

Also since we are broadening the types of values it can take, maybe we should rename is to `scheduler` instead of `scheduler_address`. And change it accordingly in the XML specs of the CLIs that invoke this.

server/cli_common/utils.py
```diff
@@ -215,14 +213,58 @@ def create_dask_client(args):
...
 
     """
+    import dask
     scheduler_address = args.scheduler_address
 
+    if scheduler_address in ('thread', 'threaded', 'threads'):
```
cdeepakroy: @manthey I suggest using just one value instead of three. Maybe `multithreading` and `multiprocessing`

manthey:
@cdeepakroy I think we should merge this as it stands, and address additional coverage in another PR.

PR #490, opened 2018-03-28
Refactor Dockerfile to use less memory.

manthey:
Before this refactor, the docker build process had a peak memory use of between 1.5 and 1.75 Gb.  Afterwards, the build process peaks at less than 0.5 Gb.  Dockerhub's build is limited in memory, so this should allow it to build there.


manthey:
@cdeepakroy This should allow docker hub to build our image again.

PR #491, opened 2018-03-29
Release the GIL during cython code.

manthey:
This allows better multithreading performance.


manthey:
@cdeepakroy You'll see the biggest different with this if you run a dask job with one worker and lots of threads.

PR #493, opened 2018-04-03
Open the job info page when submitting an analysis

### LINKED ISSUE #485: Add submit notification to Analysis Runner
Would be very helpful if after defining a job spec, that when I hit submit it notifies me it's doing something..

It could either do a tooltip type notification, or perhaps even better it just "opens" the job window so you can now see the job is running..

image

image

jbeezley:
The only problem with this is that opening the new window occurs asynchronously, so it is blocked by popup blockers.  At least in Chrome, you do get a notification the pop up was blocked and are given the option to disable it for the current page.

Fixes #485

### DIFF
--- a/plugin_tests/client/analysisSpec.js
+++ b/plugin_tests/client/analysisSpec.js
@@ -112,6 +112,25 @@ $(function () {
                 regionValue = $('#analysis_roi').val();
             });
         });
+        it('submit the job', function () {
+            var open = window.open;
+            var args;
+            window.open = function () {
+                args = arguments;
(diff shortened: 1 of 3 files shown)


manthey:
This works, but I would prefer it not to open another browser tab.

Based on the wording and illustrations of issue #485, I think it would be better to just expand the jobs panel.

manthey:
I like this much better.  I'd remove the word "successfully" from the message ("Analysis job submitted successfully." would be "Analysis job submitted.").

PR #496, opened 2018-04-09
Reload annotation list after switching users.

manthey:
After switching users and loading the same image, the annotation list was not refetched.  This meant that the first user's permissions were used to show the list of available annotations.


manthey:
This is part of the issue that @kheffah was seeing.

PR #498, opened 2018-04-13
Add annotation groups to the AnnotationSelector panel

jbeezley:
This groups the annotation list in the AnnotationSelector panel by
"group" name.  In this context, the group the annotation belongs to is a
list of one or more styles applied to at least one element.  In
addition, if an element in an annotation has no style name, then it is
listed as in the "null" group which gets displayed into a single group
named "Other".

I suspect we will want to add more "group" level actions like "download all"
or "show all".  This should be easy to do with these changes in place.

Depends on https://github.com/girder/large_image/pull/276

### DIFF
--- a/plugin_tests/client/annotationSpec.js
+++ b/plugin_tests/client/annotationSpec.js
@@ -328,6 +328,7 @@ $(function () {
 
             it('check that the drawing type persists when switching annotatations', function () {
                 runs(function () {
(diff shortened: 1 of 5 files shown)


manthey:
This works fine.  It did allow me to see a bug: if you switch to a non-default group, draw an annotation, and then switch back to the default group, the group value is still set as a non-default value.

PR #500, opened 2018-04-18
Update ansible and travis for pip 10 changes.

manthey:
We use some of the pip internals for cache management and ensuring we update the current versions of modules that depend on locally built libraries.  Unfortunately, pip doesn't have any cache management exposed publicly, which means either hardcoding paths or using now internal functions.

### DIFF
--- a/ansible/roles/openslide/tasks/main.yml
+++ b/ansible/roles/openslide/tasks/main.yml
@@ -225,18 +225,27 @@
 - name: Purge pip cache
   command: |
     python -c '
-    import pip.locations, shutil
-    shutil.rmtree(pip.locations.USER_CACHE_DIR, True)
+    try:
+    ''  import pip.locations as locations
+    except ImportError:
+    ''  import pip._internal.locations as locations
+    import shutil
+    shutil.rmtree(locations.USER_CACHE_DIR, True)
     '
   become: true
 
 - name: Reinstall python modules that depend on OpenJPEG and libtiff if they are installed.
   command: |
     python -c '
-    import pip
+    try:
+    ''  import pip.commands as commands
+    ''  from pip import main
+    except ImportError:
+    ''  import pip._internal.commands as commands
+    ''  from pip._internal import main
     packages = ["Pillow", "libtiff"]
-    for package in pip.commands.show.search_packages_info(packages):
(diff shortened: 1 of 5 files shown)


ansible/roles/openslide/tasks/main.yml
```diff
@@ -225,18 +225,27 @@
...
+    ''  import pip.commands as commands
+    ''  from pip import main
+    except ImportError:
+    ''  import pip._internal.commands as commands
+    ''  from pip._internal import main
     packages = ["Pillow", "libtiff"]
-    for package in pip.commands.show.search_packages_info(packages):
-    ''  pip.main(["install", "--force-reinstall", "--ignore-installed",
+    for package in commands.show.search_packages_info(packages):
+    ''  main(["install", "--force-reinstall", "--ignore-installed",
```
jbeezley: Can we just make this `python -m pip install ...`?

manthey: I'm using `pip` to get the exact package versions (`pip freeze` and `grep` would work) and reinstalling those versions.   From within python, we could `os.system(['pip', 'install', ...)`.  I'm not sure that is any less hacky.

manthey: `for pkg in `pip freeze | egrep '^(libtiff|Pillow)='`; do pip install --force-reinstall --ignore-installed "$pkg"; done` will do the same job.

jbeezley: I think that would be better.  There is also `pip list --format json`, but I think the freeze format is considered part of the API surface.  To solve the cache issue, can you use the `--no-cache-dir` or `--cache-dir <dir>` flags?

manthey: I want to delete the cache prior to reinstalling Pillow or libtiff, since otherwise some future command could end up pulling the wrong version from the cache.  This reinstall will add the correct versions with the local libraries to the cache (but not if we specify `--no-cache-dir`).  In the past, I found that if some other step (from Girder's ansible role, for instance) specifies a version of Pillow, that we'd end up with the version in the cache directory.

Deleting `~/.cache/pip` does the job for this version of pip.

PR #501, opened 2018-04-23
Use boundscheck(False) in the cython code.

manthey:
When we build the cython code, it advises to use `boundscheck(False)` to speed things up.


cooperlab:
Curious how much difference this will make.

manthey:
In one WSI sample, it reduced total processing time by about 1.8% (so not a lot, but measurable).

cooperlab:
That's an easy 2% though - good catch on this.

PR #506, opened 2018-06-25
Use girder mount to work with non-filesystem assetstore files.

manthey:
Support and start FUSE as part of the main HistomicsTK container.


manthey:
@dgutman With this PR (and building the appropriate docker), things will work "out of the box" with S3 assetstores.

dgutman:
Thanks!  I'll need some info when I test this later in the week..

manthey:
If we build and push docker images with it, you should just be able to run them and have it work.  I can push new docker images with it to dockerhub, or build them locally and transfer them to somewhere you can grab them.

PR #511, opened 2018-07-10
Fix output file specification for Positive Pixel Count.

manthey:
This also adds some default values.

### DIFF
--- a/server/PositivePixelCount/PositivePixelCount.xml
+++ b/server/PositivePixelCount/PositivePixelCount.xml
@@ -23,7 +23,7 @@
       <label>Hue Value</label>
       <index>1</index>
       <description>Center of the hue range in HSI space for the positive color, in the range [0, 1]</description>
-      <default>0.5</default>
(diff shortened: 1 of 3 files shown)


manthey:
@cdeepakroy I picked some values for defaults, but please change them if you think others would be better.

server/PositivePixelCount/PositivePixelCount.xml
```diff
@@ -67,8 +74,9 @@
       <description>left,top,width,height of the region of interest.  All -1 means the whole image is used.</description>
       <default>-1,-1,-1,-1</default>
     </region>
-    <image>
-      <longflag>outputLabelImage</longflag>
+    <image fileExtensions='.png'>
+      <name>outputLabelImage</name>
+      <index>1</index>
```
cdeepakroy: This will make it a mandatory output parameter. When it is run on the entire slide we cant produce this output image as there is no support for WSI writing in large image.

PR #513, opened 2018-07-13
Bump timeout cutoff on waiting for girder to report version

cdeepakroy:
The timeout cutoff for [Waiting for girder to report version](https://github.com/DigitalSlideArchive/HistomicsTK/blob/9dd96013780451f21682db29a0ed562bf43d0452/ansible/deploy_docker.py#L750) is currently set to 300 seconds. And this step takes ~600 seconds on my machine.This bumps it to 3600 to allow enough time to finish on most machines.

### DIFF
--- a/ansible/deploy_docker.py
+++ b/ansible/deploy_docker.py
@@ -747,7 +747,7 @@ def tag_with_version(key, version=None, **kwargs):
     return image
 
 
-def wait_for_girder(client, ctn, maxWait=300):
+def wait_for_girder(client, ctn, maxWait=3600):
     """
     Wait for Girder in a specific container to respond with its current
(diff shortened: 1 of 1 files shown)


ansible/deploy_docker.py
```diff
@@ -779,7 +779,9 @@ def wait_for_girder(client, ctn, maxWait=300):
         sys.stdout.flush()
     if not output:
         raise Exception('Girder never responded')
-    sys.stdout.write(' %s\n' % output['apiVersion'])
+    else:
```
manthey: This doesn't have to be in an `else:` block, since it will only be reached if the exception wasn't raised.

PR #514, opened 2018-07-13
Add a context menu to delete annotation elements

jbeezley:
This could be extended to do things like "shift-click to select multiple elements", but for now it will just add a right-click context menu to delete one annotation element.

In the case where the mouse is hovering over more than one annotation element, only one is acted on.  There is a visual highlight that indicates which element is going to be deleted.  This could make the feature difficult to use if many annotation elements are on top of each other, but I don't think that's a common enough occurrence to worry about for the moment.

### DIFF
--- a/web_client/views/body/ImageView.js
+++ b/web_client/views/body/ImageView.js
@@ -622,14 +622,18 @@ var ImageView = View.extend({
             return;
         }
 
-        this._editAnnotation(annotation);
-        const menu = this.$('#h-annotation-context-menu');
-        const position = evt.mouse.page;
-        menu.removeClass('hidden');
-        menu.css({ left: position.x, top: position.y });
-        this.popover.collection.reset();
-        this._contextMenuActive = true;
-        this.contextMenu.setHovered(element, annotation);
+        // Defer the context menu action into the next animation frame
+        // to work around a problem with preventDefault on Windows
+        window.setTimeout(() => {
+            this._editAnnotation(annotation);
+            const menu = this.$('#h-annotation-context-menu');
+            const position = evt.mouse.page;
+            menu.removeClass('hidden');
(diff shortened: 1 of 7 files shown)


jbeezley:
Additionally, I thought about adding something like a "shift-delete" to delete the hovered annotation, but I couldn't decide how it should work especially considering the ambiguity about which is activated.

cooperlab:
Thanks @jbeezley . We'll use this version for a while and think about how we might improve things in the future.

web_client/views/body/ImageView.js
```diff
@@ -571,6 +600,30 @@ var ImageView = View.extend({
...
     _setAnnotationOpacity(opacity) {
         this.viewerWidget.setGlobalAnnotationOpacity(opacity);
+    },
+
+    _rightClickElement(element, annotation, evt) {
+        if (annotation.get('_accessLevel') < AccessType.WRITE) {
+            return;
+        }
+
+        const menu = this.$('#h-annotation-context-menu');
```
manthey: The browser context menu hides the custom context menu for me.  Don't we need to prevent the event default to avoid this?  If so, geojs probably needs to include the original event in its event.

jbeezley: That's strange... I thought geojs was calling prevent default.

jbeezley: @manthey I figured out that I wasn't getting the menu on my laptop because I wasn't logged in.  I still can't reproduce the issue of the default context menu appearing.  I've tried on several browsers on Linux and MacOS.

manthey:
@jbeezley Your latest change works for me (though the test are currently failing).

jbeezley:
Okay, I'll get the test fixed shortly.

jbeezley:
Tests are now passing.

PR #518, opened 2018-08-02
Add more options to the cli_test.py script.

manthey:
This allows overriding default processing parameters from the command line.  It also improves reporting stability by ensuring that it waits until the generated annotation is processed.


manthey:
@cdeepakroy I had made these changes to make it easier to test dask options from the command line.

PR #522, opened 2018-08-06
Allowing zooming more.

### LINKED ISSUE #520: Need to have zoom beyond 20x
For nucleus annotation, it is often needed to zoom to very high magnifications to be able to make a good histologic judgement of what class it belongs to. It's OK if the image gets pixelated a bit, but some kind of "hyper-zoom" functionality would be preferred.

manthey:
Before, you could only zoom in until one image pixel filled one screen pixel.  This adds an internal parameter `_increaseZoom2x` that can be set to allow zooming in more.  By default it is now 1, allowing one image pixel to fill 2^1 x 2^1 screen pixels.  The zoom widget shows higher zoom levels by using smaller buttons to indicate that these are "extra zoomed".

This also fixes two minor issues: when switching to an image that had a different maximum zoom level, the zoom widget was not always updated. If an image had no magnification data the "Fit" button would be shown twice.

This resolves #520.

### DIFF
--- a/web_client/panels/ZoomWidget.js
+++ b/web_client/panels/ZoomWidget.js
@@ -66,9 +66,9 @@ var ZoomWidget = Panel.extend({
         max = Math.log2(this._maxMag) + 0.01;
 
         // get a list of discrete values to show as buttons
-        buttons = _.filter([1, 2.5, 5, 10, 20, 40], (v) => v <= this._maxMag);
-        buttons = _.last(buttons, 5);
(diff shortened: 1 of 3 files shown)


jbeezley:
The smaller buttons look a little odd to me.  I might try to distinguish them with a subtle color or border difference; however, I think this is fine for now.

manthey:
I'm open to style changes.  We can do them in another PR.

PR #524, opened 2018-08-09
[WIP] Add the ability to change style groups inside the context menu

jbeezley:
Before I spend time working on testing, I thought I would push this up to get feedback.  The context menu now has buttons to remove the current group or change to another existing group.  It will list up to ten existing groups sorted alphabetically.

test

### DIFF
--- a/web_client/views/popover/AnnotationContextMenu.js
+++ b/web_client/views/popover/AnnotationContextMenu.js
@@ -1,3 +1,6 @@
+import _ from 'underscore';
+
+import StyleCollection from '../../collections/StyleCollection';
 import View from '../View';
 
 import template from '../../templates/popover/annotationContextMenu.pug';
@@ -5,13 +8,20 @@ import '../../stylesheets/popover/annotationContextMenu.styl';
 
 const AnnotationContextMenu = View.extend({
     events: {
-        'click .h-remove-elements': '_removeElements'
+        'click .h-remove-elements': '_removeElements',
+        'click .h-set-group': '_setGroup',
+        'click .h-remove-group': '_removeGroup'
     },
-    initialize() {
+    initialize(settings) {
         this.reset();
     },
     render() {
-        this.$el.html(template());
+        const hovered = this._hovered || {};
+        const currentGroup = (hovered.element || {}).group;
+        this.$el.html(template({
+            groups: this._getAnnotationGroups(),
+            currentGroup
+        }));
         return this;
     },
     reset() {
@@ -30,6 +40,7 @@ const AnnotationContextMenu = View.extend({
         this.reset();
         this._hovered = { element, annotation };
         this.parentView.trigger('h:highlightAnnotation', annotation.id, element.id);
+        this.render();
     },
     _removeElements(evt) {
         evt.preventDefault();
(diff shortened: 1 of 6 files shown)


cooperlab:
Looks great. @kheffah please give this a test run with your annotations.

kheffah:
@jbeezley @cooperlab Thank you, I will test it out. May I ask what triggers the context menu? Left clicking in interactive mode, or something else? Early feedback before doing testing: it would be really preferable if the existing groups are listed in order of number of existing elements rather than alphabetically. This way if there are many groups (> 10) the ones that pop up are the most relevant ones rather than the rare groups.

kheffah:
@jbeezley first of all, thank you for your help with this. 

I think the latest docker only includes the "delete element" option.

image

Other thoughts about the context menu:
1- It is slow to show, it seems to be fetching something from the database when I right click?
2- It shows up even when I click outside an annotation.

But again, this is using the current docker version which does not chow the same context menu that you show a screenshot of here, so they may have already been addressed.

manthey:
The latest docker only has what has been merged to master.  To use a non-merged branch, you'll have to go into the docker and merge the branch with the code currently deployed there, then ask Girder to rebuild the web client and restart (there is a button to do so at the top of the plugins page).

kheffah:
@manthey Thank you! That makes sense.

jbeezley:
> 1- It is slow to show, it seems to be fetching something from the database when I right click?

No, but it is waiting for the next animation frame.  Ordinarily that shouldn't be noticeable, but it's possible the animation frame is being delayed--maybe by the code that highlights the chosen annotation.  In testing, I wasn't displaying many annotations, so it's likely I didn't notice.  I'll try it with a larger annotation dataset to see if I can improve the delay.

> 2- It shows up even when I click outside an annotation.

That shouldn't be happening.  Is it always reproducible or only occasionally?  Perhaps you could send me an annotation dump, and I can explore it.

jbeezley:
> May I ask what triggers the context menu?  Left clicking in interactive mode, or something else?

Right-click, but I assume you already figured that out.

> Early feedback before doing testing: it would be really preferable if the existing groups are listed in order of number of existing elements rather than alphabetically. This way if there are many groups (> 10) the ones that pop up are the most relevant ones rather than the rare groups.

I limited the number out of practicality, the context menu will likely overflow the bottom of the screen before you get that many anyway.  We might be able to do a scrollbar or something else to show them all.  In any case, sorting by most used seems like a good idea.

jbeezley:
@kheffah I made some changes that should improve the performance.  When you get a chance, let me know if it helped.

web_client/views/popover/AnnotationContextMenu.js
```diff
@@ -39,6 +53,53 @@ const AnnotationContextMenu = View.extend({
...
+        evt.preventDefault();
+        evt.stopPropagation();
+
+        const { annotation, element } = this._hovered;
+        this.trigger('h:edit', annotation.elements().get(element));
+    },
+    _setStyleDefinition(group) {
+        const styles = new StyleCollection();
+        return styles.fetch().done(() => {
+            const style = styles.get({ id: group || 'default' });
```
manthey: If I have an annotation with a style group that no longer exists, the subsequent line with `style.toJSON()` with fail.  Perhaps this should be changed to `const style = styles.get({id: group}) || styles({id: default})`

PR #525, opened 2018-08-09
Remove colinear points from annotations.

manthey:
When we detect nuclei and generate a polygon of the perimeter, this removes colinear points that do not alter the shape of the polygon.  In a simple test of around 325,000 nuclei, this reduces the total number of vertices from 24,094,710 to 13,954,237.

This also discarded around 1,400 degenerate polygons that had no area. The bulk of these consisted of three points; for example: [18437.0, 11894.0, 0.0], [18438.0, 11895.0, 0.0], [18437.0, 11894.0, 0.0].


manthey:
This works, but is a WIP until the tests are updated.

histomicstk/segmentation/label/trace_object_boundaries.py
```diff
@@ -117,7 +123,53 @@ def trace_object_boundaries(im_label,
...
+
+    while testpos < len(px):
+
+        # get coords of next triplet of points to test
+        if testpos == len(px) - 1:
+            if not len(keep):
+                break
+            nextpos = keep[0]
+        else:
+            nextpos = testpos + 1
```
cdeepakroy: @manthey I added your fix here

histomicstk/segmentation/label/trace_object_boundaries.py
```diff
@@ -117,7 +123,54 @@ def trace_object_boundaries(im_label,
                 im_label, dtype=np.int), conn, x_start, y_start, max_length
         )
 
+        bx, by = _remove_thin_colinear_spurs(bx, by, eps_colinear_area)
+
         X.append(bx)
```
cdeepakroy: Should we skip adding it to the list here if `len(bx) == 0`? 

Or should we let the user of this function handle it?

manthey: I think we should skip adding it here.  An object with no boundary is no object...

cdeepakroy: Alright, I dropped them!

histomicstk/segmentation/label/trace_object_boundaries.py
```diff
@@ -117,7 +124,55 @@ def trace_object_boundaries(im_label,
...
+            nextpos = testpos + 1
+
+        ind = [anchor, testpos, nextpos]
+        x1, x2, x3 = px[ind]
+        y1, y2, y3 = py[ind]
+
+        # compute area of triangle formed by triplet
+        area = np.linalg.det(
+            np.array([[x1, x2, x3], [y1, y2, y3], [1, 1, 1]])
+        )
```
manthey: This is twice the triangle's area, so we should either divide it by 2 or rename it and compare it to twice the limiting area.

histomicstk/segmentation/label/trace_object_boundaries.py
```diff
@@ -6,7 +6,8 @@
 
 def trace_object_boundaries(im_label,
                             conn=4, trace_all=False,
-                            x_start=None, y_start=None, max_length=None):
+                            x_start=None, y_start=None,
+                            max_length=None, eps_colinear_area=1.0):
```
manthey: If we set the default colinear area to a small value, then this works fine.  I recommend something like `0.01` or `0.1`.

PR #526, opened 2018-08-16
Jupyter notebook for reading wsi data using large_image

cdeepakroy:
This PR adds an example jupyter notebook show how to use `large_image` to read different kings of information from whole-slide images.


cdeepakroy:
@cooperlab and @manthey CI is now happy. Can you please review when you can?

[Here](https://github.com/DigitalSlideArchive/HistomicsTK/blob/add-large-image-notebook/docs/examples/wsi-io-using-large-image.ipynb) is the github rendering of the notebook.

cooperlab:
This is great.

Here are some comments to consider:
-omit help() - these generate very long blocks of text to read
-under getMetadata, specify that the 'tile' here refers to the tiling within the image file
-switch order of examples under getRegion - low-res whole slide image read should be last
-Showing convertRegionScale is probably enough. Enhance the description here and you can omit getRegionAtAnotherScale()

manthey:
Shouldn't the curl command be `!curl -OJ "$wsi_url"` (or `%%curl -OJ "$wsi_url"`)?  I think it needs the quotes and $ wrapped around the variable.
```

cdeepakroy:
No particular reason for using dict. I just find it more convenient when there are a lot of parameters (takes slightly less typing - no quotes). I tried to keep it consistent everywhere.

PR #530, opened 2018-08-30
[WIP] Support selecting multiple annotation elements for the context menu

jbeezley:
This is somewhat functional now.  There is an error occurring in a geojs animation frame callback that I'm having a difficult time debugging.  I had seen it before occasionally, but it didn't seem to hurt anything.  With these changes it happens every time you change the group via the context menu.

```
VM47832:54779 TypeError: Cannot read property 'r' of undefined
    at createGLPolygons (<anonymous>:61330:34)
    at gl_polygonFeature._build (<anonymous>:61407:6)
    at _update (<anonymous>:61433:15)
    at processAnimationFrame (<anonymous>:54777:17)
```

I submitted https://github.com/OpenGeoscience/geojs/pull/911 as a workaround, but it would be better to fix the underlying issue.  @manthey could you take a look?


manthey:
Somehow the style of the polygon is totally blank (not even the defaults), so when the fill color is requested, instead of a color object, it ends up with `undefined`.  I'm looking into it.

manthey:
There is some sort of race condition.  We remove a features via `this.viewerWidget.removeAnnotation(this.selectedAnnotation);`, then put them back via `this.viewerWidget.drawAnnotation(this.selectedAnnotation, {fetch: false});`, but somewhere along the way we try to draw them despite them having been removed.

manthey:
Once OpenGeoscience/geojs#915 is merged, we can use the new release with large_image.

jbeezley:
I think this is working now.  PTAL.

cooperlab:
@kheffah is ready to test this - let's get a PR and then update the Docker.

manthey:
@cooperlab Docker images are building now.

PR #542, opened 2018-09-13
WIP (do not merge yet):Update test configuration to enable testing histomicsTK inside the do…

fbudin69500:
…cker container

Mongodb URL has to be updated to not look for Mongodb on localhost but on the local
docker network.


manthey:
I'd prefer to do this via iptables rather than changing the girder checkout (see PR #546), that way you can try different version of Girder by just checking them out.

fbudin69500:
Superseded by https://github.com/DigitalSlideArchive/HistomicsTK/pull/546

PR #546, opened 2018-09-14
Improve the ability to run tests inside the docker container.

manthey:
Use iptables to allow girder tests to reach the mongo container as if it were on localhost from inside the main docker container.

Install some extra utilities, such as npx and eslint in the main docker to allow testing.

With this change, after starting the docker group with ./deploy_docker.py, the following should result in all HistomicsTK tests passing:

```
docker exec -it histomicstk_histomicstk bash -c "cd ../build && cmake -DRUN_CORE_TESTS:BOOL=OFF -DTEST_PLUGINS:STRING=HistomicsTK ../girder && make -j && ctest --output-on-failure"
```


manthey:
@cdeepakroy, @fbudin69500 This should result in dockers that will work for cmake testing directly.

cdeepakroy:
I confirm that all the tests pass with the changes made in this PR. 

I am yet to check if the tests pass if i mount in a local clone of HistomicsTK into /opt/histomicstk in the histomicstk_histomicstk container.

fbudin69500:
Works on my computer, even after mounting the source code from the host computer!

PR #548, opened 2018-09-22
Attempt to prevent the context menu from overflowing

jbeezley:
This adjusts the y coordinate of the context menu so that the bottom is
no lower than the bottom edge of the visible window.  The context menu
can still overflow if the vertical size of the window is too small to
fit the context menu.  If this becomes an issue we can make further
adjustments, e.g. put it in a scrolling container or add a "show more"
button.

### DIFF
--- a/web_client/views/body/ImageView.js
+++ b/web_client/views/body/ImageView.js
@@ -659,6 +659,7 @@ var ImageView = View.extend({
         // Defer the context menu action into the next animation frame
         // to work around a problem with preventDefault on Windows
         window.setTimeout(() => {
+            const $window = $(window);
             const menu = this.$('#h-annotation-context-menu');
             const position = evt.mouse.page;
             menu.removeClass('hidden');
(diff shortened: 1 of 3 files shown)


web_client/views/body/ImageView.js
```diff
@@ -659,7 +662,15 @@ var ImageView = View.extend({
             const menu = this.$('#h-annotation-context-menu');
             const position = evt.mouse.page;
             menu.removeClass('hidden');
-            menu.css({ left: position.x, top: position.y });
+
+            // adjust the vertical position of the context menu
+            // == 0, above the bottom; < 0, number of pixels below the bottom
+            // the menu height is bigger by 20 pixels due to extra padding
+            const belowWindow = Math.min(0, $(window).height() - position.y - menu.height() + 20);
+            // ensure the top is not above the top of the window
```
manthey: We should also do this for the horizontal.  I see a problem when I right click an element at the far right of the window (minimize the right panels to get it there).

PR #549, opened 2018-09-24
Save style group to localstorage when resetting

jbeezley:
Prior to this change, resetting the style groups only applied to the
groups currently in memory.  When reloading the page, they reverted back
to the old values.

### DIFF
--- a/web_client/dialogs/editStyleGroups.js
+++ b/web_client/dialogs/editStyleGroups.js
@@ -145,7 +145,10 @@ const EditStyleGroups = View.extend({
             styleModels = _.map(styles, function (style) {
                 return new StyleModel(style);
             });
-            this.collection.reset(styleModels, {merge: true});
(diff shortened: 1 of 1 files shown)


web_client/dialogs/editStyleGroups.js
```diff
@@ -154,6 +154,7 @@ const EditStyleGroups = View.extend({
             if (oldid && this.collection.get(oldid)) {
                 this.model.set(this.collection.get(oldid).toJSON());
             }
+            this.collection.each((model) => { model.save(); });
```
manthey: This saves groups that exist, but doesn't delete removed groups from local storage.

manthey: Perhaps we can add `while (this.collection.length) { this.collection.at(0).destroy(); }` just before we reset the collection.

PR #553, opened 2018-10-03
Change the download path to vips.

manthey:
The path we had been using no longer works. This should be the more canonical path.

Also, pin the version of the wheel package used by scikit-build.  This should be removed once scikit-build is updated.

### DIFF
--- a/ansible/roles/girder-histomicstk/tasks/main.yml
+++ b/ansible/roles/girder-histomicstk/tasks/main.yml
@@ -116,6 +116,12 @@
     state: present
   become: true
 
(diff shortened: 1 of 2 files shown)


fbudin69500:
Nice. I had created a new branch in my repo to solve the same issue. I had, for some reason, to modify a little bit how the configuration of libvips was done as ./configure did not exist in the archive I was downloading:
https://github.com/fbudin69500/HistomicsTK/commit/54edfe7a32f7a7c6cfd3663ef74ba571d41cc2c7

Your PR seems simpler!

PR #559, opened 2018-10-10
Improve blob detection tests

fbudin69500:
Due to a bug in scikit-image, it is not possible to have the exact same results
with the blob detection algorithms across all computers [1]. The variation in the
result is minimal and does not inpact in any major way the output of the algorithm.
However, it is not possible to perform a simple image to image comparison between
a ground truth image and a computed image. This would make the blob detection test
fail on some computers.

To work around this bug and still test the blob detection functionality, two new
comparison methods were implemented:

First, we do not compare directly the entire computed image to a ground truth, pixel
by pixel. Instead, we implemented an algorithm that compares the two given images by
moving them up and down one line. Only the best fitting lines are used to compute
the error between the two images. This allows to ignore 'rogue' lines that do not
match the ground truth because of the bug described above.

Second, as an additional test, we also compare that the maxima found by the blob
detection algorithm match in the computed image and in the ground truth.

Finally, the ground truth data has been updated to be saved as float 32 instead of
float16, but avoid increasing data size, numpy arrays are now compressed. The total
size of the numpy arrays used in this tests are down from about 18MB to less than 12MB.

[1] https://github.com/scikit-image/scikit-image/issues/3445

Yossarian


fbudin69500:
@cdeepakroy Solving the failing test on my new computer due to the scikit-image bug, and reducing the amount of data downloaded by using compressed numpy data.

PR #561, opened 2018-10-11
Add previous/next image buttons

jbeezley:
screenshot

This is nearly finished.  I just noticed while testing this that the endpoint I added doesn't filter out non-image items causing the web client to display errors.  That should be a fairly simple change once I get around to it.

### DIFF
--- a/server/image_browse_resource.py
+++ b/server/image_browse_resource.py
@@ -6,6 +6,10 @@ from girder.exceptions import RestException
 from girder.models.folder import Folder
 
 
+def _isLargeImageItem(item):
+    return item.get('largeImage', {}).get('fileId') is not None
+
+
 class ImageBrowseResource(ItemResource):
     """Extends the "item" resource to iterate through images im a folder."""
 
@@ -22,7 +26,7 @@ class ImageBrowseResource(ItemResource):
         folderModel = Folder()
         folder = folderModel.load(
             currentImage['folderId'], user=self.getCurrentUser(), level=AccessType.READ)
-        allImages = [item for item in folderModel.childItems(folder) if 'largeImage' in item]
+        allImages = [item for item in folderModel.childItems(folder) if _isLargeImageItem(item)]
         try:
             index = allImages.index(currentImage)
         except ValueError:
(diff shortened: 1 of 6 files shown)


jbeezley:
This is now ready.

manthey:
This closes the analysis that is open, whereas changing image via the buttons doesn't.

server/image_browse_resource.py
```diff
@@ -0,0 +1,56 @@
...
+
+        self.resourceName = 'item'
+        apiRoot.item.route('GET', (':id', 'next_image'), self.getNextImage)
+        apiRoot.item.route('GET', (':id', 'previous_image'), self.getPreviousImage)
+
+    def getAdjacentImages(self, currentImage):
+        folderModel = Folder()
+        folder = folderModel.load(
+            currentImage['folderId'], user=self.getCurrentUser(), level=AccessType.READ)
+        allImages = [item for item in folderModel.childItems(folder) if 'largeImage' in item]
```
manthey: This should probably check `if item.get('largeImage', {}).get('fileId')` or something equivalent.  An item can have a `largeImage` entry if it is pending being turned into a large image (for instance, being processed via vips), but will only have a `fileId` within the `largeImage` record if it actually has an image.

web_client/views/layout/HeaderImageView.js
```diff
@@ -33,9 +38,42 @@ var HeaderImageView = View.extend({
...
+            this.previousImageLink = null;
+            this.render();
+            return;
+        }
+
+        if (router.getQuery('analysis')) {
+            analysisQuery = `&analysis=${router.getQuery('analysis')}`;
+        }
+        $.when(
+            restRequest({
```
manthey: Every time the analysis is changed, we refetch the next and previous images ids.  Similarly, when we first load an image with the analysis open we end up fetch the next and previous image ids twice.  Do we care about this inefficiency?  Should the next and previous ids only be fetch on h:imageOpened and the be reused?  I was testing in a folder with ~7000 images and this takes a noticeable time.

jbeezley: Should be resolved now.  If performance of the endpoints becomes an issue, I can look into baking it into a mongo query rather than loading all of images into memory.

dgutman:
Can someone let me know when the latest docker image has this change

manthey:
They latest docker images were just uploaded.

PR #565, opened 2018-10-24
Update Python code style to match new PEP-8 requirements

fbudin69500:
PEP-8 has been updated to recommend breaking a line before a binary operator
instead of before as it had been recommended for years. This modification was
integrated in flake8 3.6 which is used to test HistomicsTK Python code style.

[1] https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator


manthey:
In Girder upstream, we are disabling W504 (and thus not moving operators to the next line).  I'd rather follow Girder than diverge from it.

I think when that PR gets merged to Girder, HistomicsTK will just inherit the change.

manthey:
I expect it will be hours (not days).

manthey:
CI is passing again now that Girder has been updated.

PR #570, opened 2018-11-20
Patch openslide so that it can read Girder mrxs files.

manthey:
This alters the openslide source so that, once built, mrxs files can be read through the girder mount filesystem layout.


manthey:
@cdeepakroy Can you review this?  There are some publicly available mrxs files here: http://openslide.cs.cmu.edu/download/openslide-testdata/Mirax/
If you import them into a non-filesystem assetstore or upload them to a filesystem assestore, and you are using girder mount, you should be able to view the mrxs files.

Note that https://github.com/girder/large_image/pull/318 prevents the files from being interpreted as low-res JPEGs, and if you don't have that update then you may have to toggle the large_image use for a file off and back on for it to work correctly.

PR #575, opened 2019-01-15
Pin numpy and scipy.

manthey:
Newer versions of numpy and scipy don't work with scikit-image.  Until that is fixed, pin to older versions.


cdeepakroy:
For the records, it would be great if you can log the error that was occurring here and link to the scikit-image issue/PR that is looking into it.

manthey:
This patch is necessary if you see an error like:
```
from numpy.lib.arraypad import _validate_lengths
ImportError: cannot import name _validate_lengths
```

PR #580, opened 2019-01-23
Pin pip; pip 19.0 doesn't work with pylibtiff.

manthey:


manthey:
Closing this as pip 19.0.1 fixes the problem (though their release notes don't reflect the problem I was seeing).

PR #595, opened 2019-02-12
Pin setuptools and virtualenv in the Docker build.

manthey:
Recent setuptools and virtulenv don't work directly with out current build process.  Pin them until we can fix it.


manthey:
@cdeepakroy I noticed that Dockerhub failed to build from master.  This should fix it.

PR #596, opened 2019-02-12
Specify girder worker concurrency in deploy_docker.py.

manthey:
This is a command line option to start girder_worker with a specified concurrency.  It still defaults to 2.


manthey:
@cdeepakroy when using deploy_docker.py to start a worker docker, use `--concurrency=<positive number>`.  For your GPU testing, it is probably useful to limit this to 1 until Dask GPU sharing is sorted out.

PR #597, opened 2019-02-14
Document the keyboard control for show/hide all.

manthey:
This adds to the tooltip for these controls.


manthey:
@cooperlab As far as I can tell, this is the only keyboard shortcut we've added other than letting escape cancel some things.  There are some keyboard controls that come with large_image, such as using the arrow keys to pan the image.  I'm not sure where we should show those.

PR #604, opened 2019-02-21
Handle cases when no nuclei are detected in an image

cdeepakroy:
This PR:
* introduces several changes to handle the case where no nuclei were detected in an image using the Kofahi algorithm. 
* Take nuclei foreground mask instead of foreground threshold as input in `detect_nuclei_kofahi`. This will allow users to compute the foreground mask in other ways (e.g. using an automated thresholding algorithm).


manthey:
If, on ComputeNucleiFeatures, I choose a tiny ROI such that I get no features at all, I still get an error:
```
Traceback (most recent call last):
  File "ComputeNucleiFeatures/ComputeNucleiFeatures.py", line 345, in <module>
    main(CLIArgumentParser().parse_args())
  File "ComputeNucleiFeatures/ComputeNucleiFeatures.py", line 296, in main
    ignore_index=True)
  File "/build/miniconda/lib/python2.7/site-packages/pandas/core/reshape/concat.py", line 228, in concat
    copy=copy, sort=sort)
  File "/build/miniconda/lib/python2.7/site-packages/pandas/core/reshape/concat.py", line 262, in __init__
    raise ValueError('No objects to concatenate')
ValueError: No objects to concatenate
```

manthey:
I think the javascript in the test Spec files has to still be ES5 (`var` rather than `const`), because we don't transpile it and phantomjs doesn't have ES6.

PR #609, opened 2019-03-06
Use the new base-docker-image.

manthey:
Most of these changes relate to no longer having conda and no longer needing to reinstall openslide-python and pylibtiff.


manthey:
This should only be merged once the new base image is available on docker hub.

PR #612, opened 2019-03-08
Refactor for easier deployment.

manthey:
This moves the various cli code from the server directory to the histomicstk/cli directory.  As such, when installed without girder, the cli code is available.

Although MANIFEST.in has been altered, `python setup.py sdist` doesn't quite produce a fully working pypi package.  While this package includes all relevant filse, when pip installed elsewhere with manually installed prerequisites, for reasons I don't yet understand, the included xml cli specifications are not copied into the deployed python packages.


manthey:
@cdeepakroy, @cooperlab This refactors where the cli python code resides.  I think with a little help regarding scikit-build (or switching away from scikit-build), we can make it so that this can be pushed to pypi and be fully pip installable.

histomicstk/__init__.py
```diff
@@ -19,4 +21,5 @@
     'preprocessing',
     'segmentation',
     'utils',
+    'cli',
```
cdeepakroy: Will this make the installation of ctk_cli a requirement? Could we not expose it until we figure out how to put ctk_cli on pypi or build wheels for it as part of our pypi deployment.

manthey: No.  ctk_cli isn't a requirement unless you import one of the cli main modules (e.g., `from histomicstk.cli.NucleiDetection import NucleiDetection`).  If you do that, and have not installed ctk_cli, you'll get an import error.

manthey: To be clear, you can even do `import histomicstk.cli.NucleiDetection` and you won't trigger the import error.

PR #640, opened 2019-04-29
Superpixel Segmentation CLI

slee172:
This PR makes a superpixel segmentation CLI that creates two files (features and boundaries) used as a dataset for HistomicsML.

For boundaries, compute_boundary_data function works on in parallel using Dask and extracts superpixel boundaries from WSIs.

For features, compute_superpixel_data function works on CPU/GPU and extracts features for each superpixel patch.

CLI input is a directory of WSIs and a fitted PCA model and CLI output is two files(features and boundaries) which are used as a dataset for HistmocsML.

### DIFF
--- a/histomicstk/cli/SuperpixelSegmentation/CreateDataset.py
+++ b/histomicstk/cli/SuperpixelSegmentation/CreateDataset.py
@@ -355,7 +355,7 @@ def main(args):		# noqa: C901
 
         ts_metadata = ts.getMetadata()
 
-        print (json.dumps(ts_metadata, indent=2))
+        print(json.dumps(ts_metadata, indent=2))
 
         is_wsi = ts_metadata['magnification'] is not None
 
--- a/histomicstk/cli/SuperpixelSegmentation/CreateDataset.py
+++ b/histomicstk/cli/SuperpixelSegmentation/CreateDataset.py
@@ -28,593 +28,589 @@ from skimage.segmentation import slic
 from histomicstk.cli import utils as cli_utils
 
 import logging
+
 logging.basicConfig(level=logging.CRITICAL)
 
 
 def compute_boundary_data(img_path, tile_position, args, it_kwargs,
-                          src_mu_lab=None, src_sigma_lab=None):
+						  src_mu_lab=None, src_sigma_lab=None):
+	# get slide tile source
+	ts = large_image.getTileSource(img_path)
 
-    # get slide tile source
-    ts = large_image.getTileSource(img_path)
+	# get requested tile information
+	tile_info = ts.getSingleTile(
+		tile_position=tile_position,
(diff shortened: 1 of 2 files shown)


histomicstk/cli/SuperpixelSegmentation/CreateDataset.py
```diff
@@ -0,0 +1,615 @@
+import os
```
cdeepakroy: @slee172 

The name of the CLI's python/xml file and its parent directory should match.

histomicstk/cli/SuperpixelSegmentation/CreateDataset.xml
```diff
@@ -0,0 +1,237 @@
...
+      <description>Standard deviation of reference image in LAB color space for Reinhard color normalization</description>
+      <longflag>reference_std_lab</longflag>
+      <default>0.57506023, 0.10403329, 0.01364062</default>
+    </double-vector>
+    <double-vector>
+      <name>source_mu_lab</name>
+      <label>Source Mean LAB</label>
+      <description>Mean of source image in LAB color space for Reinhard color normalization</description>
+      <longflag>source_mu_lab</longflag>
+      <default>8.96447078, -0.0879669, 0.04412108</default>
```
cdeepakroy: Source mu and std in Lab is supposed to computed from the slide being processed.

You can provide these as parameters but then the default values should be some None value like -1, -1, -1. If these parameters are not supplied then it should compute them on the fly.

histomicstk/cli/SuperpixelSegmentation/CreateDataset.xml
```diff
@@ -0,0 +1,237 @@
...
+      <default>5</default>
+    </double>
+    <double>
+      <name>superpixel_tile_size</name>
+      <label>Superpixel Analysis Tile Size</label>
+      <description>Tile size for blockwise analysis</description>
+      <longflag>superpixel_tile_size</longflag>
+      <default>4096</default>
+    </double>
+    <double>
```
cdeepakroy: Add super pixel parameters into a separate parameter group.

histomicstk/cli/SuperpixelSegmentation/CreateDataset.xml
```diff
@@ -0,0 +1,237 @@
...
+      <name>analysis_mag</name>
+      <label>Analysis Magnification</label>
+      <description>The magnification at which the analysis should be performed.</description>
+      <longflag>analysis_mag</longflag>
+      <default>5</default>
+    </double>
+    <double>
+      <name>superpixel_tile_size</name>
```
cdeepakroy: why rename this with "superpixel_" prefix?

slee172:
Generating a superpixel dataset is already implemented by using a docker image, so I closed this PR.

PR #642, opened 2019-04-29
Restrict some endpoints to require a logged in user.

### LINKED ISSUE #176: Restrict download permissions
The download endpoint should be more restricted that general item access.  Perhaps it should only be allowed if the current user has write access to the item or an appropriately scoped token.

manthey:
This adds a girder configuration value to control if the restriction is added.  When turned on via the grider config value `[histomicstk]` `restrict_downloads: True`, all standard girder download endpoints and the large_image `item/{id}/tiles/images/{image}` endpoint are restricted to logged-in-users.  This is in addition to ordinary ACL restrictions.

The restriction is turned on for our standard deployment.

This resolves #176.

### DIFF
--- a/plugin_tests/histomicstk_test.py
+++ b/plugin_tests/histomicstk_test.py
@@ -21,6 +21,7 @@ config.loadConfig()  # Must reload config to pickup correct port
 
 # boiler plate to start and stop the server
 def setUpModule():
+    config.getConfig()['histomicstk'] = {'restrict_downloads': True}
     base.enabledPlugins.append('HistomicsTK')
     base.startServer()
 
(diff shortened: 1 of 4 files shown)


plugin_tests/histomicstk_test.py
```diff
@@ -255,3 +256,24 @@ def testGetDockerImage(self):
             resp = self.request(path='/HistomicsTK/HistomicsTK/docker_image', user=user)
             self.assertStatusOk(resp)
             self.assertEqual(len(resp.json), count)
+
+    def testRestrictDownloads(self):
+        publicFolder = self.model('folder').childFolders(
+            self.user, 'user', filters={'name': 'Public'}
+        ).next()
```
jbeezley: Isn't `iterator.next` python 2 only?

manthey: I think this is a mongo cursor next.

PR #646, opened 2019-05-07
Update the context menu after styles are edited.

manthey:
Before, if you added or removed styles, the context menu could show stale style names.


manthey:
@cooperlab This may have been the cause of some of the issues you saw.  Can you see if you still have issues once this branch is in use?  If so, recording a short video of it happening will make it easier to hunt down the problem.

PR #647, opened 2019-05-08
Build and publish wheels for the library

manthey:
This adds code to build wheels:

- `Dockerfile-wheels` builds the wheels for linux.
- `build_wheels.sh` builds the docker file and copies the wheels to a temp wheels directory.
- `test_wheels.py` does a minimal check to ensure that the wheels work.  If you built the wheels locally, it will use those wheels; if not, it will use the published ones on pypi.

This also uses the published version of ctk-cli, patching the one class that we had changed.  This avoids either having to build a wheel for ctk-cli or require an old version of pip to install from a git repo.

This also switches library versioning to setuptools-scm so that it is based on git tags.  When we tag a version on master of the form v1.0.0, a new package will be published to pypi with that version.  Otherwise, whenever we merge to master (or this special wheel-ci branch for testing), a development version is published to pypi.

Because HistomicsTK depends on large_image, which depends on complex libraries (openslide and libtiff, at least), you can install everything with the wheels for those libraries via `pip install histomicstk -f https://manthey.github.io/large_image_wheels`.  Alternately, you can install working versions of libopenslide and libtiff and then do the installation via `pip install histomicstk` (you'll probably also need to pip install numpy first in this case).


README.rst
```diff
@@ -50,6 +50,16 @@ HistomicsTK can be used in two ways:
   and `examples <https://digitalslidearchive.github.io/HistomicsTK/examples.html>`__
   for more information.
 
+  This can be installed via `pip install histomicstk`.
```
thewtex: > This can be installed via `pip install histomicstk`.

to 

> This can be installed on Linux via `pip install histomicstk`.

correct?

README.rst
```diff
@@ -50,6 +50,16 @@ HistomicsTK can be used in two ways:
   and `examples <https://digitalslidearchive.github.io/HistomicsTK/examples.html>`__
   for more information.
 
+  This can be installed via `pip install histomicstk`. 
+
+  If you want to specify precise library versions for reading various
+  microscopy formats, you will need to install various libraries via your
+  package manager or from source (libopenslide-dev, libtiff-dev).
```
thewtex: Does this mean that the wheel ships with Openslide and TIFF library binaries and system ones will be used? Or is it a requirement to install these Linux package manager packages? Clarification in the documentation here would be helpful.

README.rst
manthey: Yes -- we should probably publish a variant without the binaries that would install on non-linux systems when there are appropriate other libraries installed.

README.rst
manthey: The HistomicsTK wheel only contains the built cython code -- everything else is obtained from other packages.  It depends on large_image and requires the openslide and tiff tile sources (as those are the most likely to be useful for microscopy).  Those require *either* that the appropriate libraries are installed on the system, in which case the published python packages to access those libraries work (though with some known bugs) *or* we can use the `--find-links` (`-f`) option in `pip install` to install from my built wheels for python-openslide and libtiff.  In that case, those libraries will use my built versions and NOT any system version (even if so installed).

My wheels for large_image are necessarily not official (as they aren't published by the openslide and pylibtiff projects -- we'd love to get there some day).  To further complicate things, if you try to pip install without the `-f` option and then retry pip install with it, pip will pull from the local cache and not check the find-links address unless you ask to forcibly reinstall AND not use the cache (`--force-reinstall --no-cache-dir`).

The short form is, on linux, installing via `pip install histomicstk -f https://manthey.github.io/large_image_wheels --force-reinstall --no-cache-dir` will always work, but fixes the versions of various libraries in an opinionated manner.  Anything else might not work and requires more explanation.

Recommendations on how to state this clearly and succinctly are welcome.

manthey: I pushed a change which I hope makes things clearer.

README.rst
```diff
@@ -50,15 +50,22 @@ HistomicsTK can be used in two ways:
...
-
-  To avoid installing these complex dependencies, you can install
-  using specific prebuilt wheels via 
-  `pip install histomicstk -f https://manthey.github.io/large_image_wheels`.
+  This can be installed on Linux via `pip install histomicstk`.
+
+  HistomicsTK uses the `large_image`_ library to read and various microscopy
+  image formats.  Depending on your exact system, installing the necessary 
+  libraries to support these formats can be complex.  There are some
+  non-official prebuilt libraries available for Linux, that can be included as
```
thewtex: extra comma

README.rst
```diff
@@ -50,15 +50,22 @@ HistomicsTK can be used in two ways:
...
+  part of the installation by specifying 
+  `pip install histomicstk --find-links https://manthey.github.io/large_image_wheels`.
+  Note that if your previously installed HistomicsTK or large_image without
```
thewtex: your -> you

docs/installation.rst
```diff
@@ -12,26 +12,25 @@ these scenarios.
 Installing HistomicsTK as a Python toolkit
 ------------------------------------------
 
-HistomicsTK depends on large_image_ for reading large multi-resolution
-whole-slide images in a tiled fashion. Please see the Github repo of
-large_image to find out how to install it as a Python toolkit/package.
+On Linux, HistomicsTK can be installed via pip.  You can specify the
+`--find-links` option to get prebuilt libraries for ready some common image
```
cdeepakroy: for "reading" some common image formats

PR #656, opened 2019-05-17
Add a Quarantine option.

manthey:
The a folder is specified in the plugin settings, a Quarantine button will appear next to each item that a user has write access to.  When clicked, the item will be moved to the quarantine folder.

Two endpoints have been added, HistomicsTK/quarantine/:itemId and HistomicsTK/quarantime/:itemId/restore to move an item to the quarantine folder or to put it back.

Currently, only the itemList widget shows the quarantine action.  We probably should also show it on the individual item page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ build
 _build
 .coverage
 *.egg
+.eggs
 *.egg-info
 .cache
 .idea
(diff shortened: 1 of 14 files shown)


manthey:
Note the "Q" buttons:

image

PR #663, opened 2019-05-30
Don't set the current assetstore when deploying.

manthey:
The auto-generated filesystem assetstore was always being set as the current assetstore.  This will happen on the first installation anyway, so it isn't needed.  On subsequent deployments, if the current has been changed, this was changing it back, which is confusing and undesired.


dgutman:
discussed and will test- current feature is broken

PR #679, opened 2019-08-15
Use the large_image 2.x-maintenance branch.

manthey:
Large image master will soon require Girder 3.


manthey:
@thewtex This just pins to a Girder 2 branch of large image.  Once it is merged, large image will switch so its master is the Girder 3 version.

PR #703, opened 2019-09-19
Tissue detection module

kheffah:
# Tissue Detection module

**Overview:** 
This includes tools to detect tissue from an item (slide) using its thumbnail. The basic functionality includes a series of gaussian smoothing and otsu thresholding steps to detect background versus foreground pixels. Optionally, an initial step is performed whereby color deconvolution is used to deparate hematoxylin and eosin stains (assuming H&E stained slides) to make sure only cellular areas are segmented. This proves to be useful in getting rid of sharpie markers. A size threshold is used to keep only largest contiguous tissue regions.

This slide used as a test example:

[TCGA-A1-A0SK-01Z-00-DX1](http://candygram.neurology.emory.edu:8080/histomicstk#?image=5d817f5abd4404c6b1f744bb)

**@TODO** -> Color normalization prior to deconvolution

**Here's the result:**

From left to right: Slide thumbnail, Tissue regions, Largest tissue region

image

**Where to look?**

```
|_ histomicstk/
|   |_saliency/
|      |_tissue_detection.py 
|      |_tests/
|         |_tissue_detection_test.py
|_ docs/
    |_examples/
       |_tissue_detection.ipynb
```


kheffah:
@manthey This is just a simple stand-alone light module for tissue detection. It passed all tests so whenever is convenient feel free to review and merge. 
Thanks!

manthey:
It isn't directly related to this PR, but the tests print a warning for me:
```
histomicstk/preprocessing/color_conversion/rgb_to_sda.py:48: RuntimeWarning: divide by zero encountered in log
```
This has to do with the mentioned file taking the log of 0 as part of this line: `im_sda = -np.log(im_rgb/(1.*I_0)) * 255/np.log(I_0)`.  When I_0 is not explicitly set, the stain darkness is computed from (color value + 1), and we never get a log of 0.  However, when I_0 is specified, the stain darkness is computed from (color value), which can be 0.

It seems like the rgb_to_sda functions should be clearer about this, and that something different should be done in the `tissue_detection.py`'s `_deconv_color` function.

kheffah:
@manthey Thanks, I totally missed that!! It was a simple fix, which I just did. Let me know if you approve of the PR. I am working on extensions to this PR but I'd like this to be merged first whenever you're ready.

PR #704, opened 2019-09-19
Adjust the load indicator when centroid display is used.

manthey:


manthey:
@kheffah This goes along with https://github.com/girder/large_image/pull/369.

Issue #705, opened 2019-09-21
cannot install HistomicsTK as there is no requirements.txt in the folder


satotty:
when I try to install HistomicsTK following the steps as follows and execute '$ pip install --no-cache-dir -r requirements.txt' , an error comes up saying:
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'

i also checked the HistomicsTK file but there is no such file: 'requirements.txt'

Once large_image is installed as a python package, HistomicsTK can be installed as follows:

$ git clone https://github.com/DigitalSlideArchive/HistomicsTK.git
$ cd HistomicsTK
$ pip install git+https://github.com/cdeepakroy/ctk-cli
$ pip install --no-cache-dir -r requirements.txt
$ pip install -e .

kheffah:
@satotty Sorry that's documentation that's a bit outdated. Here's what you need to do. 

**To install HistomicsTK using PyPI** :

```bash
$ pip install histomicstk
```

**To install HistomicsTK from source**:

```bash
$ git clone https://github.com/DigitalSlideArchive/HistomicsTK/
$ cd HistomicsTK/
$ python -m pip install scikit-build ninja cython numpy nimfa
$ python setup.py develop
```
The third line is there because there are some issues in the setup.py file that sometimes prevent the installation of some of these libraries, at least in my experience. 

@manthey Should I update the installation instructions accordingly?

PR #706, opened 2019-09-23
Updated installation instructions

kheffah:
This addresses issue #705 .

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,8 +49,19 @@ HistomicsTK can be used in two ways:
   `api-docs <https://digitalslidearchive.github.io/HistomicsTK/api-docs.html>`__
   and `examples <https://digitalslidearchive.github.io/HistomicsTK/examples.html>`__
   for more information.
-
-  This can be installed on Linux via ``pip install histomicstk``.
+  
+  Installation instructions on Linux:
+  
+  *To install HistomicsTK using PyPI*::
+  
+  $ pip install histomicstk
+  
+  *To install HistomicsTK from source*::
+  
+  $ git clone https://github.com/DigitalSlideArchive/HistomicsTK/
+  $ cd HistomicsTK/
+  $ python -m pip install scikit-build ninja cython numpy nimfa
+  $ python setup.py develop
 
   HistomicsTK uses the `large_image`_ library to read and various microscopy
   image formats.  Depending on your exact system, installing the necessary 
--- a/README.rst
+++ b/README.rst
@@ -54,14 +54,14 @@ HistomicsTK can be used in two ways:
   
   *To install HistomicsTK using PyPI*::
   
-  $ pip install histomicstk
+  $ python -m pip install histomicstk
   
   *To install HistomicsTK from source*::
   
   $ git clone https://github.com/DigitalSlideArchive/HistomicsTK/
   $ cd HistomicsTK/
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -49,8 +49,19 @@ HistomicsTK can be used in two ways:
...
+  
+  *To install HistomicsTK using PyPI*::
+  
+  $ pip install histomicstk
+  
+  *To install HistomicsTK from source*::
+  
+  $ git clone https://github.com/DigitalSlideArchive/HistomicsTK/
+  $ cd HistomicsTK/
+  $ python -m pip install scikit-build ninja cython numpy nimfa
```
manthey: I think this should be the packages in the setup_requires field of setup.py.  That is, `python -m pip install scikit-build Cython cmake setuptools-scm`.  If we need to explicitly install ninja anything else, they should be added to setup_requires.

README.rst
```diff
@@ -49,8 +49,19 @@ HistomicsTK can be used in two ways:
...
+  $ python setup.py develop
```
manthey: I would prefer to list this as `python -m pip install -e .` (or we could assume that pip is on the command line and just have `pip install -e .`.

README.rst
kheffah: OK, incorporated changes. It turns out `numpy` needs to be explicitly installed so I updated the instructions accordingly.

README.rst
kheffah: OK, done.

setup.py
```diff
@@ -58,12 +58,12 @@ def prerelease_local_scheme(version):
         'Cython>=0.25.2',
         'scikit-build>=0.8.1',
         'cmake>=0.6.0',
+        'numpy>=1.12.1',
     ],
     install_requires=[
         'ctk-cli>=1.5',
         # scientific packages
         'nimfa>=1.3.2',
-        'numpy>=1.12.1',
```
manthey: numpy needs to remain in the `install_requires` section (in addition to the `setup_requires` section).  The `setup_requires` section is what is necessary before you can install; the `install_requires` section is needed to run it.  If you install from a wheel, you don't have to install the `setup_requires` components, and therefore (I think), installing from a wheel wouldn't install numpy.

kheffah: Got it, thanks! OK done.

README.rst
satotty: I tried from line 61-64 on command line of my windows computer but when I executed the 64th line, an error came up. I downloaded Microsoft Visual Studio 2019 but it didn’t help anything. What should I do?

The error says:

********************************************************************************
scikit-build could not get a working generator for your system. Aborting build.

Building windows wheels for Python 3.7 requires Microsoft Visual Studio 2017.
Get it with "Visual Studio 2017":

  https://visualstudio.microsoft.com/vs/

********************************************************************************

kheffah: @satotty I'm not sure how this works on the windows commandline. Ideally, you should be working inside a conda environment to keep things clean. I'd suggest you:

1- Install Anaconda on Widows: https://docs.anaconda.com/anaconda/install/windows/
2- Create a new conda envirnment, say we all it `htk`, and install pip inside it, using something like:
```bash
conda create --name htk
conda install -n htk pip
```
3- Inside the conda command line, activate the environment: `conda activate htk`
4- Follow the same installation instructions provided in the new 'README.rst'

PR #721, opened 2019-10-08
select by region

### LINKED ISSUE #620: Lasso multiselect
Having the current multi-select feature has been very useful; others and I have been using it extensively for correcting annotations. The current multiselect needs one click per object, which is great for objects that are far apart. However, there are situations (appearing more and more frequently recently) when there are many annotations close to each other that need deletion; I'm talking 30+ objects. A lasso multiselect would be much more appropriate for bulk selection and editing. I'm not suggesting that we replace the current multiselect, but that we add an additional "annotation tool" beside the {point, rectangle, polygon, line, LASSO} which, when selected, allows the user to draw a polygon and select everything enclosed within it.

sgratzl:
closes #620 

image

allow the user to draw a rectangle (see button in the annotation panel or using the keyboard shortcut s) and open a context menu for all the elements within this rect. 

notes: some changes are just auto applied linter fixes

### DIFF
--- a/web_client/templates/panels/annotationSelector.pug
+++ b/web_client/templates/panels/annotationSelector.pug
@@ -17,6 +17,9 @@ block content
     span.h-annotation-fill-opacity-container(title=`Annotation fill opacity ${(fillOpacity * 100).toFixed()}%`)
       input#h-annotation-fill-opacity(
           type="range", min="0", max="1", step="0.01", value=fillOpacity)
+    .btn-group.btn-group-sm(role='group')
+      button.btn.btn-default.h-annotation-select-by-region(type='button', title='Select annotations by region.  Keyboard shortcut: a')
+        | #[span.icon-marquee]
 
   - var groups = _.keys(annotationGroups);
   - groups.sort();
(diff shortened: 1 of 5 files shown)


web_client/templates/panels/annotationSelector.pug
```diff
@@ -17,6 +17,9 @@ block content
     span.h-annotation-fill-opacity-container(title=`Annotation fill opacity ${(fillOpacity * 100).toFixed()}%`)
       input#h-annotation-fill-opacity(
           type="range", min="0", max="1", step="0.01", value=fillOpacity)
+    .btn-group.btn-group-sm(role='group')
+      button.btn.btn-default.h-annotation-select-by-region(type='button', title='Select annotations by region.  Keyboard shortcut: a')
```
manthey: Shortcut key is 's'

manthey:
It would be nice is the button or hotkey toggles the state of selection.  That is, clicking the button a turns off selection mode if it is on.

sgratzl:
> It would be nice is the button or hotkey toggles the state of selection. That is, clicking the button a turns off selection mode if it is on.

done

PR #722, opened 2019-10-08
prev/next for virtual folders

### LINKED ISSUE #660: Navigating virtual folders -- forward and backward arrows
When using virtual folders to organize data, the notion of "next" and "previous" slide break down. Instead, the DSA system takes you to the original folder where the slide is located, so the forward/backward arrow (next/previous item) take you to slides before/after the current slide in the **original**, not virtual, folder.
Suggested handling:
When the slide is visited from a virtual folder, the arrows should move within the virtual folder as opposed to the original folder where slide is stored.

sgratzl:
closes #660 

by handling virtual folders manually and keep track of it


manthey:
We have a PR in the digital_slide_archive plugin that changes this behavior (with an optional flag): https://github.com/DigitalSlideArchive/digital_slide_archive/pull/37/files

Since the HistomicsTK plugin requires the DSA plugin, once that is merged, it could be used.

sgratzl:
blocked by https://github.com/DigitalSlideArchive/digital_slide_archive/pull/37

PR #723, opened 2019-10-08
Switch ansible to use large image wheels.

manthey:
This is faster, smaller, and keeps the logic for building complex dependencies in one spot.

Something in the dependencies of ImageMagick has charged, so the existing approach would need to be updated in any case.


dgutman:
Just tested it on a new machine and it ran

PR #740, opened 2019-10-22
Allow dask arguments to be optional.

manthey:
The `create_dask_client` method required certain parameters to use dask; make them all optional.


kheffah:
Didn't test but logic makes sense.

PR #761, opened 2019-11-20
Make sure annotations are saved.

manthey:
We have a bug when saving annotations to a slow database or an intermittently connected server.  Specifically, when we are saving an annotation and another change occurs, we were not saving again (to prevent race conditions).  However, this means the second save might never happen.  Similarly, if the save failed, we would not retry it.

This fixes the issue by marking that we need to call save again after the first save request completes (success or fail).  In the event of failure, retry the save after a delay.

### DIFF
--- a/plugin_tests/client/annotationSpec.js
+++ b/plugin_tests/client/annotationSpec.js
@@ -1086,6 +1086,7 @@ $(function () {
                     $('#h-annotation-context-menu .h-remove-elements').click();
                     expect($('#h-annotation-context-menu').hasClass('hidden')).toBe(true);
                 });
+                girderTest.waitForLoad();
             });
 
             it('open a different image', function () {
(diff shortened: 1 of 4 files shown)


manthey:
@dgutman @kheffah This fixes our saving annotation issue.  A follow-on to this will be show a dialog when the user tries to navigate away when saving hasn't finished.

manthey:
@dgutman @kheffah Please give this a try.

A simple test on the system with a slow database: 

On the existing deployment:  Create a new annotation.  Add three points very quickly.  Wait a while. 
 Reload the webpage and then the annotation.  I expect you'll see just one point.

Checkout this branch.  Create a new annotation.  Add three points very quickly.  Wait until the spinner on the annotation tab isn't spinning.  Reload the webpage and then the annotation.  I expect you'll see all three points.

PR #765, opened 2019-12-06
Fix IndexError in simple_mask() if no peaks are found

hnakhoul:
If no peaks are found in the image histogram, the fallback definition of
TissuePeak throws an IndexError in Python 3. This commit addresses it by
converting the index to an integer.

### DIFF
--- a/histomicstk/utils/simple_mask.py
+++ b/histomicstk/utils/simple_mask.py
@@ -79,7 +79,7 @@ def simple_mask(im_rgb, bandwidth=2, bgnd_std=2.5, tissue_std=30,
     if len(Peaks) > 1:
         TissuePeak = Peaks[yHist[Peaks[1:]].argmax() + 1]
     else:  # no peak found - take initial guess at 2/3 distance from origin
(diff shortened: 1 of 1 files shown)


manthey:
@hnakhoul Is using item() better than asscalar?

hnakhoul:
While trying it out I got a note saying that asscalar is deprecated (since np version 1.16) and to use item() instead.

Not sure the change takes care of every issue, but would be grateful for a fix for those of us running HistomicsTK on Python 3.

manthey:
@hnakhoul Can you resolve the conflict I caused?  Once done, we'll merge this PR.  Thanks.

kheffah:
The commit here was cherry-picked into [this PR](https://github.com/DigitalSlideArchive/HistomicsTK/pull/852), and conflicts resolved.

kheffah:
The commit here was cherry-picked into [this PR](https://github.com/DigitalSlideArchive/HistomicsTK/pull/852), and conflicts resolved.

PR #773, opened 2020-01-06
Add user defined params for saliency

kheffah:
Replace a couple of saliency hard coded params with user defined params. This allows better control of visualization and more flexible workflows.


kheffah:
@manthey Is it possible to restart the build job for this and PR #775  ? They both have weird failures that I just don't get. Could the PIL library thing still be causing issues?

PR #778, opened 2020-01-13
Expand annotations to masks workflow

kheffah:
Significantly expand and improve the annotations to masks handler to be directly usable to getting data for training models using manually drawn annotations. 

**Overview:** 

This includes tools to parse annotations from an item (slide) into masks to use in training and evaluating imaging algorithms. Two "versions" of this workflow exist:

* Get labeled mask for any region in a whole-slide image (user-defined)

* Get labeled mask for areas enclosed within special "region-of-interest" (ROI) annotations that have been drawn by the user. This involves mapping annotations (rectangles/polygons) to ROIs and making one mask per ROI.

The user uses a csv file like the one in 
```histomicstk/annotations_and_masks/tests/test_files/sample_GTcodes.csv ```
to control pixel values assigned to mask, overlay order of various annotation groups, which groups are considered to be ROIs, etc. Note that we use the girder definition of term "group" here, which is an annotation style indicating a certain class, such as "tumor" or "necrosis".

This adds a lot of functionality on top of API endpoints that get annotations as a list of dictionaries, including handing the following complex situations:

* Getting RGB images and labeled masks at the same magnification/resolution

* User-defined regions to get, including "cropping" of annotations to desired bounds

* Getting user-drawn ROIs, including rotated rectangles and polygons

* Overlapping annotations

* "Background" class (eg. anything not-otherwise-specified is stroma)

* Getting contours and bounding boxes relative to images at the same resolution, to be used to trainign object localization models like Faster-RCNN. 

**There are four run modes:**

* **wsi**: get scaled up/down version of mask of whole slide

* **min_bounding_box**: get minimum box for all annotations in slide

* **manual_bounds**: use given ROI bounds provided by the 'bounds' param

* **polygonal_bounds**: use manually-drawn polygonal (or rectanglar) ROI boundaries

Be sure to checkout the annotations_to_masks_handler.ipynb jupyter notebook for implementation examples.

### CODE REVIEW

**Comment by kheffah** [comment]
@manthey the function ```get_image_and_mask_from_slide()``` in ```./histomicstk/annotations_and_masks/annotastions_to_masks_handler.py``` is designed in a way that should be compatible with conversion to an API endpoint. Or at least so I hope :) 

It would be a very nice endpoint to have though, as @cooperlab suggested, as it allows direct use of the data by any user/developer. They just draw annotations, then use this method to do the magic and parse everything into a dataset they can directly use for training models.

**Comment by kheffah** [comment]
@manthey Thanks for the recent fixes! Worked like a charm and now the build passes. Feel free to review this and I'll merge when it's done. Cheers!

File: histomicstk/annotations_and_masks/annotation_and_mask_utils.py [hunk 1]
```diff
@@ -39,6 +39,96 @@ def get_image_from_htk_response(resp):
...
+    ---------
+    list of dicts
+
+    """
+    if sf == 1.0:
+        return slide_annotations
+
+    for annidx, ann in enumerate(slide_annotations):
+        for elementidx, element in enumerate(ann['annotation']['elements']):
+            for key in element.keys():
```
manthey: This could be more compactly done by
```
for key, entry in element.items():
    if key in {'height', 'width'}:                    # Note that a set is marginally more efficient than a list
        element[key] = int(entry * sf)      
    elif key in {'center', 'points'}
        element[key] = (np.array(entry) * sf).astype(int).tolist()
```

histomicstk/annotations_and_masks/annotation_and_mask_utils.py
```diff
@@ -39,6 +39,96 @@ def get_image_from_htk_response(resp):
...
+        If neither MPP or MAG is provided, everything is retrieved without
+        scaling at base (scan) magnification.
+
+    Returns:
+    ----------
+    float
+        how much smaller (0.1 means 10x smaller) is requested image
+        compared to scan magnification (slide coordinates)
+    str
+        string to appnd to server request for getting slide region
```
manthey: appnd => append

histomicstk/annotations_and_masks/annotation_and_mask_utils.py
```diff
@@ -39,6 +39,96 @@ def get_image_from_htk_response(resp):
...
+
+    """
+    slide_info = gc.get("/item/%s/tiles" % slide_id)
+
+    if MPP is not None:
+        mm = 0.001 * MPP
+        sf = slide_info['mm_x'] / mm
+        appendStr = "&mm_x=%.4f&mm_y=%.8f" % (mm, mm)
```
manthey: Why are the format strings different numbers of digits?

histomicstk/annotations_and_masks/annotations_to_masks_handler.py
```diff
@@ -203,63 +211,625 @@ def get_roi_mask(
...
+    # scale up/down annotations by a factor
+    sf, _ = get_scale_factor_and_appendStr(
+        gc=gc, slide_id=slide_id, MPP=kvp['MPP'], MAG=kvp['MAG'])
+    slide_annotations = scale_slide_annotations(slide_annotations, sf=sf)
+
     # get bounding box information for all annotations
     element_infos = get_bboxes_from_slide_annotations(slide_annotations)
 
-    # get indices of rois
+    # get idx of all 'spacial' roi annotations
```
manthey: spatial?

kheffah:
@manthey OK, I've incorporated your corrections. Feel free to approve or suggest others. Thanks!

PR #779, opened 2020-01-13
Only build sphinx docs on Python 3.6

manthey:
The nbformat dependency is now only Python 3.  We could install old versions, but since Python 2 is EOL, it is more sensible to not build docs on travis in Python 2.


manthey:
Merging this to prevent master build from failing.

PR #787, opened 2020-01-15
Added default name for annotation

JellyWang-3:
I ran test.sh -V locally and had 2 tests failed:
10 - puglint_HistomicsTK (Failed)
36 - web_client_HistomicsTK.annotations (Failed)

### DIFF
--- a/web_client/templates/dialogs/saveAnnotation.pug
+++ b/web_client/templates/dialogs/saveAnnotation.pug
@@ -1,3 +1,6 @@
+- var timestamp = moment().format('YYYY-MM-DD HH:mm')
+- var defaultName = 'Annotation '+ timestamp
+
 .modal-dialog(role='document')
   .modal-content
     form.modal-form(role='form')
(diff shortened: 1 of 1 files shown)


manthey:
You've committed two files that shouldn't have been.

The puglint test fails because the edited pug file has no linefeed at the end:
```
10:   > 47|         button.btn.btn-primary.h-submit(type='submit') Save
10: 
10: Missing line feed at file end
```

The annotation test is failing because the test now expects a different value.  We'll need to update the test.

web_client/templates/dialogs/saveAnnotation.pug
```diff
@@ -9,7 +12,7 @@
         .form-group
           label(for='h-annotation-name') Name
           input#h-annotation-name.input-sm.form-control(
-            type='text', placeholder='Enter a name for the annotation', value=annotation.name)
+            type='text', placeholder='', value=defaultName)
```
manthey: If the annotation already has a name, this won't show it.  We probably want to leave the placeholder (since it will show if the user deletes the entire name), and have `value=annotation.name || defaultName`)

manthey: This will likely fix the annotation test.

JellyWang-3: I have made the changes and tested locally, it passed all the tests.
I just pushed again.

JellyWang-3:
Ok. Thank you very much.

Jelly

PR #790, opened 2020-01-16
allow backup callbacks

kheffah:
Allow running function callbacks when backing up annotation jsons. For example, one may decide to create segmentation masks from annotations and to dump them locally. Any callback is allowed, and is specified by the user.


kheffah:
@manthey travis passed, so whenever you're ready. Thanks!

kheffah:
@manthey OK it passes evrything, but not the code coverage. As usual, I do have the test, just not in the `plugin_tests` directory. I think it's ready for review whenever you 're free. Thanks!

PR #792, opened 2020-01-17
Fix polygon merger bug

kheffah:
Handle a couple of edge cases for polygon merger.


kheffah:
@manthey this only fails the code coverage but everything else OK.

histomicstk/annotations_and_masks/masks_to_annotations_handler.py
```diff
@@ -359,6 +359,7 @@ def get_contours_from_mask(
             vertix y coordinated comma-separated values
 
     """
+    assert(MASK.sum() > 2), "Mask is empty!!"
```
manthey: Just a general comment of `assert`:  If you ever run the code under `python -O`, the assert statements become no-ops.  For that reason, I prefer a conditional that raises an exception.

kheffah: @manthey Thanks! I'll keep that in min for future code. I didn't know about the -o and -oo option till now. I suspect if there are asserts in inner loops it makes a lot of sense to replace asserts with exceptions in case someone does use these optimization options.

PR #812, opened 2020-01-23
Hide new annotation button

JellyWang-3:


web_client/templates/panels/annotationSelector.pug
```diff
@@ -67,14 +67,16 @@ block content
...
     label(title='Show annotation labels when hovering with the mouse.')
       input#h-toggle-labels(type='checkbox', checked=showLabels)
       | Labels
     label(title='Highlight annotations when hovering with the mouse.')
       input#h-toggle-interactive(type='checkbox', checked=interactiveMode)
       | Interactive
-    button.btn.btn-sm.btn-primary.h-create-annotation
-      | #[span.icon-plus-squared] New
+    if user && user.get 
+      if accessLevel > 0
```
manthey: I'd rather not have hard-coded constants, since we define this.  It would be better to have this read `if accessLevel >= AccessType.WRITE`, which means that you'll have to pass `AccessType` when you invoke the template.

web_client/templates/panels/annotationSelector.pug
```diff
@@ -67,14 +67,16 @@ block content
...
   .checkbox.h-annotation-toggle
     label(title='Show annotation labels when hovering with the mouse.')
       input#h-toggle-labels(type='checkbox', checked=showLabels)
       | Labels
     label(title='Highlight annotations when hovering with the mouse.')
       input#h-toggle-interactive(type='checkbox', checked=interactiveMode)
       | Interactive
-    button.btn.btn-sm.btn-primary.h-create-annotation
-      | #[span.icon-plus-squared] New
+    if user && user.get
```
manthey: I don't think we need to check if user is defined.  I think the accessLevel will be appropriate if no user is logged in.

web_client/templates/panels/annotationSelector.pug
JellyWang-3: Got it.

web_client/templates/panels/annotationSelector.pug
JellyWang-3: Ok.

PR #803, opened 2020-01-20
Mongo local backup and SQLite Parser

kheffah:
# Mongo local backup and SQLite Parser

**Overview:**

This utility function allows backup of a single folder (and its subfolders) from a girder database locally by recursively pulling folder and item (slide) information +/- the annotations. The information is optionally saved as JSON files and/or entries in an SQLite database. Additionally, the annotations may be parsed into tabular format for easy use and querying and saved as csv files and/or entries in the same SQLite database. The advantage of saving things into SQLite format is that they can be easily queried, and this provides a simple tool for managing projects by periodically using queries to monitor progress, for example. 

The SQLite database can easily be viewed using, for example, an offline sqlite viewer like: https://sqlitebrowser.org/dl/ or even an online sqlite viewer like https://sqliteonline.com/. 

The resultant database has four tables (or two if the user decided not to parse things into annotations):
- **folders**: all girder folders contained within the folder that the user wants to backup. This includes an 'absolute girder path' convenience column. Each has a unique girder ID. 
- **items**: all items (slide). Each has a unique girder ID and is linked to the folders table by a 'girder_folder_id' column. 
- **annotation_docs**: Information about all the annotations . Each has a unique girder ID and is linked to the 'items' table by the 'itemid' column. 
- **annotation_elements**: Information about the annotation elements (polygons, rectangles, points, etc). Each has a unique girder ID and is linked to the 'annotation_docs' table by the 'annotation_girder_id' column. 

**Syntax:**

It is a single function call, as in

```python
import pandas as pd
import sqlalchemy as db
from histomicstk.annotations_and_masks.annotation_database_parser import (
    dump_annotations_locally, parse_annotations_to_local_tables)

# recursively save annotations -- parse to sqlite
dump_annotations_locally(
    gc, folderid=SAMPLE_FOLDER_ID, local=savepath,
    save_json=False, save_sqlite=True,
    callback=parse_annotations_to_local_tables,
    callback_kwargs={
        'save_csv': False,
        'save_sqlite': True,
    }
)

# connect to created database
sql_engine = db.create_engine('sqlite:///%s/Concordance.sqlite' % savepath)
dbcon = sql_engine.connect()

# Now you can query results into a pandas table
result = pd.read_sql_query("SELECT * FROM 'annotation_docs';", dbcon)
```

**Screenshots:**

image

image

image

PR #816, opened 2020-02-02
Annotations to object segmentation

kheffah:
**Overview:**

This includes tools to parse annotations from an item (slide) into masks to use in training and evaluating object localization, classification, and segmentation imaging algorithms (eg. MASK-RCNN). Two "versions" of this workflow exist:

* Get labeled mask for any region in a whole-slide image (user-defined)

* Get labeled mask for areas enclosed within special "region-of-interest" (ROI) annotations that have been drawn by the user. This involves mapping annotations (rectangles/polygons) to ROIs and making one mask per ROI.

The user uses a csv file like the one in
```histomicstk/annotations_and_masks/tests/test_files/sample_GTcodes.csv ```
to control pixel values assigned to mask, overlay order of various annotation groups, which groups are considered to be ROIs, etc. Note that we use the girder definition of term "group" here, which is an annotation style indicating a certain class, such as "tumor" or "necrosis".

**What is the difference between this and annotations_to_masks_handler?!**

The difference between this and version 1, found at
```histomicstk.annotations_and_masks.annotations_to_masks_handler```
is that this (version 2) gets the contours first, including cropping to wanted ROI boundaries and other processing using shapely, and THEN parses these into masks. This enables us to differentiate various objects to use the data for object localization/classification/segmentation tasks. If you would like to get semantic segmentation masks, i.e. you do not really care about individual objects, you can use either version 1 or this handler using the semantic run mode. They re-use much of the same code-base, but some edge cases maybe better handled by version 1. For example, since this version uses shapely first to crop, some objects may be incorrectly parsed by shapely. Version 1, using PIL.ImageDraw may not have these problems.

Bottom line is: if you need semantic segmentation masks, it is probably safer to use version 1 (annotations to masks handler), whereas if you need object segmentation masks, this handler should be used in object run mode.

Be sure to checkout the annotations_to_OBJECT_mask_handler.ipynb jupyter notebook for implementation examples.

**Where to look?**

```
   |_ histomicstk/
   |   |_annotations_and_masks/
   |   |  |_annotation_and_mask_utils.py
   |   |  |_annotations_to_object_mask_handler.py
   |   |_tests/
   |       |_ annotation_and_mask_utils_test.py
   |       |_ annotations_to_object_mask_handler_test.py
   |       |_test_files/
   |          |_sample_GTcodes.csv
   |_ docs/
       |_examples/
          |_annotations_to_OBJECT_mask_handler.ipynb
```


kheffah:
@manthey This should be reviewed and merged after PR #803 . Everything passed, just the code coverage but I have the tests as usual and they pass. 
Cheers!

kheffah:
Will incorporate recent updates to master and make a new PR that's easier to review.

PR #818, opened 2020-02-10
Annotation review gallery

kheffah:
# Rapid annotation review gallery

**Overview:**

Feature demo: https://youtu.be/H5JsC01_3jI

This workflow takes a girder folder contains annotated slides, including
special Region-of-Interest (ROI) annotations that enclose annotated regions.
It downloads visualizations of those ROIs and tiles them into a couple of
galleries to be reviewed rapidly by the pathologist. It also adds a link
that the pathologist can use to access the same ROI in the original slide
to pan around it. See the demo video above for details.

**Where to look:**

```
    |_histomicstk/
    |  |_annotations_and_masks/
    |     |_review_gallery.py
    |     |_tests/
    |        |_review_gallery_test.py
    |_ docs/
       |_examples/
          |_review_gallery.ipynb
```

PR #819, opened 2020-02-13
Update annotation permissions for folder

kheffah:
A convenience utility to update the annotation permissions for:
- individual annotations
- all annotations in a slide
- all annotations in all slides in a folder
Currently, if a user or group is added to a project AFTER some annotations were made, the annotations retain their original permissions even if the folder's permissions now include the newly added user. This is not a bad behavior, but should be remedied at least programatically using an API endpoint or a convenience function. This is a convenience function that handles this. 

Relevant files:
```
histomicstk/utils/girder_convenience_utils.py 
histomicstk/annotations_and_masks/tests/girder_convenience_utils_test.py 
```
This should be merged after PR #818 .


kheffah:
@manthey OK this passed all checks and is ready for merger after PR #818 .

PR #821, opened 2020-02-18
Update annotation styles for folder

kheffah:
A convenience utility to update the annotation styles for:
- individual annotations
- all annotations in a slide
- all annotations in all slides in a folder

Relevant files:
```
histomicstk/utils/girder_convenience_utils.py 
histomicstk/annotations_and_masks/tests/girder_convenience_utils_test.py 
```
This should be merged after PR #819 .


histomicstk/utils/girder_convenience_utils.py
```diff
@@ -74,15 +76,32 @@ def update_permissions_for_annotation(
...
+        annotation_id = annotation['_id']
+    elif annotation_id is None:
+        raise Exception(
+            "You must provide either the annotation or its girder id.")
+
     # get current permissions
     current = gc.get('/annotation/%s/access' % annotation_id)
 
     # add or replace as needed
+    GIRDERBUG = """
```
manthey: Can you make this happen through the UI?  That is, by using the access permissions dialogs for annotations in the Girder items page?  Or is it only through girder client?

kheffah: Nope only through the girder client. I'm writing an issue that also has a video capture of this happening.

manthey: Is this just because you aren't passing the `public` option to the put `access` endpoint?  If unspecified, it looks like `public` gets set to False.  That is, as an admin, can you go to the access control, make it public again, and see the elements again?

kheffah: I'm not sure I understand. Do you mean the annotation needs to be public? If so, why is setting access permissions useful? Also, In GitHub issue #828 there's a video capture. Towards the end it shows that when I download the annotation it has an empty `elements` attribute.

kheffah: @manthey OK I tried that and the annotation elements are still not visible. When I download the annotation, elements are an empty list.

manthey: My question about the public flag was probably irrelevant.  The public flag is a way to allow read access to everyone, but doesn't grant write access in any way.  I was just looking at the rest request made by the UI and how it might differ from what was being passed to girder client.  I'll look further.

kheffah: @manthey Got it. OK let me know if you need anything in terms of specific data, access to specific folders, etc. Thanks for your help with this. I know the repo migration is already a big hassle to deal with.

PR #873, opened 2020-06-23
Change opencv-python to opencv-python-headless

sumanthratna:
- ~changes `large-image[sources]` dependency to specific formats (tiff, openslide, ometiff, pil, and openjpeg)~
- changes opencv-python to opencv-python-headless
- uses [Python.gitignore](https://github.com/github/gitignore/blob/master/Python.gitignore)

UNTESTED, but CI tests passed


manthey:
The drawback to calling for specific sources is that when new sources are added, they have to be manually included, whereas asking for all standard sources via `[sources]` will pick up new sources when they are added to large_image.

Are you excluding GDAL because of its size?  Or is there other reason?

sumanthratna:
Thanks for the reply. The reason I excluded GDAL was because `pip install large-image-source-gdal` requires the GDAL library to be installed. Given that HistomicsTK probably doesn't need compatibility with geospatial data formats, I figured it'd be best to remove this. Size is part of the issue, but it's not that the GDAL library is too large for my disk, it's just that it takes unnecessary disk space as I won't be using it. 

Further, pip can't find any versions for `large-image-source-mapnik`. 

I figured you all might not want these changes, so I've been maintaining my own fork for some time, but I figured that not many people who use HistomicsTK want to install GDAL or mapnik.

sumanthratna:
I will note that you guys may want to add the following (see [this](https://github.com/girder/large_image#modules)):
- large-image-source-openjpeg
- large-image-source-nd2

I'll look more into how large_image works, but do you think it would be possible to just depend on `large-image` (no extras) and require users to install the sources they want to use? The only downside I see to this is that this is a breaking change. Fortunately, it's an easy breaking change to patch.

manthey:
@sumanthratna Since we have wheels for linux, I think a better approach would be to not install source with complex dependencies for non-linux platforms, but still install them for linux.  I've added this as PR #875.  I'd still like to keep your change to opencv-python-headless.

sumanthratna:
That sounds good! @manthey would you like me to remove the changes to large-image in setup.py in this PR? Then, the only change from this PR would be changing opencv-python to opencv-python-headless.

manthey:
@sumanthratna That would be great.  I'll merge #875.

PR #878, opened 2020-07-31
Simplify the Dockerfile.

manthey:
Before, this was based on the dsarchive/base_docker_image, which greatly sped up building the docker image when we built all dependencies in the docker.  Since all of the complex dependencies are now available as wheels, this no longer has much advantage.

We can install fewer system packages, which shrinks the size of the docker a small amount.

We can more easily be explicit about the version of Python we want to use.  Here it has been changed to 3.8.

As a starting point, if you need a newer version of cuda, that can more easily be changed, too.

Additionally, dockcross now supports Python 3.9, but we can't build with it yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,42 +4,78 @@
 #
 # All plugins of HistomicsTK should derive from this docker image
 
-FROM dsarchive/base_docker_image
+FROM nvidia/cuda:9.2-cudnn7-devel-ubuntu18.04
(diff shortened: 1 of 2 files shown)


kheffah:
All make sense. I'm especially on board with the explicit mentioning of python and cuda versions.

kheffah (approved):

kheffah:
As soon as my local docker build works I'll merge this. Just to be safe.

PR #883, opened 2021-01-03
Fix precision error

kheffah:
- Fix precision error that was resulting in slightly different magnifications in x and y axes (makes a difference when the server is called for tiles at a small magnification)
- Parametrize threshold used to check if fetched image and mask correspond to each other


kheffah:
@Leengit @manthey I tested the functions and ran the unit tests locally without problems. It seems that the problem is related to circleci. The change is very minor. I think this is ready to merge, but let me know if you have suggestions for change.

Leengit:
I do not know anything about those circleci errors and am not confident enough to ignore them.  Let's ask @manthey.

manthey:
@kheffah I'll look into what is causing CI to fail (probably tomorrow).  Most likely some upstream package has changed and we need to either pin some package version or adjust some code.

kheffah:
@manthey Great .. thanks :)

manthey:
@kheffah I haven't determined what changed since this last built on CI on Oct. 22.  I get the same failures locally, even when I pull code from my backup snapshot from that date.  The differences seems like some sort of rounding has changed.

I'll make a PR tomorrow that should at least let the tests pass on master with the current behavior.  I'll also set up a weekly build so when something changes in the future it will (hopefully) be easier to track down.

kheffah:
@manthey Sounds like a good plan, thanks very much.

kheffah:
Closing PR because it is a subset of PR #884 to simplify resolving conflicts etc.

PR #884, opened 2021-01-04
Large roi annotations to masks

kheffah:
Allow ``get_all_rois_from_slide()`` to handle very large ROIs by optionally  tiling them .. a simple function parameter controls this behavior.

This should be merged after PR #883 .


Leengit (commented):
Introducing a parameter `tau` is nice.  Given that the previously hard-coded value was 4, will a default value of tau=10 break existing software?

kheffah:
@Leengit Thanks. No, the increase does not affect any of the existing software .. it's only meant to slightly relax the threshold for throwing errors, which was too stringent when I tried smaller magnifications for some of my current work.

kheffah:
@Leengit OK, I think this is ready to merge. Like PR #883 (which is a subset of this one), the tests pass locally, and it seems that the errors are from some circleci issue and arise from ``tests/test_cli_common.py`` and ``tests/test_feature_extraction.py`` .. I didn't change either of those files or any of their dependencies, so I assume this is an old error.

Leengit:
I do not know anything about those circleci errors and am not confident enough to ignore them.  Let's ask @manthey.

histomicstk/annotations_and_masks/annotations_to_masks_handler.py
```diff
@@ -834,23 +943,34 @@ def get_all_rois_from_slide(
...
+            get_kwargs=kvp, monitor=monitorPrefix, verbose=verbose,
+        )
 
     savenames = []
 
-    for roino, idx_for_roi in enumerate(idxs_for_all_rois):
-
-        roicountStr = "%s: roi %d of %d" % (
-            monitorPrefix, roino + 1, len(idxs_for_all_rois))
+    while True:
```
manthey, line 963: Instead of `while True` through the `StopIteration` line, this could be reduce to `for roi_out in roig:`

manthey:
@kheffah I've merged in the removal of Python < 3.6 rather than add code to accommodate those versions.  Aside from one comment to simplify the code, this looks good to me.

histomicstk/annotations_and_masks/annotations_to_masks_handler.py
kheffah: @manthey Thanks! I've incorporated your correction.

kheffah:
> @kheffah I've merged in the removal of Python < 3.6 rather than add code to accommodate those versions. Aside from one comment to simplify the code, this looks good to me.

@manthey Thank you! OK I've incorporate the correction.

manthey (approved):

PR #890, opened 2021-01-11
Remove support for Python 2.7 and 3.5.

### LINKED ISSUE #888: Drop Python 2.7 and 3.5 support
Python 2.7 and 3.5 are both well past end-of-life.  We should release current master as a new version in case anyone needs these older versions.  After that, we should remove these versions from published wheels and stop running CI on them.  We don't need to change any code, but after this we can use newer python features without concern and, eventually, strip out uses of the six library.

manthey:
Closes #888.


manthey:
Merging this will make it easier to resolve #884.

PR #908, opened 2021-06-19
Refactor and extend compute nuclear features

kheffah:
Refactor and extend compute nuclear features to calculate or retain the extra features. With the exception of fractal dimension, all of these features were already calculated by ``skimage.regionprops`` but were not retained. Now they are.

**- Orientation.Orientation :  float**

        Angle between the horizonal axis and the major axis of the ellipse
        that has the same second moments as the region,
        ranging from `-pi/2` to `pi/2` counter-clockwise.

**- Size.ConvexHullArea :  int**
        
        Number of pixels of convex hull image, which is the smallest convex
        polygon that encloses the region.

**- Shape.FractalDimension : float**
 
        Minkowski–Bouligand dimension, aka. the box-counting dimension. It
        is a measure of boundary complexity. See
        https://en.wikipedia.org/wiki/Minkowski%E2%80%93Bouligand_dimension

**- Shape.HuMoments-k : float**

        Where k ranges from 1-7 are the 7 Hu moments features. The first six
        moments are translation, scale and rotation invariant, while the
        seventh moment flips its sign if the shape is a mirror image.
        See https://learnopencv.com/shape-matching-using-hu-moments-c-python/

**- Shape.WeightedHuMoments-k : float**

        Same as Hu moments, but instead of using the binary mask, using the
        intensity image.


histomicstk/features/compute_morphometry_features.py
```diff
@@ -68,78 +83,126 @@ def compute_morphometry_features(im_label, rprops=None):
...
+    From https://en.wikipedia.org/wiki/Minkowski–Bouligand_dimension ...
+    In fractal geometry, the Minkowski–Bouligand dimension, also known as
+    Minkowski dimension or box-counting dimension, is a way of determining the
+    fractal dimension of a set S in a Euclidean space Rn, or more generally in
+    a metric space (X, d).
+
+    """
+    # Only for 2d binary image
+    assert(len(Z.shape) == 2)
+    Z = Z > 0
```
manthey, line 177: In some of rougier's code, this comparison is a threshold (q.v., https://github.com/rougier/fractal_dimension_analysis/blob/main/fractal_analysis_fxns.py#L60).  Should we be following that code more closely, or is there a reason for the difference?

kheffah: @manthey Oh it's just because in the example, the images they are reading are grayscale, with dark background. I think it's more sensible to assume the input images are either binary or grayscale where background is darker. regionprops nuclear mask is binary so this doesn't make a difference anyway so I got rid of the extra unused parameter.

manthey (approved):

PR #912, opened 2021-06-29
Update cellularity_detection_superpixels.py

albernsrya:
enhance the current code in order to make it amenable and congruent with new functional or non-functional requirements or to make it more tolerant to change.
improve code readability


kheffah:
Hi @albernsrya Thank you for your interest in this. Unfortunately the linting (spaces, style, etc) is predetermined by the PEP8 standard, so we cannot really change that. I may have missed it, but are there changes to the functions or arguments, etc?

albernsrya:
thank you for ypur feedback, I fixed it, and refactored the code a bit

kheffah (approved):

PR #917, opened 2021-11-19
Edit bug in accessing attribute "weighted_centroid"

AmarUCLA:
The new scikit-image version now uses the attribute name "centroid_weighted" instead of "weighted_centroid". In addition to this, there were bugs with accesses the attribute as prop.centroid_weighted, so now changed it to accessing the attribute as prop['centroid_weighted']

### DIFF
--- a/histomicstk/segmentation/nuclear/max_clustering.py
+++ b/histomicstk/segmentation/nuclear/max_clustering.py
@@ -74,7 +74,7 @@ def max_clustering(im_response, im_fgnd_mask, r=10):
     # compute object properties
     obj_props = skimage.measure.regionprops(im_label, im_response_nmzd)
 
-    obj_props = [prop for prop in obj_props if np.isfinite(prop.weighted_centroid).all()]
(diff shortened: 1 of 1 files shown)


histomicstk/segmentation/nuclear/max_clustering.py
```diff
@@ -74,7 +74,7 @@ def max_clustering(im_response, im_fgnd_mask, r=10):
     # compute object properties
     obj_props = skimage.measure.regionprops(im_label, im_response_nmzd)
 
-    obj_props = [prop for prop in obj_props if np.isfinite(prop.weighted_centroid).all()]
+    obj_props = [prop for prop in obj_props if np.isfinite(prop['centroid_weighted'].all()]
```
manthey, line 77: You've dropped a parenthesis -- `np.isfinite(prop['centroid_weighted'].all()` should be `np.isfinite(prop['centroid_weighted']).all()`.

PR #918, opened 2021-11-19
Fix bug with accessing "weighted_centroid" attribute

AmarUCLA:
The new scikit-image version now uses the attribute name "centroid_weighted" instead of "weighted_centroid". In addition to this, there were bugs with accesses the attribute as prop.centroid_weighted, so now changed it to accessing the attribute as prop['centroid_weighted']


manthey:
With scikit-image 0.19, the code with "weighted_centroid" is still allowed.  With this change, we need some code to handle either case (so the library still works as long as we support Python 3.6, since the last scikit-image version that works on 3.6 is 0.17).

PR #925, opened 2022-02-23
Add analysis to create superpixel pixelmaps and annotations

naglepuff:
This PR adds a new analysis script called `SuperpixelSegmentation`. It outputs a pixelmap image and an annotation file that can be used to overlay the pixelmap on top of the base image.

### DIFF
--- a/histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
+++ b/histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
@@ -178,7 +178,7 @@ def createSuperPixels(opts):  # noqa
 
     if opts.outputAnnotationFile:
         print(opts.outputAnnotationFile)
-        region = utils.get_region_dict(opts.roi, None, ts)['region']
+        region_dict = utils.get_region_dict(opts.roi, None, ts)
         annotation = {
             'name': 'Superpixel Pixelmap %s' % (
                 os.path.splitext(os.path.basename(opts.outputAnnotationFile))[0]),
@@ -187,8 +187,8 @@ def createSuperPixels(opts):  # noqa
                 'type': 'pixelmap',
                 'girderId': 'outputImageFile',
                 'transform': {
-                    'xoffset': region.get('left', 0),
-                    'yoffset': region.get('top', 0),
+                    'xoffset': region_dict.get('region', {}).get('left', 0),
+                    'yoffset': region_dict.get('region', {}).get('top', 0),
                 },
                 'values': [0] * found,
                 'categories': [{
--- a/histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
+++ b/histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
@@ -1,5 +1,4 @@
 import json
-import logging
 import math
 import os
 
@@ -11,8 +10,6 @@ import skimage
 from histomicstk.cli import utils
 from histomicstk.cli.utils import CLIArgumentParser
 
-logging.basicConfig()
-
 
 def createSuperPixels(opts):  # noqa
     averageSize = opts.superpixelSize ** 2
@@ -20,8 +17,8 @@ def createSuperPixels(opts):  # noqa
     tileSize = opts.tileSize + overlap
     boundaries = opts.boundaries
 
-    logging.info('>> Reading input image')
-    logging.info(opts.inputImageFile)
+    print('>> Reading input image')
+    print(opts.inputImageFile)
 
     ts = large_image.open(opts.inputImageFile)
     meta = ts.getMetadata()
@@ -30,14 +27,16 @@ def createSuperPixels(opts):  # noqa
     tiparams = {}
     tiparams = utils.get_region_dict(opts.roi, None, ts)
 
-    logging.info('>> Generating superpixels')
+    print('>> Generating superpixels')
+    if opts.compactness == 0:
+        print('>> Using SLIC Zero for segmentation')
     for tile in ts.tileIterator(
         format=large_image.constants.TILE_FORMAT_NUMPY,
         tile_size=dict(width=tileSize, height=tileSize),
         tile_overlap=dict(x=overlap, y=overlap),
         **tiparams,
     ):
-        logging.info('%d/%d (%d x %d) - %d' % (
+        print('%d/%d (%d x %d) - %d' % (
             tile['tile_position']['position'], tile['iterator_range']['position'],
             tile['width'], tile['height'],
             found))
@@ -67,26 +66,37 @@ def createSuperPixels(opts):  # noqa
                     mask[suby:suby + submask.shape[0], :submask.shape[1]] *= submask
             n_pixels = numpy.count_nonzero(mask)
         n_segments = math.ceil(n_pixels / averageSize)
-        while True:
+        if opts.compactness == 0:
             segments = skimage.segmentation.slic(
                 img,
                 n_segments=n_segments,
-                compactness=compactness,
+                slic_zero=True,
                 sigma=opts.sigma,
                 start_label=0,
                 enforce_connectivity=True,
-                # max_num_iter=100,
                 mask=mask,
             )
-            # We now have an array that is the same size as the image
             maxValue = numpy.max(segments) + 1
-            if maxValue >= n_segments / 4:
-                break
-            logging.info('Adjusting compactness from %g - got %d rather than %d' % (
-                compactness, maxValue, n_segments))
-            compactness *= 10
+        else:
+            while True:
+                segments = skimage.segmentation.slic(
+                    img,
+                    n_segments=n_segments,
+                    compactness=compactness,
+                    sigma=opts.sigma,
+                    start_label=0,
+                    enforce_connectivity=True,
+                    mask=mask,
+                )
+                # We now have an array that is the same size as the image
+                maxValue = numpy.max(segments) + 1
+                if maxValue >= n_segments / 4:
+                    break
+                print('Adjusting compactness from %g - got %d rather than %d' % (
+                    compactness, maxValue, n_segments))
+                compactness *= 10
         if compactness != opts.compactness:
-            logging.info('Adjusted compactness to %g - got %d' % (
+            print('Adjusted compactness to %g - got %d' % (
                 compactness, maxValue))
         if overlap:
             # Keep any segment that is at all in the non-overlap region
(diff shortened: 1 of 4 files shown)


manthey:
There are currently options for setting colors and labels for categories that don't exist, but no settings for the default (no_category) category.  I think we want to have settings for that (and probably, to correspond with the defaults in HistomicsUI default this category's label to `default`.  The 1/2/3 categories might be better off added just in the UI and not in the algorithm, as the algorithm isn't assigning any categories to the super pixels.

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
```diff
@@ -0,0 +1,225 @@
...
+
+import large_image
+import numpy
+import pyvips
+import scipy
+import skimage
+from histomicstk.cli import utils
+from histomicstk.cli.utils import CLIArgumentParser
+
+logging.basicConfig()
```
manthey, line 14: If you want to use logging rather than prints, then we should change all the print statements (or remove the logging module).

naglepuff:
>  The 1/2/3 categories might be better off added just in the UI and not in the algorithm, as the algorithm isn't assigning any categories to the super pixels.

Just to clarify, would you prefer to axe the 1/2/3 categories from this PR entirely and add the ability to create categories for pixelmap annotations to HistomicsUI? (perhaps in my `pixelmap-interactivity` branch)

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
```diff
@@ -0,0 +1,225 @@
...
+def createSuperPixels(opts):  # noqa
+    averageSize = opts.superpixelSize ** 2
+    overlap = opts.superpixelSize * 4 * 2 if opts.overlap else 0
+    tileSize = opts.tileSize + overlap
+    boundaries = opts.boundaries
+
+    print('>> Reading input image')
+    print(opts.inputImageFile)
+
+    ts = large_image.getTileSource(opts.inputImageFile)
```
manthey, line 26: `getTileSource` and `open` are equivalent, but I think `open` is clearer.

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
```diff
@@ -0,0 +1,167 @@
...
+      <longflag>roi</longflag>
+      <description>Region of interest within which the analysis should be run</description>
+      <channel>input</channel>
+      <default>-1, -1, -1, -1</default>
+    </region>
+    <integer>
+       <name>tileSize</name>
+       <longflag>tileSize</longflag>
+       <label>Working tile size</label>
+       <description>Specify the size of the working tile. If there is no overlap, superpixxel boundaries will appear at these locations.</description>
```
manthey, line 47: s/superpixxel/superpixel/

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
```diff
@@ -0,0 +1,167 @@
...
+      <longflag>overlap</longflag>
+      <description>If specified, overlap tile computation to avoid edge effects.</description>
+      <channel>input</channel>
+      <default>false</default>
+    </boolean>
+    <boolean>
+      <name>boundaries</name>
+      <label>Boundaries</label>
+      <longflag>boundaries</longflag>
+      <description>If specified, draw separate superpixels as boundaries around actual superpixels.</description>
```
manthey, line 71: The boundaries are separate pixelmap categories, but not independent superpixels.  This should be something like `Mark the boundary around each superpixel.  Boundaries are specified with a separate pixelmap index value that is always 1 higher than the pixelmap index value of the superpixel's interior`

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
```diff
@@ -0,0 +1,167 @@
...
+      <channel>input</channel>
+      <default>true</default>
+    </boolean>
+    <float>
+      <name>compactness</name>
+      <label>Compactness</label>
+      <longflag>compactness</longflag>
+      <description>SLIC Algorithm Compactness</description>
```
manthey, line 79: We should either link to or copy the description of compactness and sigma from https://scikit-image.org/docs/dev/api/skimage.segmentation.html?highlight=slic#skimage.segmentation.slic

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
```diff
@@ -0,0 +1,167 @@
...
+      <channel>input</channel>
+      <default>1.0</default>
```
manthey, line 81: skimage defaults to a compactness of 10 and sigma of 0.  We should probably ask @cooperlab what defaults are actually good for histology images.  Visually, a compactness of 0.25 looks better to me, but that is based more on personal aesthetics than on algorithm utility.

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
```diff
@@ -0,0 +1,225 @@
...
+            },
+            {
+                'label': opts.label_3,
+                'fillColor': opts.fillColor_3,
+                'strokeColor': opts.strokeColor_3,
+            },
+        ]
+        region_dict = utils.get_region_dict(opts.roi, None, ts)
+        annotation = {
+            'name': 'Superpixel Pixelmap %s' % (
```
manthey, line 205: This ends up with a long, redundant default name.  We can drop the leading 'Superpixel Pixelmap', as that ends up in the default output name.  I think adding some parameters might be useful (e.g., `'%s - compactness %g sigma %g' % (os.path.splitext(os.path.basename(opts.outputAnnotationFile))[0]), compactness, sigma)`).

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
manthey: @cooperlab Do you have any opinions on this versus the "slic-zero" variant of the algorithm?  We could expose both by using a compactness of 0 to invoke the slic-zero algorithm (if so, internally, we'd have to not pass compactness or use a value of 1 or skimage will fail).

cooperlab: I recommend defaulting to slic-zero. People who know what compatedness means can fool with the parameters. A good compactedness parameter depends entirely on the tissue type.

cooperlab:
> 

Are you referring to the ability to adjust display colors by users through the UI?

cooperlab:
> There are currently options for setting colors and labels for categories that don't exist, but no settings for the default (no_category) category. I think we want to have settings for that (and probably, to correspond with the defaults in HistomicsUI default this category's label to `default`. The 1/2/3 categories might be better off added just in the UI and not in the algorithm, as the algorithm isn't assigning any categories to the super pixels.

Are you referring to the ability to adjust display colors by users through the UI?

manthey:
> > There are currently options for setting colors and labels for categories that don't exist, but no settings for the default (no_category) category. I think we want to have settings for that (and probably, to correspond with the defaults in HistomicsUI default this category's label to `default`. The 1/2/3 categories might be better off added just in the UI and not in the algorithm, as the algorithm isn't assigning any categories to the super pixels.
> 
> Are you referring to the ability to adjust display colors by users through the UI?

Yes.  I think we'll end up with the ability in the UI to change the category of a superpixel to any of the defined groups in the UI -- these have colors associated with them, and we would render the superpixels as those colors.

cooperlab:
> 

Group colors are set by the algorithm, or are defined by users as they do currently for simpler elements?

manthey:
> Group colors are set by the algorithm, or are defined by users as they do currently for simpler elements?

The superpixel algorithm shouldn't set the group colors, since the generation of superpixels doesn't set categories (it sets them all to `0` to start).  The user can define groups and colors in the UI, and these would be applied to superpixels in the same way they can be applied to points and polygons.

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
naglepuff: I've updated the parameters so that the default value for `compactness` is `0`, and the algorithm will interpret that as a cue to use SLICO. I am noticing that this results in very boxy superpixels on my test images. b032305

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
naglepuff: I decided to go with prints.

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
naglepuff: Switched to `open` in 1fe8ac5

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
```diff
@@ -0,0 +1,167 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<executable>
+  <category>HistomicsTK</category>
+  <title>Create Superpixel Pixelmap</title>
```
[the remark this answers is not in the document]
naglepuff: Changed in 1fe8ac5

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
naglepuff: Fixed in 1fe8ac5

histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.xml
manthey: Looking at skimage source, the compactness value is still used with SLIC0.  Specifically, the image is scaled to full dynamic range and then to 1/compactness before being passed to the SLIC or SLIC0 algorithm.  In a quick experiment, the values DO matter, with a compactness of 0.1 following image features much better than the default (10).

I think we need to go back to having a boolean flag for SLIC0, and default the compactness to 1 and the sigma to 0.

naglepuff: I've made the suggested changes and can verify that appears to bean improvement over using `compactness = 0` to invoke SLIC0.
2b71db4

manthey (approved):
As a trivial improvement, we should probably include the compactness and sigma in the annotation name when SLIC 0 is used, but that can be a future change.

PR #984, opened 2023-04-13
Update color_deconvolution.py

Estelle900:
Complete 2,3 w matrix to allow next step of processing, allows deconv with 2 stains without residual specified

### DIFF
--- a/histomicstk/preprocessing/color_deconvolution/color_deconvolution.py
+++ b/histomicstk/preprocessing/color_deconvolution/color_deconvolution.py
@@ -68,6 +68,13 @@ def color_deconvolution(im_rgb, w, I_0=None):
 
     """
     # complement stain matrix if needed
+    if w.shape[1] < 3:
+        wc = np.zeros((w.shape[0], 3))
+        for i in range(w.shape[0]):
+            for j in range(w.shape[1]):
+                wc[i, j] = w[i, j]
+        w = wc
+    
     if np.linalg.norm(w[:, 2]) <= 1e-16:
         wc = complement_stain_matrix(w)
     else:


histomicstk/preprocessing/color_deconvolution/color_deconvolution.py
```diff
@@ -68,6 +68,11 @@ def color_deconvolution(im_rgb, w, I_0=None):
 
     """
     # complement stain matrix if needed
+    if w.shape[1] < 3:
```
manthey, line 71: In looking at this, the docs say `w` is array like, so we should probably also add `w = np.array(w)` before we check the shape.

histomicstk/preprocessing/color_deconvolution/color_deconvolution.py
```diff
@@ -68,6 +68,11 @@ def color_deconvolution(im_rgb, w, I_0=None):
...
+        wc = np.zeros((w.shape[0], 3))
+        wc[:,:w.shape[1]] = w
+        w = wc
+
```
manthey, line 75: And, let's add a test to the `tests/test_color_deconvolution.py` file.  Maybe:
```
    def test_short_array(self):
        im_path = datastore.fetch('Easy1.png')
        im = skimage.io.imread(im_path)[..., :3]

        w = [[0.650, 0.072],
             [0.704, 0.990],
             [0.286, 0.105]]

        conv_result = htk_dcv.color_deconvolution(im, w, 255)

        im_reconv = htk_dcv.color_convolution(conv_result.StainsFloat,
                                              conv_result.Wc, 255)
        np.testing.assert_allclose(im, im_reconv, atol=1)
```

manthey (commented):
Thank you.

PR #996, opened 2023-06-19
 Support frame/style parameter in Nuclei Detection

subinkitware:
1. Updated the code to support color inversion for greyscale images
2. Added a drop-down menu in the UI for receive user input for color inversion.
3. Color de-convolution is only triggered if the user loads an image with more than one channel.


manthey:
It looks like this code doesn't have the frame/style input values.  Was that on another branch?

histomicstk/cli/NucleiDetection/NucleiDetection.xml
```diff
@@ -218,4 +218,17 @@
       <default>1</default>
     </integer>
   </parameters>
+    <parameters advanced="true">
+    <label>Color inversion</label>
+    <description>Choose if color inversion is needed</description>
+    <string-enumeration>
```
manthey, line 224: There is a boolean type which would be more appropriate.  But, we might want a 3-state value: unspecified or "default" or something similar to invert on single band and not on RGB, and explicit true/false or yes/no values to override the default.

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -29,22 +32,41 @@ def detect_tile_nuclei(slide_path, tile_position, args, it_kwargs,
...
-
-    # perform color normalization
-    im_nmzd = htk_cnorm.reinhard(im_tile,
-                                 args.reference_mu_lab,
-                                 args.reference_std_lab,
-                                 src_mu=src_mu_lab,
-                                 src_sigma=src_sigma_lab)
+    # get tile image & check if image is single channel
+    print('The given shape of image is ', tile_info['tile'].shape)
+    if (len(tile_info['tile'].shape) < 2) or (tile_info['tile'].shape[2] <= 1):
```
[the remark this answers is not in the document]
subinkitware: I have updated the code with shortened if function

histomicstk/cli/NucleiDetection/NucleiDetection.xml
subinkitware: I have updated the color inversion tool box to have one more option as you recommended "unspecified". right now unspecified option points to no color inversion.

histomicstk/utils/sample_pixels.py
```diff
@@ -7,7 +7,7 @@
 
 def sample_pixels(slide_path, sample_fraction=None, magnification=None,
                   tissue_seg_mag=1.25, min_coverage=0.1, background=False,
-                  sample_approximate_total=None, tile_grouping=256):
+                  sample_approximate_total=None, tile_grouping=256, flag_color_inversion=False):
```
manthey, line 10: I think the parameter name of `flag_color_inversion` doesn't quite fit the style used elsewhere, and, strictly we are inverting the intensities rather than a color in a single band image, so perhaps a better parameter name would just be `invert_image`.

subinkitware: Updated the variable name to `invert_image`

histomicstk/utils/sample_pixels.py
```diff
@@ -70,7 +70,15 @@ def sample_pixels(slide_path, sample_fraction=None, magnification=None,
         format=large_image.tilesource.TILE_FORMAT_NUMPY,
         scale=scale_lres
     )
-    im_lres = im_lres[:, :, :3]
+
+    # TODO - add code for single channel image
+    print('n\\ >>Low res image shape', im_lres.shape)
+    if len(im_lres.shape) <= 2 or im_lres.shape[2] == 1:
+        im_lres = np.dstack((im_lres, im_lres, im_lres))
+        if flag_color_inversion:
```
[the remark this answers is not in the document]
subinkitware: I agree, the best way to go will be to apply the inversion regardless of the number of bands. I have updated the code to have the inversion for all kinds of images. I haven't removed any TODO yet... since there is one more route the images take when is_wsi is not triggered. I will need to update that code also to have a complete effect for the image inversion.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
```
manthey, line 1: Did you mean to check in some of these files?   They probably should be added to the .gitignore (this and MANIFEST.in).

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -3,6 +3,7 @@
 import os
 import time
 
+import dask
```
manthey, line 6: One of the reasons this wasn't imported at the top level is that it is a heavy import (that is, it takes time), so users of the library that aren't using dask end up with a slower start.  It should be moved back into the function where it gets used.

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -17,20 +18,53 @@
...
+    print('\n >> is it wsi ? {}, is it whole image ? {}'.format(is_wsi, process_whole_image))
+
+    return ts, is_wsi
+
+
+def image_inversion_flag_setter(args=None):
+    invert_image = False
+    if args.ImageInversionForm == "Yes":
+        invert_image = True
+    if args.ImageInversionForm == "No" or args.ImageInversionForm == "unspecified":
```
manthey, line 41: I think rather than "unspecified", this should have a "default" mode which, when a single band image will invert and an RGB image would not invert.  By making that the default, for the case we are pursuing the inversion will happen on our fluorescence channel frame but not on our H&E images.

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -154,8 +304,12 @@ def main(args):
 
     it_kwargs = {
         'tile_size': {'width': args.analysis_tile_size},
-        'scale': {'magnification': args.analysis_mag},
+        'scale': {'magnification': args.analysis_mag}
     }
+    try:
+        it_kwargs['frame'] = int(args.frame)
```
manthey, line 310: We should be able to get rid of this try/except block by doing 
```
if getattr(args, 'frame', None) is not None:
    it_kwargs['frame'] = int(args.frame)
```
This will allow legitimate errors to still throw (such as someone oddly passing a non-number to frame).

histomicstk/cli/NucleiDetection/NucleiDetection.xml
```diff
@@ -218,4 +218,36 @@
...
+    <label>Image inversion</label>
+    <description>Choose if image inversion is needed</description>
+    <string-enumeration>
+      <name>ImageInversionForm</name>
+      <description>Image inversion may be needed for greyscale images with bright nuclei and dark background. Choose if color inversion is needed.</description>
+      <channel>input</channel>
+      <longflag>output_form</longflag>
+      <element>Yes</element>
+      <element>No</element>
+      <element>unspecified</element>
```
manthey, line 249: See my comments elsewhere -- I think the default should be `default` rather than `unspecified` with some logic based on number of bands.

histomicstk/cli/NucleiDetection/__init__.py
```diff
@@ -0,0 +1,5 @@
+__all__ = (
+
+    # functions and classes of this package
+    'NucleiDetection'
+)
```
manthey, line 5: Is there a reason you added this?  By default, an empty `__init__.py` should have exposed the other python files in that folder.

histomicstk/utils/sample_pixels.py
```diff
@@ -149,7 +161,16 @@ def _sample_pixels_tile(slide_path, iter_args, positions, sample_fraction,
             continue
 
         # get current tile image
-        im_tile = tile['tile'][:, :, :3]
+        # TODO - check if single channel
```
manthey, line 164: We'll want to remove the `TODO` part of these comments since they are done.

histomicstk/utils/simple_mask.py
```diff
@@ -184,7 +184,6 @@ def estimate_variance(x, y, peak):
     SimpleMask
 
     """
-
```
manthey, line 187: I realize this is a trivial change, but let's undo it to reduce the code difference.

histomicstk/cli/NucleiDetection/NucleiDetection.py
subinkitware: The import dask is moved to `detect_nuclei_with_dask` function. After the change, it is tested on some sample images to make sure it is working. On the first run it is taking some time to load but from the second time the execution is faster.

histomicstk/cli/NucleiDetection/NucleiDetection.py
subinkitware: I tried this method, but the `getattr()` cannot have None assigned to it's output. So, i've changed the code a bit to check if `args.frame` exist and is a digit. Otherwise it throws an exception.
`if not args.frame or args.frame.startswith('{#control'):
        args.frame = None
    elif not args.frame.isdigit():
        raise Exception("The given frame value is not an integer")
    else:
        it_kwargs['frame'] = args.frame`

histomicstk/cli/NucleiDetection/__init__.py
subinkitware: This code was added during the debugging process. I will remove this now and convert the `__init__.py` into an empty file.

histomicstk/utils/sample_pixels.py
subinkitware: All the TODO's and some print functions used for debugging are removed from the code.

histomicstk/cli/NucleiDetection/NucleiDetection.py
subinkitware: I have added the default mode to frame and style. As you mentioned before, If it's a single-channel image the default option will invert the image and an RGB image would not invert. I have tested them on sample RGB and single-channel images.

histomicstk/cli/NucleiDetection/NucleiDetection.xml
```diff
@@ -218,4 +218,36 @@
       <default>1</default>
     </integer>
   </parameters>
+  <parameters advanced="true">
```
manthey, line 221: As a minor change, can we move these new values to above the Dask block?

manthey: The rationale is that we have the Dask block last on all of the CLI that have it.

histomicstk/cli/utils.py
```diff
@@ -49,7 +49,8 @@ def get_stain_matrix(args, count=3):
     return np.array([get_stain_vector(args, i + 1) for i in range(count)]).T
 
 
-def segment_wsi_foreground_at_low_res(ts, lres_size=2048):
+def segment_wsi_foreground_at_low_res(
+        ts=None, lres_size=2048, invert_image=False, frame=None, default_img_inversion=False):
```
manthey, line 53: We _have_ to pass `ts`, so the default shouldn't be `None`.

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -76,14 +121,141 @@ def detect_tile_nuclei(slide_path, tile_position, args, it_kwargs,
...
 
 
-def main(args):
+def process_wsi_as_whole_image(ts, invert_image=False, args=None, default_img_inversion=False):
+    print('\n>> Computing tissue/foreground mask at low-res ...\n')
+
+    start_time = time.time()
+    # segment wsi foreground at low resolution
+    im_fgnd_mask_lres, fgnd_seg_scale = \
+        cli_utils.segment_wsi_foreground_at_low_res(args,
```
manthey, line 130: This doesn't take `args` as a parameter.

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -76,14 +121,141 @@ def detect_tile_nuclei(slide_path, tile_position, args, it_kwargs,
...
+        'tile_size': {'width': args.analysis_tile_size},
+        'scale': {'magnification': args.analysis_mag}
+    }
+
     total_start_time = time.time()
 
     print('\n>> CLI Parameters ...\n')
 
-    print(args)
+    print(args, type(args))
```
manthey, line 258: args will be a Namespace; it probably isn't useful to report this in general, so this could be reverted to just `print(args)`.

manthey: Though maybe it'd be better to `pprint.pprint(vars(args))`

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -17,20 +17,58 @@
...
     # get requested tile
     tile_info = ts.getSingleTile(
         tile_position=tile_position,
         format=large_image.tilesource.TILE_FORMAT_NUMPY,
         **it_kwargs)
 
-    # get tile image
-    im_tile = tile_info['tile'][:, :, :3]
+    # get tile image & check number of channels
+    print('The given shape of image is ', tile_info['tile'].shape)
```
manthey, line 64: I'd rather remove this print statement; it makes the job record harder to look through.

manthey:
I made a test file we can use: https://data.kitware.com/#item/649d7b7a5121da8eb1a8144c
This is an extract of TCGA-02-0010-01Z-00-DX4.07de2e55-a8fe-40ee-9e98-bcb78050b9f7.svs.  Specifically, extracted from `ts.getRegion(region=dict(left=1000,top=16000,width=4000,height=3000),format='numpy')[0]`.  It has six frames with that region, all single-band greyscale.  These are, in order, 0: inverted deconvolved hematoxylin, 1: deconvolved hematoxylin, 2: deconvolved eosin, 3: original red band, 4: original green band, 5: original blue band.

This should be useful for testing, since it should be small enough to run the complete nuclei detection on it in a reasonable amount of time.

Specifically, we can run the original algorithm on it with `frame=3, style={"bands":[{"frame":3,"palette":"#f00"},{"frame":4,"palette":"#0f0"},{"frame":5,"palette":"#00f"}]}` (with and without an ROI to test the two code paths).  We can also test the single-band code paths with and without inversion on the hematoxylin channels.  I'd expect `frame=0` to produce the same results as the original and `frame=1, invert=no` to also do so.

histomicstk/cli/utils.py
subinkitware: Great suggestion, I missed that one. I've changed the default value for ts.

histomicstk/cli/NucleiDetection/NucleiDetection.py
subinkitware: The print args were there on the primary code repo. I kind of find it useful when running the job to have a quick check if all the parameters I've given are good. As you suggested pprint makes more sense for printing the params.

manthey:
And, for test consistency, the same region is extracted as the original RGB: https://data.kitware.com/#item/649d939c5121da8eb1a81453

subinkitware:
> I made a test file we can use: https://data.kitware.com/#item/649d7b7a5121da8eb1a8144c This is an extract of TCGA-02-0010-01Z-00-DX4.07de2e55-a8fe-40ee-9e98-bcb78050b9f7.svs. Specifically, extracted from `ts.getRegion(region=dict(left=1000,top=16000,width=4000,height=3000),format='numpy')[0]`. It has six frames with that region, all single-band greyscale. These are, in order, 0: inverted deconvolved hematoxylin, 1: deconvolved hematoxylin, 2: deconvolved eosin, 3: original red band, 4: original green band, 5: original blue band.
> 
> This should be useful for testing, since it should be small enough to run the complete nuclei detection on it in a reasonable amount of time.
> 
> Specifically, we can run the original algorithm on it with `frame=3, style={"bands":[{"frame":3,"palette":"#f00"},{"frame":4,"palette":"#0f0"},{"frame":5,"palette":"#00f"}]}` (with and without an ROI to test the two code paths). We can also test the single-band code paths with and without inversion on the hematoxylin channels. I'd expect `frame=0` to produce the same results as the original and `frame=1, invert=no` to also do so.

I was also expecting the results from `frame=0, invert_image=True` and `frame=1,invert_image=False` to have the same result. But, there is a huge difference in the number of nuclei. The output is 7400 and 670 respectively. 

Adding the `style` component is slowing down the test code. I'm trying to figure out what is the issue there.

manthey:
> I was also expecting the results from `frame=0, invert_image=True` and `frame=1,invert_image=False` to have the same result. But, there is a huge difference in the number of nuclei. The output is 7400 and 670 respectively.

Yes -- they should be the same, since frame 0 and frame 1 are literally inverses of each other.

subinkitware:
> style={"bands":[{"frame":3,"palette":"#f00"},

The issue was that the image inversion` was called after `color normalization`. That created some kind of tiling which artificially increased the number of nuclei. It was resolved by moving the inversion function before the color normalization. Even though the updated code passes all tests, It fails when I test a greyscale WSI without any cropping separately using the UI. I'm looking into that issue now. I would try adding some test code to invoke the functions inside the `is_wsi and process_whole_image:`

tests/test_nuclei_segmentation.py
```diff
@@ -118,3 +120,155 @@ def test_segment_nuclei_guassian_voting(self):
...
+                                              -1.0],
+                              stain_3='null',
+                              stain_3_vector=[-1.0,
+                                              -1.0,
+                                              -1.0],
+                              style=None)
+
+    def test_image_inversion_flag_setter(self):
+        invert_image = nucl_det.image_inversion_flag_setter(self.args)
+        assert invert_image, True
```
manthey, line 164: I think you want this to be 
```
invert_image,  default_img_inversion = nucl_det.image_inversion_flag_setter(self.args)
assert invert_image == True
assert default_img_inversion = False
```
As written, `invert_image` will be a tuple and assert checks that the (tuple, True) is not False, which isn't the test that was intended.

tests/test_nuclei_segmentation.py
```diff
@@ -118,3 +120,155 @@ def test_segment_nuclei_guassian_voting(self):
...
+            invert_image=True,
+            is_wsi=is_wsi,
+            src_mu_lab=None,
+            src_sigma_lab=None)
+        np.testing.assert_allclose(len(nuclei_list), 7497, 1e+2)
+
+    def test_nuclei_detection_frame_style(self):
+
+        # retrieve image from datastore
+        input_image_path = os.path.join(datastore.fetch('tcgaextract_ihergb.tiff'))
```
manthey, line 221: This (also elsewhere), the `os.path.join()` outer function isn't doing anything -- `datastore.fetch` returns a single path and joining a single path is a no-operation.

tests/test_nuclei_segmentation.py
subinkitware: Yeah, I agree. This test code was written when the function was only returning a boolean value. Then I added the second return default_img_inversion. I missed updating this test function also since it was giving me the test PASSED result.

tests/test_nuclei_segmentation.py
subinkitware: I agree, all these functions are updated now.

manthey:
We can test the whole NucleiDetection workflow with our test files.  Let's add this test file https://gist.github.com/manthey/39f5f37a023e9cd63bcd9d6f473db3f8 (add it to the `tests` directory).  You'll need to add the appropriate test files in the datastore.  I have some comments on the end -- we might want to add the files that don't have a magnification set and make the roi/wsi variants for each.

manthey:
I added a comment on that gist that shows a way to parameterize the tests.

histomicstk/cli/NucleiDetection/NucleiDetection.xml
```diff
@@ -192,6 +192,38 @@
...
+    </string>
+  </parameters>
+  <parameters advanced="true">
+    <label>Image inversion</label>
+    <description>Choose if image inversion is needed</description>
+    <string-enumeration>
+      <name>ImageInversionForm</name>
+      <description>Image inversion may be needed for greyscale images with bright nuclei and dark background. default option will automatically invert the image if it is sinlge channel. Choose if color inversion is needed.</description>
+      <channel>input</channel>
+      <longflag>output_form</longflag>
```
manthey, line 221: We probably want to rename this, as it isn't related to the output.  Maybe `invert_image` instead of `output_form`.

histomicstk/cli/NucleiDetection/NucleiDetection.xml
```diff
@@ -192,6 +192,38 @@
...
+      <label>Style Options</label>
+      <description>Image style options for compositing a multi-frame image</description>
+      <default>{#control:#current_image_style#}</default>
+    </string>
+  </parameters>
+  <parameters advanced="true">
+    <label>Image inversion</label>
+    <description>Choose if image inversion is needed</description>
+    <string-enumeration>
+      <name>ImageInversionForm</name>
```
manthey, line 218: This block should also have a `<label>` element (maybe `<label>Invert Image</label>`).

histomicstk/cli/utils.py
```diff
@@ -65,10 +66,21 @@ def segment_wsi_foreground_at_low_res(ts, lres_size=2048):
...
+    if len(im_lres.shape) <= 2 or im_lres.shape[2] == 1:
+        im_lres = np.dstack((im_lres, im_lres, im_lres))
+        if default_img_inversion:
+            invert_image = True
+    else:
+        im_lres = im_lres[:, :, :3]
+
+    # perform image inversion
+    if invert_image:
+        im_lres = np.max(im_lres) - im_lres
```
manthey, line 83: At least some of the issues we've seen is because, when we composite images using a style parameter, it can return the image as a numpy array with a dtype of `float64` even though the data is intrinsically `uint8`.  `im_lres` in this case is `float64`, and the `simple_mask` function calls `skimage.rgb2gray`, which, if the data type is float, assumes the values are scaled from [0,1] and produces erroneous results.

Adding something like:
```
    if issubclass(im_lres.dtype.type, np.floating) and np.max(im_lres) > 1:
        if np.min(im_lres) >= 0 and np.max(im_lres) < 256:
            im_lres = im_lres.astype(np.uint8)
        if np.min(im_lres) >= 0 and np.max(im_lres) < 65536:
            im_lres = im_lres.astype(np.uint16)
```
would fix this for the conditions we care about.

subinkitware: I tried a similar strategy yesterday. There was some warning that the `np.exp()` was overloaded. I guess it was because we were sending float64 values to it and it had to do a lot of calculations to give the output. So I tried casting to `im_lms = np.exp(im_lms.astype(np.uint8))`. It was able to detect nuclei but was giving erroneous results. This looks like a better way to cast even before it goes to the exponential function.

manthey:
I think some of the other issues we have are when we are using a single channel source and the whole image (invoked when we know the magnification, which probably shouldn't be a determinant, but that is an issue for a different PR).  Specifically, we use reinhard_stats to compute a average color adjustment that gets applied across all tiles for consistency.  If we don't do this, we adjust each tile in turn.  This often produces odd results on a greyscale image, as it was intended for an H&E image.

Then, by default we perform a color deconvolution.  If we do this on greyscale images, it actually adjusts the 0th band and we get different results than if we don't perform it.  Since the algorithms were tuned for this process, we probably still want to do this and revisit the greyscale path later -- ideally we need to characterize the stats of a hematoxylin layer and use that as the target adjustment instead of the current adjustment followed by deconvolution.

manthey:
I think this means that our code path would be to (a) skip compute_reinhard_norm for single band images, (b) always perform deconvolution.  We would then create an issue to revisit reinhard stats and single band processing.

manthey (approved):

PR #999, opened 2023-06-28
998 record parameters and primary results in annotation metadata

subinkitware:


manthey (approved):

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -383,8 +383,11 @@ def main(args):
         os.path.basename(args.outputNucleiAnnotationFile))[0]
 
     annotation = {
-        'name': annot_fname + '-nuclei-' + args.nuclei_annotation_format,
-        'elements': nuclei_list
+        'name': 'NucleiDetection',
+        'src_mu_lab': src_mu_lab,
```
manthey, line 387: For the annotation to validate, we want these additional fields to be in the `attributes` key.  For consistency with PositivePixelCount, we should call the input parameters just `params`.  And, let's not change the output name as it has some useful information in it.  That is:
```
   annotation = {
        'name': annot_fname + '-nuclei-' + args.nuclei_annotation_format,
        'elements': nuclei_list,
        'attributes': {
            'src_mu_lab': src_mu_lab,
            'src_sigma_lab': src_sigma_lab,
            'params': vars(args),
        },
}
```

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -384,7 +384,12 @@ def main(args):
 
     annotation = {
         'name': annot_fname + '-nuclei-' + args.nuclei_annotation_format,
-        'elements': nuclei_list
+        'elements': nuclei_list,
+        'attributes': {
+            'src_mu_lab': src_mu_lab,
```
manthey, line 389: It looks like `src_mu_lab` and `src_sigma_lab` are either `None` or a numpy array.  As a numpy array, they don't json serialize, so we probably have to do something like `'src_mu_lab': None if src_mu_lab is None else src_mu_lab.to_list(),`

subinkitware: Yeah, they are a numpy array with 3 elements. I update these changes.

manthey (approved):

PR #1005, opened 2023-07-10
casting image dtype for consistant scaling.

### LINKED ISSUE #1002: Single band nuclei detection should use consistent scaling across image
One of the reasons we want to calculate a whole-image scaling parameter is to avoid processing differences on different tiles.

subinkitware:
resolving this fixes #1002

### DIFF
--- a/histomicstk/cli/NucleiDetection/NucleiDetection.py
+++ b/histomicstk/cli/NucleiDetection/NucleiDetection.py
@@ -70,6 +70,13 @@ def detect_tile_nuclei(slide_path, tile_position, args, it_kwargs,
     else:
         im_tile = tile_info['tile'][:, :, :3]
 
+    # cast the float dtype to uint16
+    if issubclass(im_tile.dtype.type, np.floating) and np.max(im_tile) > 1:
+        if np.min(im_tile) >= 0 and np.max(im_tile) < 256:
+            im_tile = im_tile.astype(np.uint8)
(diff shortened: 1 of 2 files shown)


manthey:
Do you have a specific example where this changes the results?  That is, is there a test we can add or I can run locally to see the behavior change?

subinkitware:
> Do you have a specific example where this changes the results? That is, is there a test we can add or I can run it locally to see the behavior change?

The additional casting doesn't change the results. I have tested a few RGB, Single channel, and multichannel images. All their `im_tile` array are already in `unint8` I'm not sure if we encounter float as an `im_tile` datatype in a WSL file. If it doesn't, we can remove this casting code.

subinkitware:
> Do you have a specific example where this changes the results? That is, is there a test we can add or I can run locally to see the behavior change?

Inside the `detect_nuclei_with_dask` function there are already methods that check if it is a single channel/float datatype and converts the image accordingly. So I believe the data will be consistent in the case of single-channel images.  We might not need to add additional code for data consistency.

subinkitware:
- I have used the large_image.getThumbnail() get generate a low-res image of the WSI to the size 1024

@manthey - The Cli `ComputeNucleiFeatures` also uses the mu and sigma calculation of the lab space for data processing. So, can I also make this similar updates on that files in this same PR?

subinkitware:
A follow-up PR #1043 has been generated to abstract the nuclei detection code.

PR #1007, opened 2023-07-12
Macenko pca out of bounds fix

### LINKED ISSUE #1006: Potential bug in separate_stains_macenko_pca.py | Function: argpercentile
Hi!

First off, thanks to all developers and maintainers for this amazing package!

I am using HistomicsTK on Ubuntu 20.04.5 LTS.

During my work with it I noticed a potential bug in the file `separate_stains_macenko_pca.py' in the function argpercentile.

https://github.com/DigitalSlideArchive/HistomicsTK/blob/15262df548a2579846a0dcfd8b7f41e30ddff58e/histomicstk/preprocessing/color_deconvolution/separate_stains_macenko_pca.py#L121C1-L125C41

For instance, this function is used in line 98 with a value for p = 0.99:
https://github.com/DigitalSlideArchive/HistomicsTK/blob/4748f9d2e00216a59056c1a75c87f65dc6b271aa/histomicstk/preprocessing/color_deconvolution/separate_stains_macenko_pca.py#L98C4-L98C56

With p = 0.99, as soon as the size of the provided array is smaller than 51, the calculated index will be larger than the length of the array. This leads to an "out of bounds" error in the subsequent call of numpy.argpartition(arr, i)[i] on line 125.

arr.size = 50, p = 0.99
i = 50 -> Will cause an error

I guess the easiest fix would be something like this:
```
i = int(p * arr.size + 0.5)
if i >= arr.size:
    i -= 1
return numpy.argpartition(arr, i)[i]
```

Having said all of this, I honestly am not sure if this combination of a small input array and a high percentage value is not supposed to occur in a normal application flow. Therefore, I am uncertain if this would merely mask another underlying problem.

Thank you for your time!

m101010:
fixes #1006

### DIFF
--- a/tests/test_color_deconvolution.py
+++ b/tests/test_color_deconvolution.py
@@ -2,7 +2,8 @@ import numpy as np
 import skimage.io
 
 from histomicstk.preprocessing import color_deconvolution as htk_dcv
-from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import argpercentile as htk_ap
+from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import \
+    argpercentile as htk_ap
 
 from .datastore import datastore
 
--- a/tests/test_color_deconvolution.py
+++ b/tests/test_color_deconvolution.py
@@ -2,6 +2,7 @@ import numpy as np
 import skimage.io
 
 from histomicstk.preprocessing import color_deconvolution as htk_dcv
+from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import argpercentile as htk_ap
 
 from .datastore import datastore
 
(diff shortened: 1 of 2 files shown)


tests/test_color_deconvolution.py
```diff
@@ -2,6 +2,7 @@
 import skimage.io
 
 from histomicstk.preprocessing import color_deconvolution as htk_dcv
+from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import argpercentile as htk_ap
```
manthey, line 5: The linter complains.  Running `tox -e format` will reformat this line to:
```
from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import \
    argpercentile as htk_ap
```

m101010, line 5:
```suggestion
from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import \
    argpercentile as htk_ap
```

m101010:
```suggestion
from histomicstk.preprocessing.color_deconvolution.separate_stains_macenko_pca import \
    argpercentile as htk_ap
```
Ok - so if I understand correctly, we can suggest a change directly to the pull request

manthey:
Thank you.

manthey (approved):

PR #1009, opened 2023-07-12
added tile overlap parameter

subinkitware:


histomicstk/cli/NucleiDetection/NucleiDetection.xml
```diff
@@ -159,6 +159,13 @@
       <description>Minimum area that each nucleus should have</description>
       <longflag>min_nucleus_area</longflag>
       <default>80</default>
+    </double>
```
manthey, line 162: If this is in pixels, it really should be an integer.  I'd move it to immediately after the `analysis_tile_size` parameter.  (And that parameter, too, really should be an integer).

subinkitware: Yes, I agree. I checked the function which takes this as an argument. It was also expecting an integer. I've updated the code with these changes.

subinkitware:
This new updated code solves some of the edge cases that happen when the tile size is smaller than the given overlap size. When it is smaller, a portion of the bottom and right edges nuclei are not getting segmented since the algorithm removes the mask present in that region.  Now, I kept a check if the tile size is smaller than the given overlap size there will not be any overlap removal.

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -276,9 +267,19 @@ def main(args):
...
-    # retrive style and frame
+    # retrive style
     if not args.style or args.style.startswith('{#control'):
         args.style = None
+
+    # initial arguments
+    it_kwargs = {
+        'tile_size': {'width': args.analysis_tile_size},
+        'scale': {'magnification': args.analysis_mag},
+        'tile_overlap': {'x': args.tile_overlap_value, 'y': args.tile_overlap_value, 'edges': True},
```
manthey, line 278: I don't think there is any benefit in using `edges: True`.  You'll end up with more tiles, but not usefully as the additional area will largely get trimmed or duplicated.

histomicstk/segmentation/label/delete_overlap.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    # Compute the half of the overlap values
+    left_overlap = max(overlap_info['left'] // 2, 0)
+    right_overlap = max(overlap_info['right'] // 2, 0)
+    top_overlap = max(overlap_info['top'] // 2, 0)
+    bottom_overlap = max(overlap_info['bottom'] // 2, 0)
+
+    # Check if the tile size is smaller than 1.5 times the overlap in height or width
+    if tile_size['height'] < overlap_info['top']  \
```
manthey, line 36: I don't understand where this is necessary.  You can't specify an overlap as large as the tile, so there will always be some core tile.  If we are discarding nuclei that are completely in the overlap and the overlap is larger than the largest nuclei, we should never discard too much.  If we discard nuclei that are partially in the outer half of the overlap and the overlap is larger than twice the largest nuclei, we should never discard too much.  Maybe I'm missing something.

histomicstk/segmentation/label/delete_overlap.py
```diff
@@ -0,0 +1,52 @@
...
+    bottom_overlap = max(overlap_info['bottom'] // 2, 0)
+
+    im_label_del = np.zeros_like(im_label)
+
+    if np.any(im_label):
+
+        # Create a border mask based on the overlap values
+        im_border_mask = np.zeros_like(im_label)
+        im_border_mask[:left_overlap, :] = left_overlap > 0
+        im_border_mask[-right_overlap:, :] = right_overlap > 0
```
manthey, line 38: I don't think this will do what you want when `right_overlap` is 0.  In this case, you're doing `im_border_mask[0:, :] = False`, which is going to undo the `left_overlap` mask.  Further, when you do the same to the `bottom_overlap` where it is 0, it will undo all of the overlap masking.

I think you need to do:
```
if right_overlap:
  im_border_mask[-right_overlap:, :] = True
```
for right and bottom.

histomicstk/segmentation/label/delete_overlap.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Returns:
+        ndarray: Image label with overlapping regions deleted.
+    Note:
+        This function assumes the necessary imports, such as `np`, are already present.
+
+    """
+
+    # Compute the half of the overlap values
+    left_overlap = max(overlap_info['left'] // 2, 0)
```
manthey, line 26: I don't think the overlap info can ever have negative values.  Is the `max` necessary?

histomicstk/segmentation/label/delete_overlap.py
```diff
@@ -0,0 +1,52 @@
...
+    right_overlap = max(overlap_info['right'] // 2, 0)
+    top_overlap = max(overlap_info['top'] // 2, 0)
+    bottom_overlap = max(overlap_info['bottom'] // 2, 0)
+
+    im_label_del = np.zeros_like(im_label)
```
manthey, line 31: We should move this to just before it is used.  No need to allocate memory until it is useful.

histomicstk/segmentation/label/delete_overlap.py
```diff
@@ -0,0 +1,52 @@
...
+        border_indices = np.unique(im_label[im_border_mask > 0])
+        border_indices = border_indices[border_indices > 0]
+
+        if len(border_indices) == 0:
+            return im_label
+
+        # Condense and delete the border regions from the image label
+        im_label_del = condense(delete(im_label, border_indices))
+
+    return im_label_del
```
manthey, line 52: It reaches here if `np.any(im_label)` is false.  Should we just return `im_label` in that case? (rather than making a new zero array and returning it)

histomicstk/segmentation/label/delete_overlap.py
subinkitware: I've made the changes you mentioned above. Also the order of the elements is invested since it is `[row, column]` in the NumPy array.

histomicstk/segmentation/label/delete_overlap.py
subinkitware: Great observation, I've made the changes according to your comments. The same allocation also happened in the `delete_border.py` code. So, I've also changed that too.

histomicstk/segmentation/label/delete_overlap.py
subinkitware: I agree, I've updated this to the `delete_border.py` code also.

.gitignore
```diff
@@ -92,6 +92,8 @@ HistomicsTKDocker.code-workspace
 histomicstk/Debugging
 .devcontainer/
 MANIFEST.in
+.pre-commit-config.yaml
+run_tests_code_cleanup.sh
```
manthey, line 96: I merged a change which doesn't require explicit clean up between runs of `tox -e docs` (though it still is necessary when testing different versions of python because scikit-build leaves artifacts).  You might not need your local script any more.  (You can leave this here, though).

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -106,6 +96,12 @@ def detect_tile_nuclei(slide_path, tile_position, args, it_kwargs,
 
         im_nuclei_seg_mask = htk_seg_label.delete_border(im_nuclei_seg_mask)
 
+    # Delete overlapping border nuclei
+    if any(tile_info.values()) > 0:
```
manthey, line 100: Did you mean `if any(tile_info['tile_overlap'].values()) > 0:`?

subinkitware: Yes. I've made that changes in the code but forgot to push it. That was the only addition to the code.

histomicstk/segmentation/label/delete_border.py
```diff
@@ -41,6 +39,9 @@ def delete_border(im_label):
         if len(border_indices) == 0:
             return im_label
 
+        # Condense and delete the border regions from the image label
+        im_label_del = np.zeros_like(im_label)
         im_label_del = condense(delete(im_label, border_indices))
-
-    return im_label_del
+        return im_label_del
+    else:
```
manthey, line 46: Because the previous block returns, the `else` is unnecessary.  The code flow might be easier to read if we did the converse and then more of the code would be outdented:
```
if not np.any(im_label):
    return im_label
im_border_mask = ...
...
```

.gitignore
subinkitware: That's great, I wrote this bash script to do other things too. Otherwise, I'll have to go around and find the exact line of code to run. It does testing, cleaning, deployment, and autopep8 to a specific location if the tox -e format doesn't work.
`#!/bin/bash
set +e

# Prompt the user to enter the first string
read -p "Select the type a-all, c-clean, d-deploy, p-autopep8, t-test: " VAR

# Function: Run tests
run_tests() {
  echo "------------> Running nuclei detection test"
  tox -e py310 -- -k test_nuclei_segmentation
  echo "------------> Running cli results test"
  tox -e py310 -- -k test_cli_results
  echo "------------> All Good"
}

# Function: Run cleanup
run_cleanup() {
  echo "------------> clean the code for erros"
  tox -e format
  echo "------------> Run Flake code checking"
  tox -e flake8
  echo "------------> Run pre-commit code check"
  pre-commit run --all-files
  echo "------------> All Good"
}

# Function: Run deployment
run_deploy() {
  echo "------------> Deploying docker"
  docker build -f Dockerfile -t sek_histomicstk:2.0 .
  echo "------------> All Good"
}

# Function: Run deployment
run_autopep8() {
  read -p "provide relative path to file: " VAR2
  autopep8 "$VAR2" --in-place --pep8-passes 2000 --verbose
  echo "------------> All Good"
}

# Function: Run everything
run_everything() {
  run_tests
  run_cleanup
  run_deploy
}

if [ "$VAR" == "t" ]; then
  run_tests
fi

if [ "$VAR" == "c" ]; then
  run_cleanup
fi

if [ "$VAR" == "d" ]; then
  run_deploy
fi

if [ "$VAR" == "p" ]; then
  run_autopep8
fi

if [ "$VAR" == "a" ]; then
  run_everything
fi
`

manthey: I'm also a big fan of having a local bash script to manage my projects -- I usually have them outside of the repo directory and in my path.

subinkitware: Okay, I'll keep it outside the main repo then.

manthey: You don't have to -- adding the line to the .gitignore is fine.

manthey (approved):

PR #1011, opened 2023-07-13
Remove overlapping nuclei segmentation

subinkitware:
Remove overlapping nuclei present in the whole image to avoid double counting.


histomicstk/segmentation/label/remove_overlap_nuclei.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Create a shapely Polygon from a list of points.
+
+    Args:
+        points (list): A list containing tuples representing the points of the polygon.
+
+    Returns:
+        shapely.geometry.Polygon: The polygon created from the points.
+    """
+    return Polygon(coordinates).buffer(0)
```
manthey, line 18: The `buffer(0)` is unnecessary, at which point we may want to skip having this a separate function.

histomicstk/segmentation/label/remove_overlap_nuclei.py
```diff
@@ -0,0 +1,58 @@
...
+        nuclei_list (list): A list of dictionaries, each containing 'points' representing a polygon.
+
+    Returns:
+        list: A new list with overlapping nuclei removed.
+    """
+    # Extract all points from the nuclei_list to create polygons
+    coordinates_collection = [nuclei['points'] for nuclei in nuclei_list]
+
+    # Use ProcessPoolExecutor for parallel processing of polygon creation
+    with ProcessPoolExecutor(os.cpu_count()) as executor:
```
manthey, line 38: The default number of process pool workers is cpu count, so we don't have to pass it.

histomicstk/segmentation/label/remove_overlap_nuclei.py
```diff
@@ -0,0 +1,58 @@
...
+        polygons = list(
+            executor.map(
+                create_polygon,
+                (coordinates for coordinates in coordinates_collection)))
```
manthey, line 42: Is there a virtue of creating the `coordinates_collection`?  Couldn't you just do `nuclei['points'] for nuclei in nuclei_list`?  If not, then you should be able to skip this comprehension (e.g., `executor.map(create_polygon, (coordinates for coordinates in coordinates_collection))` should be the same as `executor.map(create_polygon, coordinates_collection)`)

histomicstk/segmentation/label/remove_overlap_nuclei.py
```diff
@@ -0,0 +1,58 @@
...
+    # Find indexes of overlapping polygons
+    remove_list = [
+        index for index in range(
+            len(polygons)) if any(
+            ix for ix in rt.query(
+                polygons[index]) if ix < index and polygons[index].intersects(
+                    polygons[ix]))]
+
+    # Remove the overlapping polygons from the original list
+    output_list = [nuclei_list[i] for i in range(len(nuclei_list)) if i not in remove_list]
```
manthey, line 56: We could combine the remove_list and output_list into a single command:
```
output_list = [nuclei for index, nuclei in enumerate(nuclei_list) if not any(...)]
```
where the ... is the expression in your current `any()`.

histomicstk/segmentation/label/remove_overlap_nuclei.py
subinkitware: I initially tried without `buffer(0)` but there are some self-intersecting polygons when trying this on a WSI. It throws an error with the polygon intersection part of the code if buffer(0) is not used.

histomicstk/segmentation/label/remove_overlap_nuclei.py
subinkitware: I agree, the `coordinate_collection` is redundant. I've replaced it with the `nuclei['points'] for nuclei in nuclei_list`

histomicstk/segmentation/label/remove_overlap_nuclei.py
subinkitware: here is a sample error message when not using the `.buffer` function. `shapely.errors.GEOSException: TopologyException: side location conflict at 3530.8899999999999 267.79000000000002 0. This can occur if the input geometry is invalid.`.  The invalid geometry is created when we use self-intersecting points to generate a polygon. `.buffer(0)` removes any self-intersecting geometry.

manthey:
If you have overlap and remove overlapping turned on and ask for bounding boxes instead of boundaries, this currently throws an error.  I think, for bounding box output, we would have to process as boundaries anyway to remove overlap and then reduce to bounding boxes afterwards.  

An alternative would be to just raise an error if remove overlapping nuclei AND bound box output are both selected (though it'd be better to handle it).

subinkitware:
> If you have overlap and remove overlapping turned on and ask for bounding boxes instead of boundaries, this currently throws an error. I think, for bounding box output, we would have to process as boundaries anyway to remove overlap and then reduce to bounding boxes afterward.
> 
> An alternative would be to just raise an error if removing overlapping nuclei AND bound box output are both selected (though it'd be better to handle it).

I think it can be handled in this PR itself. for bbox the library is returning a center point, height, and width. we could use that information to check for overlaps. I'll start working on it.

manthey:
> I think it can be handled in this PR itself. for bbox the library is returning a center point, height, and width. we could use that information to check for overlaps. I'll start working on it.

We don't just want to see if the bounding boxes overlap as that will cull nuclei that don't themselves overlap.  We have to do the intersection check on the polygon boundaries.

subinkitware:
> > I think it can be handled in this PR itself. for bbox the library is returning a center point, height, and width. we could use that information to check for overlaps. I'll start working on it.
> 
> We don't just want to see if the bounding boxes overlap as that will cull nuclei that don't themselves overlap. We have to do the intersection check on the polygon boundaries.

Okay, So is it like `if bbox overlaps ---> Create nuclei segmentation inside bbox ----- > check if nuclei intersect ------ > If intersects ----> Remove the nuclei.`

manthey:
> Okay, So is it like `if bbox overlaps ---> Create nuclei segmentation inside bbox ----- > check if nuclei intersect ------ > If intersects ----> Remove the nuclei.`

I don't think we want to run the segmentation multiple times, even if it is just on potential overlaps.  I think, if format is bbox and remove overlap is set, we would note that we need to convert our results to bbox and then generate the nuclei as if the format was boundary through removing the overlap.  Afterwards, we would convert the polyline annotations to rectangle annotations:
```
for nuclei in nuclei_list:
    nuclei['type'] = 'rectangle'
    nuclei['rotation'] = 0
    minx = min(p[0] for p in nuclei['points'])
    maxx = max(p[0] for p in nuclei['points'])
    miny = min(p[1] for p in nuclei['points'])
    maxy = max(p[1] for p in nuclei['points'])
    nuclei['center'] = [(minx + maxx) / 2, (miny + maxy) / 2, 0]
    nuclei['width'] = maxx - minx
    nuclei['height'] = maxy - miny
    nuclei.pop('closed')
    nuclei.pop('points')
```

histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -110,6 +125,10 @@ def detect_tile_nuclei(tile_info, args, src_mu_lab=None,
     flag_nuclei_found = np.any(im_nuclei_seg_mask)
 
     if flag_nuclei_found:
+        if args.nuclei_annotation_format == 'bbox' and args.remove_overlapping_nuclei_segmentation:
```
manthey, line 128: This might be clearer to do
```
if flag_nuclei_found:
    format = args.nuclei_annotation_format
    if args.nuclei_annotation_format == 'bbox' and args.remove_overlapping_nuclei_segmentation:
        format = 'boundary'
    nuclei_annot_list = cli_utils.create_tile_nuclei_annotations(im_nuclei_seg_mask, tile_info, format)
```

rather than have two return paths.

PR #1015, opened 2023-07-17
ColorDeconvolution-annotation attributes added to metadata

subinkitware:
I removed some of the repeating parameters from the current annotation data since they are already present in the `args` provided in `attributes`

### DIFF
--- a/histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
+++ b/histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
@@ -62,6 +62,11 @@ def main(args):
                     'yoffset': region.get('top', 0),
                 },
             }],
+            'attributes': {
+                'cli': Path(__file__).stem,
+                'params': vars(args),
+                'version': histomicstk.__version__,
+            },
         }, {
             'name': 'Deconvolution %s ' % (
                 args.stain_2 if args.stain_2 != 'custom' else str(args.stain_2_vector)),
@@ -73,6 +78,11 @@ def main(args):
                     'yoffset': region.get('top', 0),
                 },
             }],
+            'attributes': {
+                'cli': Path(__file__).stem,
+                'params': vars(args),
+                'version': histomicstk.__version__,
+            },
         }, {
             'name': 'Deconvolution %s ' % (
                 args.stain_3 if args.stain_3 != 'custom' else str(args.stain_3_vector)),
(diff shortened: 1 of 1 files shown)


histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -89,9 +84,17 @@ def main(args):
                     'yoffset': region.get('top', 0),
                 },
             }],
-        }]
+        },
```
manthey, line 87: Rather than create another annotation, this should be added to the current annotations.

histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -89,9 +94,14 @@ def main(args):
...
             }],
+            'attributes': {
+                'cli': Path(__file__).stem,
+                'params': vars(args),
+                'version': histomicstk.__version__,
+            }
         }]
         if args.stain_3 == 'null':
-            annotation[2:] = []
+            annotation[2:3] = []
```
manthey, line 104: You don't need this change.

histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -76,11 +78,14 @@ def main(args):
...
+                'cli': Path(__file__).stem,
+                'params': vars(args),
+                'version': histomicstk.__version__,
+            },
         }, {
-            'name': 'Deconvolution %s - %s' % (
-                args.stain_3 if args.stain_3 != 'custom' else str(args.stain_3_vector),
-                os.path.splitext(os.path.basename(args.outputAnnotationFile))[0]),
-            'description': 'Used params %r' % vars(args),
+            'name': 'Deconvolution %s ' % (
```
manthey, line 87: What is the logic of changing what these had been named?  

If it makes sense, then we don't want a trailing space as part of the name.

subinkitware: I removed that because it was adding the output file name `Color Deconvolution-outputAnnotationFile` to the annotation name which is already present in the newly added `attribute: { params: vars(args)...`. Also, the name is shorter and represents `process: Deconvolution` and `strain: E / H`. I will remove the trailing white spaces in the name.

manthey (approved):

PR #1016, opened 2023-07-18
SuperpixelSegmentation-annotation-attributs-added-to-metadata

subinkitware:
Most of the information generated by the code is already added to the metadata. so, I've only added the `attributes`. One change I would like to add is to replace the `opts` name with `args` for consistency.

### DIFF
--- a/histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
+++ b/histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
@@ -236,6 +236,11 @@ def createSuperPixels(opts):  # noqa
                 'categories': categories,
                 'boundaries': opts.boundaries,
             }],
+            'attributes': {
+                'params': vars(opts),
+                'cli': Path(__file__).stem,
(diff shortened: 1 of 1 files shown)


histomicstk/cli/SuperpixelSegmentation/SuperpixelSegmentation.py
```diff
@@ -235,14 +236,18 @@ def createSuperPixels(opts):  # noqa
...
+                'version': histomicstk.__version__,
+            },
         }
         if len(bboxes) and str(opts.bounding).lower() != 'separate':
             annotation['elements'][0]['user'] = {'bbox': bboxesUser}
         if len(bboxes) and str(opts.bounding).lower() != 'internal':
             bboxannotation = {
                 'name': '%s bounding boxes' % os.path.splitext(
                     os.path.basename(opts.outputAnnotationFile))[0],
-                'description': 'Used params %r' % vars(opts),
```
manthey, line 245: Also add the attributes to this annotation.

manthey (approved):

PR #1021, opened 2023-07-21
BackgroundIntensity-added-annotation-metadata

subinkitware:
This code returns `returnParameterFile`. It was added to the attributes. I was unable to run the code with the code with the images I have. It would be great to have a discussion about each of these tools in HistomicsTK.

### DIFF
--- a/histomicstk/cli/BackgroundIntensity/BackgroundIntensity.py
+++ b/histomicstk/cli/BackgroundIntensity/BackgroundIntensity.py
@@ -28,8 +28,8 @@ def main(args):
 
     annotation = {
         'name': 'BackgroundIntensity',
-        'intensity_values': list(map(str, I_0)),
         'attributes': {
+            'intensity_values': I_0.tolist(),
             'params': vars(args),
             'cli': Path(__file__).stem,
             'version': histomicstk.__version__,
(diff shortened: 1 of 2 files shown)


manthey:
We should change this from a return parameters file to an annotation (like was done in #1023).

histomicstk/cli/BackgroundIntensity/BackgroundIntensity.xml
```diff
@@ -60,12 +60,13 @@
       <description>Use instead of sample_fraction to specify roughly how many pixels to sample.  The fewer tiles are excluded, the more accurate this will be.</description>
       <default>-1</default>
     </integer>
-    <double-vector>
-      <name>BackgroundIntensity</name>
-      <label>Background Intensity</label>
+    <file fileExtensions=".anot" reference="sample_slide_path">
```
manthey, line 63: The reference has to match the name of the input file's parameter (`slide_path` in this case)

histomicstk/cli/BackgroundIntensity/BackgroundIntensity.py
```diff
@@ -15,8 +19,25 @@ def main(args):
...
+
+    #
+    # Write annotation file
+    #
+
+    print('\n>> Writing annotation file ...\n')
+
+    annotation = {
+        'name': 'BackgroundIntensity',
+        'intensity_values': list(map(str, I_0)),
```
manthey, line 31: This needs to be in the `attributes` dictionary.  Additionally, rather than casting this to a string, we should probably just do `'intensity_values': I_0.tolist(),`.

manthey (approved):

PR #1017, opened 2023-07-18
SeperateStainsXuSnmf-outputfile-metadata

subinkitware:
added annotation metadata. I'm a bit confused as to what to change in the XML file. Also, I was unable to run the code for `seperatestainsxusnmf` with the images I have.

PR #1023, opened 2023-07-26
Instead of output a return parameters file, output an annotation.

manthey:
This has the same results, but in json instead of key-value pairs.  It also includes the input parameters.

This replaces #1017

### DIFF
--- a/histomicstk/cli/SeparateStainsXuSnmf/SeparateStainsXuSnmf.py
+++ b/histomicstk/cli/SeparateStainsXuSnmf/SeparateStainsXuSnmf.py
@@ -1,12 +1,16 @@
+import json
+from pathlib import Path
+
 import numpy
 
+import histomicstk
 import histomicstk.preprocessing.color_deconvolution as htk_cdeconv
 from histomicstk.cli import utils
 from histomicstk.cli.utils import CLIArgumentParser
 
 
-def main(args):
-    args = utils.splitArgs(args)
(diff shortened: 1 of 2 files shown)


manthey:
@subinkitware Please test this (I'll note that it is probably very slow on a large image, but that is no different than before) and approve it if it works for you.


Note the special parameter name for the annotation that ends in "AnnotationFile" and the reference of that output to the name of the input image parameter.  This tells the hooks in HistomicsUI to convert it to an annotation.

subinkitware:
> @subinkitware Please test this (I'll note that it is probably very slow on a large image, but that is no different than before) and approve it if it works for you.
> 
> Note the special parameter name for the annotation that ends in "AnnotationFile" and the reference of that output to the name of the input image parameter. This tells the hooks in HistomicsUI to convert it to an annotation.

I was able to run it on a WSI and on another small image. It takes quite a bit of time to execute as you mentioned. Should I be expecting to see some changes in the image output as the process finishes? or is it just saving the stain colors once I run `SeperateStainsXuSnmf`

manthey:
> I was able to run it on a WSI and on another small image. It takes quite a bit of time to execute as you mentioned. Should I be expecting to see some changes in the image output as the process finishes? or is it just saving the stain colors once I run `SeperateStainsXuSnmf`

It just writes an annotation without any elements that contains the stain colors.

PR #1033, opened 2023-07-28
Update defaults color deconvolution

### LINKED ISSUE #1030: Ensure each cli runs without needing to set anything
Ideally, each CLI would run in the HistomicsUI without any changes to its settings; they may not be optimal, but they would throw any errors.  I would test this on a TCGA image that is a reasonably large size.  Additionally, if the CLI supports using a rectangular ROI, it should run without any changes except setting that ROI.

List of CLI's to be updated.

- [x] background intensity (updated the sample fraction amount)
- [x] color deconvolution
- [x] compute nuclei features (mathey updated the features library)
- [x] positive pixel count (updated-pending approval)
- [x] separate stains Xusnmf (defaults already present in the code)
- [x] superpixel segmentation (defaults already present in the code)

subinkitware:
-Here I have changed the .xml file extensions from `.png` to `.tiff`, same will be reflected in the description

-Color deconvolution is divided into two functions
1) color deconvolution WSI - if the user is not specifying the region
2) color deconvolution ROI - if the user provides the region.

-  Both of the above functions return a parameter called `transform`
- Since there is no region present in the WSI I have given a return as `{'No transform available for WSI'}`, This can be changed to anything that is appropriate.
- Going further down the code the return parameter will be saved in the `transform: {....}` tag in the annotation.
- This update fixes #1030

### DIFF
--- a/histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
+++ b/histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
@@ -20,6 +20,8 @@ def colorDeconvolve(args):
     ts = large_image.getTileSource(args.inputImageFile)
     region = {}
     it_kwargs = {}
+    tileSize=8192
+    it_kwargs['tile_size'] = dict(width=tileSize, height=tileSize)
 
     # Provides crop area if ROI present in arguments
     if np.all(np.array(args.region) != -1):
@@ -30,13 +32,10 @@ def colorDeconvolve(args):
             'height': args.region[3],
             'units': 'base_pixels'
         }
-        sink0.crop = (it_kwargs['region']['left'], it_kwargs['region']['top'],
-                      it_kwargs['region']['width'], it_kwargs['region']['height'])
-        sink1.crop = (it_kwargs['region']['left'], it_kwargs['region']['top'],
-                      it_kwargs['region']['width'], it_kwargs['region']['height'])
-        sink2.crop = (it_kwargs['region']['left'], it_kwargs['region']['top'],
-                      it_kwargs['region']['width'], it_kwargs['region']['height'])
-        region = utils.get_region_dict(args.region, args.maxRegionSize, ts)['region']
+        sink0.crop = tuple(args.region)
+        sink1.crop = tuple(args.region)
+        sink2.crop = tuple(args.region)
+        region = utils.get_region_dict(args.region, None, ts)['region']
 
     # Create stain matrix
     print('>> Creating stain matrix')
--- a/histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
+++ b/histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
@@ -1,9 +1,9 @@
 import json
 import logging
-import numpy as np
 from pathlib import Path
 
 import large_image
+import numpy as np
 
 import histomicstk
 import histomicstk.preprocessing.color_deconvolution as htk_cd
@@ -12,9 +12,10 @@ from histomicstk.cli.utils import CLIArgumentParser
 
 logging.basicConfig()
 
-def colordeconvolve(args):
 
+def colorDeconvolve_ROI(args):
     import skimage.io
+
     ts = large_image.getTileSource(args.inputImageFile)
 
     im_input = ts.getRegion(
@@ -46,9 +47,15 @@ def colordeconvolve(args):
 
     region = utils.get_region_dict(args.region, args.maxRegionSize, ts)['region']
 
-    return region
+    transform = {
+        'xoffset': region.get('left', 0),
+        'yoffset': region.get('top', 0),
+    }
 
-def colordeconvolve_WSI(args):
+    return transform
+
+
+def colorDeconvolve_WSI(args):
     sink0 = large_image.new()
     sink1 = large_image.new()
     sink2 = large_image.new()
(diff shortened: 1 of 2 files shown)


manthey:
We should be able to use the same process regardless of whether there is an ROI or the whole image -- the tileIterator would just have the region parameter with an ROI and not on the WSI.

The transform for a WSI would just be offset_x/offset_y both equal 0 or not set at all.  Turning it into a string will break it -- it is a dictionary telling where the image annotation is placed on the parent image.   See https://girder.github.io/large_image/annotations.html#image-overlays.

subinkitware:
In addition to the cropping with ROI on the latest commit, the `transform` dict is moved from the attributes to elements in the annotation. That was needed to show the results on UI.

histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -12,20 +13,30 @@
...
-    import skimage.io
-
-    # Read Input Image
-    print('>> Reading input image')
-
-    print(args.inputImageFile)
-
+def colorDeconvolve(args):
+    sink0 = large_image.new()
+    sink1 = large_image.new()
```
manthey, line 18: if `args.stain_2 == 'null'`, we shouldn't be creating sink1, and likewise with `args.stain_3` and sink2.  I realize we were creating pngs before, but not adding them as annotations.

subinkitware:
The new updates contain the following changes

- the name of `sink0` was changed to `sink1` since the args are defined as `stain1,....2,...3`
- if the condition is added which checks If the user gives the stain values.
- the `<index>3</index>` was removed from the `.xml` file since it was throwing an error from girder if the `stain3` is given as 'null' and a filename present in the `stain_3` dialog box.

histomicstk/cli/ColorDeconvolution/ColorDeconvolution.xml
```diff
@@ -25,26 +25,25 @@
...
       <index>2</index>
     </image>
-    <image fileExtensions=".png" reference="inputImageFile">
+    <image fileExtensions=".tiff" reference="inputImageFile">
       <name>outputStainImageFile_3</name>
       <label>Output Image of Stain 3</label>
-      <description>Output Image of Stain 3 (*.png)</description>
+      <description>Output Image of Stain 3 (*.tiff)</description>
       <channel>output</channel>
-      <index>3</index>
```
manthey, line 47: I would expect removing the index parameter to fail some sort of validation.  This is officially specifying that the CLI takes positional inputs (e.g., `<command> <inputImageFile> <outputStainImageFile_1> <outputStainImageFile_2> <outputStainImageFile_3>`).  Without the `<index>3</index>` parameter and no flags, there is no way to specify the third output file.

The actual issue is that I had said not to emit the second and/or third stain image files unless they would be used.  We need to put that back.  Better (but not this PR) would be to add a flag in the slicer_cli_web extensions to ctk_cli that would allow for optional output files.  girder_worker via slicer_cli_web throws an exception is we were told such a file should exist and it doesn't.

subinkitware: > I would expect removing the index parameter to fail some sort of validation. This is officially specifying that the CLI takes positional inputs (e.g., `<command> <inputImageFile> <outputStainImageFile_1> <outputStainImageFile_2> <outputStainImageFile_3>`). Without the `<index>3</index>` parameter and no flags, there is no way to specify the third output file.
> 
> The actual issue is that I had said not to emit the second and/or third stain image files unless they would be used. We need to put that back. Better (but not this PR) would be to add a flag in the slicer_cli_web extensions to ctk_cli that would allow for optional output files. girder_worker via slicer_cli_web throws an exception is we were told such a file should exist and it doesn't.

Okay, So you are recommending removing the last commit which adds these changes, and on another PR add a flag in the `slicer_cli_web` extensions.

.gitignore
```diff
@@ -93,6 +93,7 @@ HistomicsTKDocker.code-workspace
 MANIFEST.in
 .pre-commit-config.yaml
 run_tests_code_cleanup.sh
+seliniumTest.py
```
manthey, line 96: Where did this come from? Are you using selenium in some manner?

histomicstk/cli/ColorDeconvolution/ColorDeconvolution.xml
manthey: Until we make a change in some upstream library (probably slicer_cli_web), we have to leave the index parameter in the xml file AND emit the images files for all three stains.

.gitignore
subinkitware: Yeah, I wrote some code to automate the UI processes after a new docker is deployed. The code can go until a cli job is submitted with/without an ROI. 

I got the spelling wrong for the filename.

manthey:
When run on a large image, this ends up being very slow due to tile size.  I think we need to add 
```
    tileSize = 4096
    it_kwargs['tile_size'] = dict(width=tileSize, height=tileSize)
```
but 4096 may not be the right value -- I'd try it on your system and make sure it doesn't use too much memory.  You might want to try 2048 and 8192 as well.

Also, because we now handle arbitrarily large images, we can get rid of the `maxRegionSize` parameter from the xml (just pass `None` in `utils.get_region_dict`).

histomicstk/cli/ColorDeconvolution/ColorDeconvolution.py
```diff
@@ -12,20 +13,30 @@
...
+    # Provides crop area if ROI present in arguments
+    if np.all(np.array(args.region) != -1):
+        it_kwargs['region'] = {
+            'left': args.region[0],
+            'top': args.region[1],
+            'width': args.region[2],
+            'height': args.region[3],
+            'units': 'base_pixels'
+        }
+        sink0.crop = (it_kwargs['region']['left'], it_kwargs['region']['top'],
```
manthey, line 33: These could just be `sink0.crop = tuple(args.region)`

manthey (approved):

PR #1034, opened 2023-08-01
Harden compute nuclei features

manthey:
Don't fail if we can't identify cytoplasm because the tile is small or nuclei are close to overlapping.

### DIFF
--- a/histomicstk/features/compute_gradient_features.py
+++ b/histomicstk/features/compute_gradient_features.py
@@ -93,6 +93,8 @@ def compute_gradient_features(im_label, im_intensity,
     cannyG = canny(im_intensity)
 
     for i in range(numLabels):
+        if rprops[i] is None:
+            continue
 
         # get gradients of object pixels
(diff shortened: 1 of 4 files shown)


manthey:
@subinkitware Looking at the nuclei classification code, it finds nuclei and then assumes that any pixels near but outside the nuclei are cytoplasm.  If there are no such pixels, then it fails to compute cytoplasm feature values.  This can happen where nuclei are close together, or, because of tiling, a nucleus occupies essentially all of a fractional tile.  This hardens the process by skipping those computations where we can't label cytoplasm in this manner.

subinkitware:
Okay, now I understand the problem. This must be the issue happening when WSL is loaded into the `ComputeNucleiFeatures`. The system is unable to find cytoplasm from tiles that have the nuclei segmented already. This problem needed a bit of understanding of the theory too.

subinkitware (approved):

PR #1036, opened 2023-08-01
Resolve issues with sink.crop

subinkitware:
- while running the `PPC` with default values there are no issues.
- But when we provide ROI and forget to mention the `outputLabelImage`, an error is thrown saying
`sink.crop=... None type object not callable`

- To resolve that I've made both `outputannotationfile` file and `outputlabelimage` as required fields. I believe the way to do that is to add the `<index>2</index>` parameter to the .XML file. That is how I did it. If it is something different please let me know.

### DIFF
--- a/histomicstk/cli/PositivePixelCount/PositivePixelCount.py
+++ b/histomicstk/cli/PositivePixelCount/PositivePixelCount.py
@@ -193,9 +193,10 @@ def main(opts):
     }
 
     # Save the annotation dictionary to the output annotation file
-    with open(opts.outputAnnotationFile, 'w') as annotation_file:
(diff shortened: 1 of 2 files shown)


manthey:
The alternative is to not call sink.crop if sink is None.  There is a valid use case where you want to count the positive pixels but you don't want an output image.

manthey:
Can we also add a guard before writing the annotation file?  That is, just before https://github.com/DigitalSlideArchive/HistomicsTK/blob/d7f837f0789d6a65a0ae8669421763ed72fae109/histomicstk/cli/PositivePixelCount/PositivePixelCount.py#L196, add `if opts.outputAnnotationFile:`

manthey (approved):

PR #1042, opened 2023-08-04
remove overlapping nuclei is a default process.

### LINKED ISSUE #1039: Nuclei Detection Overlap defaults
I think we should make an overlap value of `-1` be a special value that means `(max_radius + 1) * 4`, and then make that the default value  AND turn on remove overlapping nuclei by default.

subinkitware:
I  made the following changes in `nucleidetection.xml`

- updated the overlap value to 84 - which is (max_radius + 1) * 4
- set the `remove overlap nuclei` option to `True` by default.

One additional question on this PR.
- Is there a way to show the value provided by the `.py` file on the UI ?
- For example, the overlap value changes automatically when the max_radius is changed. + user can go to the UI and update it too.

- completing this will fix #1039


histomicstk/cli/NucleiDetection/NucleiDetection.py
```diff
@@ -275,6 +275,10 @@ def main(args):
     else:
         process_whole_image = False
 
+    # Provide default value for tile_overlap
+    tile_overlap = args.tile_overlap_value if args.tile_overlap_value != - \
```
manthey, line 279: This line break halfway through a number makes it very hard to see what is going on.  Better would be
```
    tile_overlap = args.tile_overlap_value
    if tile_overlap == -1:
        tile_overlap = (args.max_radius + 1) * 4
```
If you really want to use the ternary, it'd still be better to do
```
    tile_overlap = (
        args.tile_overlap_value if args.tile_overlap_value != -1 else
        ((args.max_radius + 1) * 4))
```

PR #1045, opened 2023-08-07
Abstract nuclei detection

subinkitware:
In this update, the function `detect_tile_nuclei` which is common to both `NucleiDetection.py` and `ComputeNucleiFeatures.py` is moved to `segmentation---->nuclear--->detect_tile_nuclei.py`

- In addition to that I have added a parameter `return_fdata` to `detect_tile_nuclei` function. If that is `True` it will return the fdata that is required for calculating the nuclei features.
- In the case of just `NucleiDetection` the `return_fdata` is kept `False`

### DIFF
--- a/histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
+++ b/histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
@@ -8,100 +8,13 @@ import large_image
 import numpy as np
 
 import histomicstk
-import histomicstk.features as htk_features
-import histomicstk.preprocessing.color_deconvolution as htk_cdeconv
 import histomicstk.preprocessing.color_normalization as htk_cnorm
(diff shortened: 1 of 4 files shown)


manthey (approved):
As a follow up, we should make the compute features cli have the same defaults as the nuclei detection regarding overlap and removing duplicates.  For completeness, we could add these parameters into the compute features options in the Nuclei Segmentation section.

subinkitware:
> As a follow up, we should make the compute features cli have the same defaults as the nuclei detection regarding overlap and removing duplicates. For completeness, we could add these parameters into the compute features options in the Nuclei Segmentation section.



> As a follow up, we should make the compute features cli have the same defaults as the nuclei detection regarding overlap and removing duplicates. For completeness, we could add these parameters into the compute features options in the Nuclei Segmentation section.

I did work on running nuclei classification cli with the same defaults in PR #1051, But I added all the required parameters programmatically into a function `set_reference_val(args)`.

PR #1050, opened 2023-08-17
Tutorial nuclei classification

subinkitware:
Tutorial for nuclei classification using DSA and HistomicsTK.

The downloadable resources are yet to be added.


manthey:
Since this is all markdown, we aren't really using the features of a notebook very much.  Do you think it'd be better to just have it be a markdown file with some upload images?  GitHub renders it just the same, so there may be no virtue to that.

manthey:
Usually we try to put image resources for docs in the same repo.  They aren't that large (and if they are png files, you can run `optipng` on them to make them as small as possible), and if someone checks out the repo it is closer to being stand alone.  I'd only do this for the images (placing them in the doc_files folder, for instance).

subinkitware:
> Usually we try to put image resources for docs in the same repo. They aren't that large (and if they are png files, you can run `optipng` on them to make them as small as possible), and if someone checks out the repo it is closer to being stand alone. I'd only do this for the images (placing them in the doc_files folder, for instance).

I have used the `optimpng` and transferred all the files to the doc_files repo. But there is a model file as `.pkl`(6 MB). Currently, it is in my personal public folder at data.kitware.com. Can you provide me a location to save it as a part of the HistomicsTK folder?

manthey (approved):

PR #1051, opened 2023-08-17
Nuclei classification can run without feature / annotation files

subinkitware:
- If the user doesn't provide any `nuclei features` or `nuclei annotation` files, the system generates both of these files based on default parameters.
- Once that is generated it classifies the nuclei based on the model provided by the user. 

**- Issues faced**
1. Number of nuclei and features are not the same since they are generated by two different code sets with different methodologies. ( mainly it affects the WSI, for small image patches they are the same )
2. Nuclei features sometimes create NaN entries in the feature file ( observed for WSI ).
3. automatic `feature` and `annotation` generation takes a long time if it's a large WSI.

**- Temporary fix for these issues**
1. The features that exceed the number of nuclei are removed.
2. Rows with NaN are dropped from the generated nuclei features.


manthey:
@subinkitware Do you remember what this needs to be ready for review?

subinkitware:
> @subinkitware Do you remember what this needs to be ready for review?

Yeah, this is ready for review, I just rechecked the code. In a gist this update is resolving this issue:  In older code, user has to upload the classification model, annotations, and features to run a classification job. Otherwise, it won't let the user run it. With this update, if the user doesn't provide any features or nuclei segmentation, HistomicsTK calculates both annotations & features by invoking the segmentation & feature extraction code in the other cli's

manthey (approved):

PR #1055, opened 2023-08-21
Update nuclei feature extraction

subinkitware:
Here the nuclei feature extraction is updated with the following features

1. Previously the annotations in the features output file (xmin, ymin, xmax, ymax) were not referenced to the input WSI dimensions, their coordinates were pulled from the input_tile_image. 
2. I added a reference to the `tile_info` so that the annotations in the feature file match the coordinates of the WSI.
3. There is a mismatch between the number of nuclei extracted in the annotation file and in the features file. They need to be synchronized. - working on that issue.

### DIFF
--- a/histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
+++ b/histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
@@ -306,7 +306,8 @@ def main(args):
             for annot_list, fdata in tile_result_list if fdata is not None],
             ignore_index=True
         )
-
+    # replace any NaN in the feature colums with Zero vaue
+    nuclei_fdata.fillna(0)
     nuclei_detection_time = time.time() - start_time
 
     print(f'Number of nuclei = {len(nuclei_annot_list)}')
--- a/histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
+++ b/histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
@@ -2,13 +2,18 @@ import json
 import logging
 import os
 import time
+from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
 
 import large_image
 import numpy as np
+import pandas as pd
+import shapely
+from shapely.geometry import Polygon
 
 import histomicstk
 import histomicstk.preprocessing.color_normalization as htk_cnorm
+import histomicstk.segmentation.label as htk_seg_label
 import histomicstk.segmentation.nuclear as htk_nuclear
 import histomicstk.utils as htk_utils
 from histomicstk.cli import utils as cli_utils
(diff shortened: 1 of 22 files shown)


histomicstk/cli/ComputeNucleiFeatures/ComputeNucleiFeatures.py
```diff
@@ -35,6 +39,78 @@ def check_args(args):
...
+    return (data[0], Polygon(data[1]).buffer(0))
+
+
+def find_and_remove_non_overlapping_polygons(parameters):
+    annot_polygon, feature_tree, \
+        intersecting_features_order, intersecting_polygons_order = parameters
+    intersecting_candidates = feature_tree.query(annot_polygon[1])
+    if intersecting_candidates.any():
+        for element in intersecting_candidates:
+            if element not in intersecting_features_order:
```
manthey, line 52: This is another O(n^2) loop.

manthey:
I think (but haven't proved it), that we can skip the whole synchronization step you have if we do a little more bookkeeping up front.

Specifically, I think the list of nuclei and computed features has a one-to-one relationship before we remove overlaps.  Both routines use scikit-image's regionprops to enumerate data.  If these aren't in agreement, then we should change the code to only call regionprops once and reuse the data from that.

Given that, if we modify the overlap code to something like:
```
def remove_overlap_nuclei(nuclei_list, nuclei_format, fdata=None):
    polygons = [create_polygon(nuclei['points']) for nuclei in nuclei_list]
    rt = shapely.strtree.STRtree(polygons)
    output_indices = [index for index, nuclei in enumerate(nuclei_list) if not any(
        ix for ix in rt.query(
            polygons[index]) if ix < index and polygons[index].intersects(
            polygons[ix]))]
    output_list = [nuclei_list[idx] for idx in output_indices]
    if nuclei_format == 'bbox':
        output_list = convert_polygons_tobbox(output_list)
    if fdata is not None:
        fdata = fdata.loc[output_indices]
        return output_list, fdata
    return output_list
```
then we would remove the data the same way from the output nuclei and the feature set.

manthey:
Reusing the regionprops would probably greatly speed up the process, as I imagine it is a major part of the time.  In this case, I'd call the feature computation code first since it gets the region props with the original image, and then return those props and pass them to the nuclei finder.  I haven't looked at how much work this would be to do without breaking whatever api surface we want to preserve.

subinkitware:
> I think (but haven't proved it), that we can skip the whole synchronization step you have if we do a little more bookkeeping up front.
> 
> Specifically, I think the list of nuclei and computed features has a one-to-one relationship before we remove overlaps. Both routines use scikit-image's regionprops to enumerate data. If these aren't in agreement, then we should change the code to only call regionprops once and reuse the data from that.
> 
> Given that, if we modify the overlap code to something like:
> 
> ```
> def remove_overlap_nuclei(nuclei_list, nuclei_format, fdata=None):
>     polygons = [create_polygon(nuclei['points']) for nuclei in nuclei_list]
>     rt = shapely.strtree.STRtree(polygons)
>     output_indices = [index for index, nuclei in enumerate(nuclei_list) if not any(
>         ix for ix in rt.query(
>             polygons[index]) if ix < index and polygons[index].intersects(
>             polygons[ix]))]
>     output_list = [nuclei_list[idx] for idx in output_indices]
>     if nuclei_format == 'bbox':
>         output_list = convert_polygons_tobbox(output_list)
>     if fdata is not None:
>         fdata = fdata.loc[output_indices]
>         return output_list, fdata
>     return output_list
> ```
> 
> then we would remove the data the same way from the output nuclei and the feature set.

I started this way assuming that the number of nuclei coming from nuclei segmentation and nuclei feature extraction are the same. But they are not. The nuclei segmentation (without any overlap removal) has a function 'htk_seg.label.trace_object_boundaries' which changes the number of nuclei found. I did do a test run to get the exact numbers (`with boundary - 2856`, `bbox segmentation (pure rmsprops) - 2867`)

subinkitware:
> preserve



> Reusing the regionprops would probably greatly speed up the process, as I imagine it is a major part of the time. In this case, I'd call the feature computation code first since it gets the region props with the original image, and then return those props and pass them to the nuclei finder. I haven't looked at how much work this would be to do without breaking whatever api surface we want to preserve.

I could create a separate utils function that could take the region props and generate the annotations required with the same number of features. Thus, it won't need any changes with the current functions on utils.

subinkitware:
This is interesting, The region props give different numbers of nuclei on different runs. I think that is the one which causing the issues not `htk_seg.label.trace_object_boundaries`. As you mentioned above reusing the regionprops should solve it.

subinkitware:
#1070 is the updated solution.

PR #1062, opened 2023-08-28
Don't use a futures pool to compute polygons.

manthey:
On a full WSI, this is very slow because of all the data transfer between processes or threads.  On a WSI with 300k nuclei, doing it as a list comprehension takes less than a minute on my machine, but takes hours using a processpool (and four or five minutes using a threadpool).


manthey:
@subinkitware I know you profiled this on your machine.  Did you do it on a large WSI or just a small image?

subinkitware (approved):

PR #1069, opened 2023-09-01
Sync number of nuclei and features

subinkitware:
The disparity between the number of nuclei found using segmentation and rows of feature maps is caused by two things

1. Region props generate different numbers of nuclei on each run
2. Creating a nuclei outline using `htk_seg.label.trace_object_boundaries` can also reduce the number of nuclei

I assume these issues could be fixed by the following
1. Sharing region props variable between nuclei detection and feature extraction.
2. Combining both nuclei detection and feature extraction into a single function.
3. Triggering feature extraction only if a nuclei boundary is found.

Among all the processes involved in the nuclei feature extraction. Extracting individual features is a resource-intensive process. running them parallel could save some time when processing WSI.

The following are the things that need to be done

- [x] combine both nuclei and feature extraction into one
- [x] Run feature extraction only if boundaries are detected.
- [x] Fix bugs
- [x] Compute the time of processing
- [x] Test if parallel processing of the feature extraction reduces the total processing time for WSI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,28 +303,6 @@ jobs:
             cd ..
             python3 -m pip install histomicstk --pre --find-links project/wheels --find-links https://girder.github.io/large_image_wheels
             python3 -c "import histomicstk"
-  release:
-    working_directory: ~/project
-    docker:
-      - image: cimg/python:3.9
-      - image: docker:git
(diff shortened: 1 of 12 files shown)


subinkitware:
This method may not be suitable

PR #1070, opened 2023-09-06
synced nuclei features and nuclei detection.

subinkitware:
- This code contains changes in 'utils', but it doesn't change the behavior of the functions in any way unless you provide the extra attribute.
- The output rows of feature extraction are the same as the number of nuclei.
- 'Regionprops' is shared between feature extraction and nuclei detection (Faster than all the other methods I added)
- A slight change in 'nuclei detection' was also added since the output only contains a nuclei array if not triggered to generate features.


histomicstk/segmentation/label/remove_overlap_nuclei.py
```diff
@@ -48,32 +48,38 @@ def convert_polygons_tobbox(nuclei_list):
...
 
     # Build the STRtree from all the polygons
     rt = shapely.strtree.STRtree(polygons)
 
     # Find and remove any overlapping polygons
-    output_list = [nuclei for index, nuclei in enumerate(nuclei_list) if not any(
-        ix for ix in rt.query(
-            polygons[index]) if ix < index and polygons[index].intersects(
-            polygons[ix]))]
+    if return_selected_nuclei:
```
manthey, line 72: I don't think you want this `if` statement, since if it is true, `output_list` won't be defined and you use the condition a few lines below.

Issue #1105, opened 2024-03-23
Remove redundant access to DataFrame


X-TRON404:
The `histomicstk.features` module contains functions that create a `DataFrame` object and then access various indices within the DataFrame to apply modifications in a for loop. This process introduces significant latency. I am collaborating with @cooperlab  on this issue and have devised a potential solution. @manthey, could you please assign this issue to me? I am ready to begin work on a new branch to address it.

PR #1106, opened 2024-03-23
fix: redundant writes to dataframe #1105

X-TRON404:
I have submitted a pull request to address issue #1105. This entails first establishing a data structure, making modifications to it, and then generating a dataframe—all in a single instance.

### DIFF
--- a/histomicstk/features/compute_gradient_features.py
+++ b/histomicstk/features/compute_gradient_features.py
@@ -64,7 +64,6 @@ def compute_gradient_features(im_label, im_intensity, num_hist_bins=10, rprops=N
        and statistics tables and formulae," Crc Press, 1999.
 
     """
-
     # List of feature names
     feature_list = [
         'Gradient.Mag.Mean',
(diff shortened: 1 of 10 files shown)


histomicstk/features/compute_gradient_features.py
```diff
@@ -78,57 +78,56 @@ def compute_gradient_features(im_label, im_intensity,
...
     numLabels = len(rprops)
-    fdata = pd.DataFrame(np.zeros((numLabels, numFeatures)),
-                         columns=feature_list)
 
     Gx, Gy = np.gradient(im_intensity)
     diffG = np.sqrt(Gx**2 + Gy**2)
     cannyG = canny(im_intensity)
 
-    for i in range(numLabels):
-        if rprops[i] is None:
```
cooperlab, line 96: @X-TRON404 I think you are missing this statement

cooperlab (approved):

X-TRON404:
I've made an update to the section of the code where we initially create the features  DataFrame and then modify it by accessing its elements repeatedly. In the improved version, I'm now constructing the features using a native Python data structure first, and then transferring these features to the DataFrame in a single operation. This change has resulted in a performance increase of up to ~75%.

Issue #1139, opened 2024-10-21
Logical error in function compute_fsd_features


name-used:
Version: bf0f18ecea93087dfbf541d8c3c63fed1de9a283
Commits on May 13, 2024
Branch master

There are two errors at file histomicstk.features.compute_fsd_features.py: line 77 and line 188.

1. np.argwhere returns boundary points in coordinate order instead of natural order.
Think about this: 
12345
60007
89ABC
The correct contour order is "123457CBA986" or any rotation of this contour such as "57CBA9861234".
But the code in line 77 returns "123456789ABC".
If rotate this matrix into:
57C
40B
30A
209
168
The returning becomes to "57C4B3A29168".
There is no Fourier correlation between these two sequences.
Therefore, the features based on this code are sensitive to any rotation.
2. Curvature here is actually the angle of inclination.
Even if the inclination angle of the first coordinate has been subtracted, this number is still a value in radians.
The problem is that in radian calculation, 0==2pi.
Due to the natural error in floating-point calculations, an almost zero radian value can easily be calculated as 2pi - eps.
In contrast, the Fourier transform of real functions does not consider 0==2pi, which means that in similar transformation processes, values close to 0 will undergo jumps with each transformation.
This undoubtedly does not meet the calculation purpose of contour features.

For Correction Suggestions:

1. Use cv2.findContours
```
# find boundaries
# Bounds = np.argwhere(
#     find_boundaries(lmask, mode='inner').astype(np.uint8) == 1,
# )
[Bounds], _ = cv2.findContours(
    (find_boundaries(lmask, mode='inner').astype(np.uint8) == 1).astype(np.uint8),
    # cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_SIMPLE,
    cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_NONE,
)
Bounds = Bounds[:, 0, :]
```
2. Calculate as complex coordinates
```
Curvature = Curvature - Curvature[0]
# calculate FFT
# fX = np.fft.fft(Curvature ).T
z = 1 * np.cos(Curvature) + 1j * np.sin(Curvature)
fX = np.fft.fft(z).T
```

cooperlab:
@FattanePourakpour can you take a look at this? This is clearly wrong - a function like `trace_object_boundaries` or opencv should be used to trace boundaries as recommended above (this would be `bwboundaries` if you are coming from Matlab).

[...]

PR #1143, opened 2024-10-31
Update boundary detection in FSD feature extraction

FattanePourakpour:
`np.argwhere` does not trace boundaries accurately; instead, `trace_object_boundaries` is used.


cooperlab:
@FattanePourakpour please address the complex curvature problem also mentioned in Issue #1139 and then let's look at getting the tests to pass.

Most tests compare to stored values so we would need to generate new values since the new algorithm is fundamentally changed.

manthey:
@FattanePourakpour I see you have some commits why say "black".  The codebase predates the existence of black and uses flake8 and autopep8.  `tox -e format` will apply the autopep8 rules.  They differ from black (for instance, preferring single quotes), and linting will fail with black's rules.

FattanePourakpour:
@manthey thank you for the clarification. I will do `tox -e format `and commit again.

manthey:
@FattanePourakpour The CI is failing because the test_compute_fsd_features no longer matches the stored ground truth file.  We can update the ground truth to match the new values if they are what we expect.

FattanePourakpour:
@manthey Yes, since the new algorithm has fundamentally changed, the ground truth needs to be updated to match the new values for the tests to pass. 
Please let me know if there’s anything I can assist with.

manthey:
@FattanePourakpour I merged master and updated ground truth values.  One thing that took me a while was that the `trace_object_boundaries` expects the edges to be false so that it can avoid checking bounds in the cython code.  I added some code to ensure this, as otherwise we were accessing arbitrary memory (and the tests were flaky, though that seems like a success at the detecting it).

manthey (approved):

cooperlab:
@manthey thank you. We will look at this to see how we can update testing in the future.

This is a fundamental change in the feature extraction so we need to note that the values will be very different in the release (different but correct and likely better).

PR #1159, opened 2025-03-25
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

TODOs

- [ ] may be entirely drop custom CI run if pre-commit would detect, although it might be  beneficial for annotation of PRs


manthey:
@yarikoptic codespell is already in the .pre-commit-config.yaml, so this introduces a second copy.  It is a good question why it was chosen not to apply to ipynb files (my vague belief is that ages ago it raised issues about some parts that it shouldn't have, but I can't remember).  I'd rather not have configurations in multiple places -- can the precommit hook use the toml file or vice versa?

yarikoptic:
d'oh -- I should improve my script to alert on presence in pre-commit! ;)  .ipynb were likely ignored to avoid false positives for all those embedded images etc. I have added a typical embedded image regex to match/skip and there seems to be no other false positives so far.
So I moved config into `pyproject.toml` , removed skipping ipynb.  Note that there is outstanding
- https://github.com/codespell-project/codespell/issues/3196
for excludes, but seems to not hit anything here, so should be safe.

manthey (approved):

PR #1167, opened 2025-06-02
Source Development.Module instead of Development

barrettMCW:
it's looking like this should fix things!
Thanks for working on this with me!


barrettMCW:
`python: can't open file '/home/circleci/project/setup.py': [Errno 2] No such file or directory`
which makes sense. the github workflow should work though:
https://github.com/barrettMCW/HistomicsTK/actions/runs/15397121586

barrettMCW:
changes merged in another commit

PR #1184, opened 2025-08-11
Rust Bindings for Macenko PCA

barrettMCW:
From #1181
Hey y'all! I should have something that builds and adds a histomicstk._rust module with all the bindings. As it stands, I have them separate for testing purposes, just appended _rs to the original function name. Assuming all looks good to everyone I can replace the python versions with the rust bindings.

Do we want to add more bindings from rust? Like utils or _linalg? (get_principal_components probably performs about the same, but magnitude and normalize might be faster)

Noticed some things too. There's slight discrepancies between the matrices still, I'm not sure what it is. It's all identical up to f32, but beyond that precision it gets a little off. Also realized my benchmarks were slightly biased, still a major improvement.
Here's a non-biased result on masked 4µ:
```
Rust PCA took 3.0037682056427 seconds
HTK PCA took 11.629833936691284 seconds
Rust PCA used 8.7 GB
HTK PCA used 11.4 GB
```

### DIFF
--- a/macenko-pca/.editorconfig
+++ b/macenko-pca/.editorconfig
@@ -1,0 +1,36 @@
+# EditorConfig — https://editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.py]
+indent_size = 4
+max_line_length = 88
+
+[*.{yml,yaml}]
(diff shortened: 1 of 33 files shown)


barrettMCW:
@manthey Finally passing all the circleci stuff! If everything looks good as is, I can replace the old python versions with the rust version and double check that things still pass.

barrettMCW:
@manthey Hey! Just checking in. Is there anything I can do to help?

